PR #123, opened 2023-03-16
created initial canary helm chart

rltvty:


.github/workflows/ci_helm_canary.yml
```diff
@@ -0,0 +1,68 @@
...
+      run_id: ${{ steps.skip_check.outputs.skipped_by.id }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@v5
+        with:
+          concurrent_skipping: "never"
+          skip_after_successful_duplicate: "true"
+          paths: '["charts/kaskada-canary/**"]'
+          # Theo, what does the below line do?
+          do_not_skip: '["workflow_dispatch", "schedule"]'
```
rltvty, line 39: @therapon what does this do? should I keep this line or drop?

therapon: This says 

do not skip if this workflow was kicked off by a schedule event or a call from another workflow

PR #144, opened 2023-03-22
feat: Add Object Store Crate to Engine

kevinjnguyen:
# Description

Adds an initial integration with the object store crate. Updates the file service for `get_source_metadata` to use the object store rather than the s3 helper. 

## Changes
* Adds object store and url to the cargo toml
* Adds object store url and object store modules.
* Adds a download method (differs from sample code)
* Adds the object store registry to the file service.
* Updates the file store implementation to take the object store registry with intent to remove s3 helper.
* Replaces the s3 helper download with an object store download.

## Follow Up
* Integrate across more of the services
* Add GCP/GCS builder
* Remove s3 helper


crates/sparrow-main/src/serve/file_service.rs
```diff
@@ -55,48 +60,74 @@ impl FileService for FileServiceImpl {
...
+pub enum Error {
+    #[display(fmt = "unable to get source path from request")]
+    SourcePathError,
+    #[display(fmt = "unable to create temporary file")]
+    TempFileDownloadFileError,
+    #[display(fmt = "unable to download file from object store")]
+    ObjectStoreError,
+    #[display(fmt = "schema error: '{_0}'")]
+    SchemaError(String),
+}
```
jordanrfrazier, line 90: clippy things

crates/sparrow-main/src/serve/file_service.rs
```diff
@@ -55,48 +60,74 @@ impl FileService for FileServiceImpl {
...
-    s3: &S3Helper,
+    object_store_registry: &ObjectStoreRegistry,
     source: &FilePath,
-) -> anyhow::Result<FileMetadata> {
+) -> error_stack::Result<FileMetadata, Error> {
     let source = source
         .path
         .as_ref()
-        .context(context_code!(Code::InvalidArgument, "Missing source_path"))?;
+        .context(context_code!(Code::InvalidArgument, "Missing source_path"))
```
jordanrfrazier, line 100: `context` comes from `use anyhow::Context` - let's try to remove anyhow here if we're swapping to error_stack. So can just remove this line basically. And if you want more context, can add a string to the `Error::SourcePath`

crates/sparrow-main/src/serve/file_service.rs
```diff
@@ -105,13 +136,18 @@ pub(crate) async fn get_source_metadata(
...
-    let metadata = RawMetadata::try_from(&source)?;
-    let schema = Schema::try_from(metadata.table_schema.as_ref()).with_context(|| {
-        format!(
-            "Unable to encode schema {:?} for source file {:?}",
-            metadata.table_schema, source
-        )
-    })?;
+    let metadata = RawMetadata::try_from(&source)
+        .into_report()
+        .change_context(Error::SchemaError("unable to get raw metadata".to_owned()))?;
```
jordanrfrazier, line 141: could `attach_printable_lazy!(format!("{}", schema)`, that would help debugging.

crates/sparrow-main/src/serve/file_service.rs
```diff
@@ -105,13 +136,18 @@ pub(crate) async fn get_source_metadata(
...
+    let schema = Schema::try_from(metadata.table_schema.as_ref())
+        .into_report()
+        .change_context(Error::SchemaError(
+            format!(
+                "Unable to encode schema {:?} for source file {:?}",
+                metadata.table_schema, source
+            )
+            .to_owned(),
+        ))?;
```
jordanrfrazier, line 150: drop the `to_owned()` I guess

crates/sparrow-runtime/src/stores/object_store_url.rs
```diff
@@ -0,0 +1,269 @@
...
+use itertools::Itertools;
+
+use crate::ObjectStoreRegistry;
+
+use super::object_stores::Error;
+
+/// A string referring to a file in an object store.
+/// TODO: Debug this. It doesn't like the serialize on the URL.
+// #[derive(Debug, serde::Serialize, serde::Deserialize, Clone)]
+// #[serde(transparent)]
```
jordanrfrazier, line 16: Want to pair up on this?

crates/sparrow-main/src/serve/file_service.rs
bjchambers: We could also have one that makes it easier to print the schema -- although here, we don't yet have the schema I think...

crates/sparrow-runtime/Cargo.toml
```diff
@@ -69,6 +69,8 @@ tokio-util.workspace = true
 tonic.workspace = true
 tracing.workspace = true
 uuid.workspace = true
+object_store = { version = "0.5.5", features = ["aws", "gcp"] }
```
bjchambers, line 72: Should add these to the workspace, and use `object_store.workspace = true` here. This lets us have the versions defined in one place and used consistently in all the sub crates.

kevinjnguyen: Added in the latest revision.

crates/sparrow-runtime/src/stores/object_store_url.rs
kevinjnguyen: Paired on Slack thread.

crates/sparrow-runtime/src/stores/object_stores.rs
```diff
@@ -0,0 +1,184 @@
...
+            Ok(object_store)
+        } else {
+            let object_store = create_object_store(&key)?;
+            self.put_object_store(key, object_store.clone());
+            Ok(object_store)
+        }
+    }
+
+    fn get_object_store(&self, key: &ObjectStoreKey) -> Option<Arc<dyn ObjectStore>> {
+        // TODO: How to safely unwrap this? The read() returns a Poisoned error.
```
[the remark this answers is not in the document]
kevinjnguyen: Updated with some basic `error_stack` errors.

crates/sparrow-main/src/serve/file_service.rs
kevinjnguyen: Added

crates/sparrow-runtime/src/lib.rs
```diff
@@ -40,6 +41,7 @@ pub use batch::*;
 pub use metadata::*;
 pub use read::*;
 use sparrow_api::kaskada::v1alpha::execute_request::Limits;
+pub use stores::*;
```
jordanrfrazier, line 44: You shouldn't need both `pub mod stores` and `pub use stores::*`. Generally prefer `pub mod` when you want to preserve the import path, if it gives context when used. For example, with `pub mod stores`, other files would import the `path()` function in object_stores_url like `sparrow-runtime::stores::ObjectStoreUrl::path(..)`. But if you use `pub use stores::*`, then someone could call that function with `sparrow-runtime::ObjectStoreUrl::path(..)`.

crates/sparrow-runtime/src/metadata/raw_metadata.rs
```diff
@@ -1,13 +1,33 @@
...
 use crate::metadata::file_from_path;
+use crate::object_store_url::ObjectStoreKey;
+use crate::{ObjectStoreRegistry, ObjectStoreUrl};
+
+#[derive(derive_more::Display, Debug)]
+pub enum Error {
+    #[display(fmt = "object store error for path: {_0}")]
+    ObjectStore(String),
+    #[display(fmt = "no object store registry")]
+    NoObjectStoreRegistry,
```
jordanrfrazier, line 21: nit: I'd probably call this `MissingObjectStoreRegistry`, or `ObjectStoreRegistryNotFound`

crates/sparrow-runtime/src/metadata/raw_metadata.rs
```diff
@@ -1,13 +1,33 @@
...
+    #[display(fmt = "local file error")]
+    LocalFile,
+    #[display(fmt = "download error")]
+    DownloadIssue,
+    #[display(fmt = "reading schema error")]
+    ReadSchemaIssue,
```
jordanrfrazier, line 27: nit: Suffixing these with `Issue` is just adding a synonym for `Error`. I would leave off, and prefer `Error::Download { context: String }` etc, probably. Or just attach context manually if preferred.

crates/sparrow-runtime/src/metadata/raw_metadata.rs
```diff
@@ -21,43 +41,117 @@ pub struct RawMetadata {
 }
 
 impl RawMetadata {
-    pub fn try_from(source_path: &Path) -> anyhow::Result<Self> {
+    /// Createa a `RawMetadata` from a `v1alpha::file_path::Path`.
```
jordanrfrazier, line 44: "create"

Also, are there docs that explain the difference between `RawMetadata` and `TableMetadata` (or whatever the other one used often is)?

crates/sparrow-main/src/serve/file_service.rs
```diff
@@ -55,63 +54,53 @@ impl FileService for FileServiceImpl {
...
+    object_store_registry: Arc<ObjectStoreRegistry>,
     request: tonic::Request<GetMetadataRequest>,
 ) -> anyhow::Result<tonic::Response<GetMetadataResponse>> {
     let request = request.into_inner();
 
     let file_metadatas: Vec<_> = request
         .file_paths
         .iter()
-        .map(|source| get_source_metadata(&s3, source))
+        .map(|source| get_source_metadata(Some(object_store_registry.as_ref()), source))
```
jordanrfrazier, line 65: Any particular reason the registry is passed as optional? Feels like since we're creating it on startup, we know it'll exist.

crates/sparrow-runtime/src/metadata/raw_metadata.rs
```diff
@@ -21,43 +41,117 @@ pub struct RawMetadata {
...
+        let object_store_key = object_store_url
+            .key()
+            .change_context_lazy(|| Error::ObjectStore(path.to_owned()))?;
+        match (object_store_key, object_store_registry) {
+            (ObjectStoreKey::Local, _) => {
+                let path = object_store_url
+                    .path()
+                    .change_context_lazy(|| Error::ObjectStore(path.to_owned()))?
+                    .to_string();
+                let path = format!("/{}", path);
```
jordanrfrazier, line 122: This probably deserves some documentation. What happens if the `path` already has a prefix `/`?

crates/sparrow-runtime/src/stores/object_store_url.rs
```diff
@@ -0,0 +1,271 @@
...
+            "file" => Ok(ObjectStoreKey::Local),
+            "mem" => Ok(ObjectStoreKey::Memory),
+            // S3 is the traditional S3 prefix for reading from S3.
+            // S3a is the protocol designed for scalability with Hadoop reading in mind.
+            // See: https://aws.amazon.com/blogs/opensource/community-collaboration-the-s3a-story/
+            "s3" | "s3a" => {
+                let bucket = self
+                    .url
+                    .host_str()
+                    .ok_or_else(|| Error::UrlMissingHost(self.url.clone()))?
```
jordanrfrazier, line 39: Wonder if it's worth introducing `lazy_format!` for cases like these to avoid the clone

jordanrfrazier (approved):

crates/sparrow-runtime/src/metadata/raw_metadata.rs
kevinjnguyen: This is documented in the rust doc for `table_schema` ([here](https://github.com/kaskada-ai/kaskada/blob/main/crates/sparrow-runtime/src/metadata/raw_metadata.rs#L20))

wren/compute/manager.go
```diff
@@ -868,12 +868,12 @@ func (m *Manager) GetFileSchema(ctx context.Context, fileInput internal.FileInpu
...
 	case kaskadafile.TypeCsv:
-		filePath = &v1alpha.FilePath{Path: &v1alpha.FilePath_CsvPath{CsvPath: ConvertURIForCompute(fileInput.GetURI())}}
+		filePath = &v1alpha.FilePath{Path: &v1alpha.FilePath_CsvPath{CsvPath: fileInput.GetURI()}}
 	case kaskadafile.TypeParquet:
-		filePath = &v1alpha.FilePath{Path: &v1alpha.FilePath_ParquetPath{ParquetPath: ConvertURIForCompute(fileInput.GetURI())}}
+		filePath = &v1alpha.FilePath{Path: &v1alpha.FilePath_ParquetPath{ParquetPath: fileInput.GetURI()}}
 	default:
 		subLogger.Warn().Msg("user didn't specifiy file type, defaulting to parquet for now, but will error in the future")
-		filePath = &v1alpha.FilePath{Path: &v1alpha.FilePath_ParquetPath{ParquetPath: ConvertURIForCompute(fileInput.GetURI())}}
+		filePath = &v1alpha.FilePath{Path: &v1alpha.FilePath_ParquetPath{ParquetPath: fileInput.GetURI()}}
```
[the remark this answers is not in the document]
kevinjnguyen: This changed to normalize the paths moving forward. In some places we use `file:///` and other places we do not. Moving forward, all paths will need the prefix of their source e.g. s3:// or gs://.

crates/sparrow-runtime/src/metadata/raw_metadata.rs
kevinjnguyen: Updated with some docs.

crates/sparrow-main/src/serve/file_service.rs
kevinjnguyen: I originally made it optional since it made the tests not require the creation of it/passing but making it required simplifies the logic so I made those changes.

PR #182, opened 2023-03-29
fix: fix temp file deletion

jordanrfrazier:
Explicitly closes the temp file, which should(tm) ensure that the temporary file is not eligible for deletion until that happens.


jordanrfrazier:
I'll get with eric when he's online to see if I can test this locally myself first  before merging

PR #191, opened 2023-03-30
(docs): Updates paragraph around hello world links

therapon:


docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,6 +1,11 @@
 = Getting Started
 
-There are several ways to get started with Kaskada.
+Kaskada's query language focuses on event-based data processing making it easier to reason about when things happen, the state at specific points in time, and how results change over time.
+Kaskada queries are simple, declarative and composable providing a more expressive way to query event-based data.
```
bjchambers, line 4: Maybe for now, add a link to https://kaskada.io/docs-site/kaskada/main/overview/what-is-kaskada.html? That has more content that resembles a tour.

For a longer term solution, it may make sense to evolve *this* page to be a quick tour of Kaskada. Then we could say something like this Getting Started page is a tour of what you can do with Kaskada. If you want to read it before these getting started guides you can, otherwise you can read the getting started guide as a more focused introduction to specific things.

bjchambers (approved):

docs-src/modules/getting-started/pages/index.adoc
therapon: Added the link around the text `focuses on event-based data processing` 

Seemed the most relevant text to the target page with timelines.

PR #213, opened 2023-04-05
changed last_publish_time to i64

epinzur:


proto/kaskada/kaskada/v1alpha/pulsar.proto
```diff
@@ -48,5 +47,5 @@ message PulsarSubscription {
   // subscribing to the pulsar broker.
   string subscription_id = 2;
 
-  google.protobuf.Timestamp last_publish_time = 3;
+  int64 last_publish_time = 3;
```
bjchambers, line 50: To verify -- this is a backwards incompatible change. We should not make this if we have released any code using the previous definition that could potentially try to talk with the updated version.

rltvty: we have not released any code that uses this.  In fact no current code uses this.

crates/sparrow-main/src/prepare.rs
```diff
@@ -179,7 +179,7 @@ impl PrepareCommand {
                     PulsarSubscription {
                         config: Some(pulsar_config),
                         subscription_id: pulsar_subscription,
-                        last_publish_time: None,
+                        last_publish_time: 0,
```
bjchambers, line 182: FWIW, I have seen many problems with systems that use `0` to represent "the time before everything else"... we may want to handle "what if something has a time of 0"... but it's probably OK?

bjchambers (approved):

crates/sparrow-main/src/prepare.rs
rltvty: this will eventually come from wren, instead of the current hard coded value.

PR #231, opened 2023-04-17
Object Store for FileService and PrepareService

kevinjnguyen:
Adds ObjectStore support to FileService and PrepareService, currently supporting local, s3, and google storage.

Work will need to be done on another PR to add ObjectStore support to the ComputeService. For now Compute only supports S3 & local storage.

### DIFF
--- a/wren/compute/prepare.go
+++ b/wren/compute/prepare.go
@@ -172,7 +172,7 @@ func (m *Manager) executePrepare(ctx context.Context, owner *ent.Owner, prepareJ
 		prepareReq := &v1alpha.PrepareDataRequest{
 			SourceData:       sourceData,
 			Config:           computeTable.Config,
-			OutputPathPrefix: ConvertURIForCompute(prepareOutputURI),
+			OutputPathPrefix: prepareOutputURI,
 			FilePrefix:       prepareFilePrefix,
 			SlicePlan:        prepareJob.SlicePlan,
 		}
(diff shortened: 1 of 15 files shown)


wren/service/table.go
```diff
@@ -292,32 +292,17 @@ func (t *tableService) loadFileIntoTable(ctx context.Context, owner *ent.Owner,
...
-
-	newObject, err := t.objectStoreClient.CopyObjectIn(ctx, fileInput.GetURI(), toPath)
-	if err != nil {
-		return nil, errors.WithMessagef(err, "loading file: %s into table: %s", fileInput, kaskadaTable.Name)
-	}
-
-	fileIdentifier, err := t.objectStoreClient.GetObjectIdentifier(ctx, newObject)
-	if err != nil {
-		return nil, errors.WithMessagef(err, "getting identifier for file: %s in table: %s", fileInput, kaskadaTable.Name)
-	}
```
epinzur, line 305: I don't think this should be removed.

wren/compute/prepare.go
```diff
@@ -171,7 +171,7 @@ func (m *Manager) executePrepare(ctx context.Context, owner *ent.Owner, prepareJ
 		prepareReq := &v1alpha.PrepareDataRequest{
 			SourceData:       sourceData,
 			Config:           computeTable.Config,
-			OutputPathPrefix: ConvertURIForCompute(prepareOutputURI),
+			OutputPathPrefix: prepareOutputURI,
```
epinzur, line 174: if sparrow now supports all the standard file URIs correctly (`file:///, s3://, gs://, https://`) then there are probably several other places where `ConvertURIForCompute` and its partner `ConvertURIForManager` should be removed.

Ideally both functions should be removed in their entirety.

epinzur (changes requested):
please re-add the code path in wren that copies files to the owned storage location, so that we don't change current functionality of the system.

jordanrfrazier (approved):

epinzur (approved):

PR #235, opened 2023-04-19
Improve query section of docs

kerinin:
This was sort of a grab bag, hopefully it's more useful now.


docs-src/modules/developing/pages/queries.adoc
```diff
@@ -1,11 +1,155 @@
...
+=== Pipelined operations
+
+Pipe syntax allows multiple operations to be chained together. Write your operations in the same order you think about them. It's timelines all the way down, making it easy to aggregate the results of aggregations.
+
+[source,Fenl]
+----
+{
+  largest_spend_over_2_purchases: purchase.amount
+  | when(purchase.category == "food") 
+  | sum(window=sliding(2, $input)) # Inner aggregation
```
therapon, line 70: maybe add a note on what `$input` is?

docs-src/modules/developing/pages/queries.adoc
```diff
@@ -1,11 +1,155 @@
...
+  | max()                          # Outer aggregation
+}
+----
+
+=== Row generators
+
+Pivot from events to time-series. Unlike grouped aggregates, generators produce rows even when there's no input, allowing you to react when something doesn't happen.
```
therapon, line 77: This is a first for me :) 
What is the generator? `mean()`? 
If we are to use the term "generators" we should point out which part of the expression is the generator. Maybe also a link to other supported generators?

docs-src/modules/developing/pages/queries.adoc
```diff
@@ -1,11 +1,155 @@
...
+[source,Fenl]
+----
+{
+  signups_per_hour: count(signups, window=since(daily()))
+  | when(daily())
+  | mean()
+}
+----
+
+=== Continuous expressions
```
therapon, line 88: What is the difference (if any) between generators and continuous expression?

docs-src/modules/developing/pages/queries.adoc
```diff
@@ -1,11 +1,155 @@
...
+Observe the value of aggregations at arbitrary points in time. Timelines are either “discrete” (instantaneous values or events) or “continuous” (values produced by a stateful aggregations). Continuous timelines let you combine aggregates computed from different event sources.
+
+[source,Fenl]
+----
+let product_average = review
+| with_key(review.product_id)
+| mean()
+
+in {
+  average_product_review: product_average | lookup(purchase.product_id)),
```
therapon, line 99: Maybe a note to explain `lookup` ?

docs-src/modules/developing/pages/queries.adoc
```diff
@@ -1,11 +1,155 @@
...
+}
+----
+
+Shift values forward (but not backward) in time, allowing you to combine different temporal contexts without the risk of temporal leakage. Shifted values make it easy to compare a value “now” to a value from the past.
```
therapon, line 103:
```suggestion
Shift values forward (but not backward) in time, allows you to combine different temporal contexts without the risk of temporal leakage. Shifted values make it easy to compare a value “now” to a value from the past.
```

docs-src/modules/developing/pages/queries.adoc
```diff
@@ -1,11 +1,155 @@
...
+
+=== Native time travel
```
therapon, line 105: There is no English text here :) What is the intent of this code snippet?

docs-src/modules/developing/pages/queries.adoc
therapon, line 103: Should this be another section?

docs-src/modules/developing/pages/queries.adoc
kerinin: Going to just remove it.

docs-src/modules/developing/pages/queries.adoc
kerinin: Generators / tick functions are operations that produce rows based on time, rather than data.

clients/python/src/kaskada/client.py
```diff
@@ -159,9 +159,13 @@ def get_client(client: Optional[Client] = None) -> Client:
...
+        raise ValueError(
+            "No client was provided, and no global client is configured. Consider setting the global client by creating a session, for example 'kaskada.api.session.LocalBuilder().build()`."
+        )
     elif client is None:
         client = KASKADA_DEFAULT_CLIENT
 
-    validate_client(client)
+    if client is not None:
+        validate_client(client)
+
```
therapon, line 170: This will still error on mypy since the function returns `client` which is still `Optional` 

I used 

```
assert client is not None
```
and then after that line you can have 

```
validate_client(client)
return client
```

docs-src/modules/developing/pages/queries.adoc
```diff
@@ -1,11 +1,160 @@
...
+| mean()
+
+in {
+  average_product_review: product_average | lookup(purchase.product_id)),
+}
+----
+
+In this example the `lookup` function is used to observe a value computed for a different entity.
+The variable `product_average` computes the average review using the product's ID as entity.
+The lookup starts with each `purchase`, then looks up the current value of the product's average review, for the procduct ID specified in the purchase.
```
therapon, line 106:
```suggestion
The lookup starts with each `purchase`, then looks up the current value of the product's average review, for the product ID specified in the purchase.
```

clients/python/src/kaskada/client.py
```diff
@@ -159,9 +159,13 @@ def get_client(client: Optional[Client] = None) -> Client:
...
-        raise ValueError("No client was provided, and no global client is configured. Consider setting the global client by creating a session, for example 'kaskada.api.session.LocalBuilder().build()`.")
+        raise ValueError(
+            "No client was provided, and no global client is configured. Consider setting the global client by creating a session, for example 'kaskada.api.session.LocalBuilder().build()`."
+        )
     elif client is None:
         client = KASKADA_DEFAULT_CLIENT
 
-    validate_client(client)
+    if client is not None:
+        validate_client(client)
```
kevinjnguyen, line 169:
```suggestion
    assert client is not None
    validate_client(client)
```

docs-src/modules/developing/pages/queries.adoc
```diff
@@ -1,11 +1,160 @@
...
+
+A given query can be computed in different ways.
+
+=== Configuring how timelines are converted into tables
+
+You can either return a table describing each change in the timeline, or a table describing the "final" value of the timeline.
 
 Every query produces a timeline which may be returned in two different ways -- the final results (at a specific time) or all historic results.
-The "result behavior" configures which results are produced.
+The "result behavior" configuresj which results are produced.
```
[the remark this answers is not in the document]
kerinin: yup

kevinjnguyen (approved):

kevinjnguyen (approved):

PR #238, opened 2023-04-19
Use the correct path syntax and highlight with bash

kerinin:
Bash has prettier formatting than shell.


therapon:
A previous PR broke the type checking step for python 
009fefd0ce978deffa914a0c4ce227ab3e331e0e4ba04fe5edea0c7c884efe01L166

`mypy` is not clever enough to deduce that something is no longer optional :) 

I have a fix for it https://github.com/kaskada-ai/kaskada/pull/237/files#diff-009fefd0ce978deffa914a0c4ce227ab3e331e0e4ba04fe5edea0c7c884efe01R167

But please make sure if you are changing python code that all steps pass before merging to main

therapon (approved):

PR #240, opened 2023-04-19
Updates Python Client to use file URI Prefix

### LINKED ISSUE #239: bug: file:///
**Description**
There is currently discrepancy between the usage of the Python Client and the CLI. There is various support from the API to the CLI for the file paths allowed. In the Python Client, we are not able to pass in the file URI e.g. `file://awkward.tacos.parquet` since it creates the path under the hood.

**To Reproduce**
1. Load data to a table using the `file://path` instead of the path.
```python
kaskada.load("table_name", "file:///Users/kevin.nguyen/Desktop/super-secret.parquet")
```

**Actual Behavior**
Throws a validation error.

**Expected Behavior**
The paths provided should be the same across the Python Client and the CLI.

kevinjnguyen:
# Description
Closes #239 

## Changes
* Updates the python client to accept the file:// prefix for file paths and removes the logic for converting to a URI


clients/python/src/kaskada/utils.py
```diff
@@ -98,19 +98,6 @@ def handleGrpcError(rpc_error: grpc.Call):
...
-    # Remote file stores are prefixed appropriately.
-    if (
-        file.startswith("s3://")
-        or file.startswith("gs://")
-        or file.startswith("https://")
-    ):
-        return file
-    # TODO: Verify that the object actually exists locally
-    path = Path(file).absolute()
-    return f"file://{path}"
```
epinzur, line 111: unless I'm mis-understanding the request, I think we should still keep this functionality

epinzur (changes requested):
please add the ability to pass `file://` uri from python, without removing the current path-helping functionality when a relative or absolute path is passed.

epinzur (approved):

PR #262, opened 2023-04-21
added resource delete to the cli

### LINKED ISSUE #250: feat: Allow deleting resources with the CLI
**Summary**
It should be possible to delete resources you create with the CLI. Currently you can create and update resources using `sync`, but there's no way to delete a resource (table, view, or materialization)

epinzur:
also:
* re organized the other cli code into per-section packages.
* changed the code to not send an empty client-id if no client-id flag is passed. This fixes an issue where the python client and the CLI were using different client-ids by default

closes #250

### DIFF
--- a/docs-src/modules/developing/pages/materializations.adoc
+++ b/docs-src/modules/developing/pages/materializations.adoc
@@ -78,6 +78,7 @@ purchase statistics for each user. This definition depends on business
 logic and might require some iteration to get just right.
 
 [source,python]
+.with Python:
 ----
 from kaskada import materialization
 from kaskada.api.session import LocalBuilder
@@ -114,6 +115,7 @@ for your user. An optional search string can filter the response set.
 Here is an example of listing tables:
 
 [source,python]
+.with Python:
 ----
 from kaskada import materialization
 from kaskada.api.session import LocalBuilder
@@ -128,6 +130,7 @@ materialization.list_materializations("Purchase")
 You can get a materialization using its name:
 
 [source,python]
+.with Python:
 ----
 from kaskada import materialization
 from kaskada.api.session import LocalBuilder
@@ -148,6 +151,7 @@ new expression.
 You can delete a materialization using its name:
 
 [source,python]
+.with Python:
 ----
 from kaskada import materialization
 from kaskada.api.session import LocalBuilder
@@ -158,6 +162,7 @@ materialization.delete_materialization("PurchaseStats")
 ----
 
 [source,bash]
+.with the CLI:
 ----
 ./kaskada-cli materialization delete -m PurchaseStats
 ----
--- a/docs-src/modules/developing/pages/materializations.adoc
+++ b/docs-src/modules/developing/pages/materializations.adoc
@@ -68,7 +68,7 @@ materializations:
 ----
 ////
 
-== Managing Materializations with Python
+== Managing Materializations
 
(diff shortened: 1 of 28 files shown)


clients/cli/cmd/load.go
```diff
@@ -25,32 +26,26 @@ var loadCmd = &cobra.Command{
...
 func init() {
-	rootCmd.AddCommand(loadCmd)
-
-	loadCmd.Flags().String("file-type", "parquet", "(Optional) The type of file to load.  Either 'parquet' or 'csv'.")
-	loadCmd.Flags().String("table", "", "The table to load data into.")
-	loadCmd.Flags().String("file-path", "", "The path of the file to load on the Kaskada instance.")
-
-	viper.BindPFlag("file-type", loadCmd.Flags().Lookup("file-type"))
-	viper.BindPFlag("table", loadCmd.Flags().Lookup("table"))
-	viper.BindPFlag("file-path", loadCmd.Flags().Lookup("file-path"))
```
epinzur, line 55: viper is more for system configuration than for run-time use.  we should remove the viper usage inside most commands.

docs-src/modules/developing/pages/materializations.adoc
```diff
@@ -157,5 +157,10 @@ session = LocalBuilder().build()
 materialization.delete_materialization("PurchaseStats")
 ----
 
+[source,bash]
```
epinzur, line 160: should this be `[source,cli]` or `[source,kaskada-cli]` to get the mouseover client name to be "more" correct?

kerinin: The second argument specifies the syntax highlighter. We could provide a highlighter for the CLI, but I think it would be better to just use the Bash highlighting.

docs-src/modules/developing/pages/materializations.adoc
```diff
@@ -157,5 +157,10 @@ session = LocalBuilder().build()
...
+----
```
kerinin, line 161: I think we should call out that this is a separate way of deleting a materialization. I think as a first pass (until we have tabs) we could add a title to the example, like

```
[source,Python]
.Deleting a materialization with Python
----
from kaskada import materialization

session = LocalBuilder().build()

materialization.delete_materialization("PurchaseStats")
----

[source,bash]
.Deleting a materialization with the CLI
----
./kaskada-cli materialization delete -m PurchaseStats
----
```

As is, it sort of reads like you need to _both_ delete it in Python and then run a CLI command.

docs-src/modules/developing/pages/tables.adoc
```diff
@@ -110,6 +110,11 @@ session = LocalBuilder().build()
 table.delete_table(table = "Purchase")
 ----
 
+[source,bash]
+----
+./kaskada-cli table delete -t Purchase
```
kerinin, line 115: Same comment re: titles

docs-src/modules/developing/pages/views.adoc
```diff
@@ -137,19 +137,23 @@ session = LocalBuilder().build()
 view.delete_view("PurchaseStats")
 ----
 
+[source,bash]
+----
+./kaskada-cli view delete -v PurchaseStats
```
kerinin, line 142: ...same

kerinin (approved):

PR #269, opened 2023-04-21
docs: Add example based on timeline presentation

bjchambers:
This currently has a partial implementation of timeline visualization (with some rough edges identified in the notebook) to be worked on.

Once complete this notebook can be used to generate the SVGs (and an embeddable version) of the timeline examples.


examples/kaskada_oss.ipynb
```diff
@@ -0,0 +1,397 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "Purchases = pd.DataFrame.from_dict([\n",
+    "    { 'time': '2022-10-27', 'user': 'Ben',   'amount': 5,  'item': 'widget' },\n",
```
bjchambers, line 95: Sets up fake data

examples/kaskada_oss.ipynb
```diff
@@ -0,0 +1,397 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "plot = plot_timelines(aggregate.dataframe, '_time', '_key', timelines = [\n",
+    "    Timeline('amount', label = 'Purchases.amount'),\n",
```
bjchambers, line 253: Current iteration of plotting. Things missing for the purposes of the timeline examples:

1. Using different dataframes for different timelines (to get different points).
2. Different entities on different timelines
3. Categorical (string) values

I'm going to look at adding those to the hacky visualization script, since it's useful to generate the images from this notebook (and provides a place for people to play with the queries once they've read or heard the content).

examples/kaskada_oss.ipynb
```diff
@@ -0,0 +1,397 @@
...
+    "    Timeline('sum_amount', label = 'sum(Purchases.amount)', continuous = True)\n",
+    "])\n",
+    "\n",
+    "show(plot)"
```
bjchambers, line 257: Show renders the plots in the notebook. We can also save a standalone HTML format (eg., for embedding in blog posts) or an SVG (for conversion to PNG and embedding in presentations).

kerinin (approved):
Could we put his in a subdirectory? The `examples` folder is getting pretty crowded - I'd prefer to put everything in subdirectories, but at minimum I think new things (especially new things that include files outside the notebook) ought to go in a subdirectory.

bjchambers:
> Could we put his in a subdirectory? The `examples` folder is getting pretty crowded - I'd prefer to put everything in subdirectories, but at minimum I think new things (especially new things that include files outside the notebook) ought to go in a subdirectory.

A few thoughts -- for examples, we should probably start a new repo at some point. Many other projects do this so that it is easier for users to see "this is where the examples are" without getting lost in the mass of the main repo.

I think it could make sense, although my intention is actually for that "extra file" to be used in multiple notebooks and/or become part of the Python client (conditional on bokeh being installed) since it is pretty small, and would be a nice "visualization" extension.

I think for now, it may make sense to put this in a `talks` subdirectory (if we want to be more generic) or `timeline-talk` (if we want to be more specific).

PR #270, opened 2023-04-24
updated buf and plugins

### LINKED ISSUE #242: buf build plugins in CI have been deprecated
The buf plugins we are currently using in CI have been deprecated.

Protobuf CI sections are reporting warnings like these: 
```
WARN	Plugin "buf.build/grpc/go" is deprecated
WARN	Plugin "buf.build/protocolbuffers/go" is deprecated
WARN	Plugin "buf.build/grpc-ecosystem/grpc-gateway" is deprecated
WARN	Plugin "buf.build/protocolbuffers/python" is deprecated
WARN	Plugin "buf.build/go-sdk/protoc-gen-validate" is deprecated
WARN	Plugin "buf.build/grpc/python" is deprecated
```

Buf recently moved out of beta, and I think a bunch of things got renamed.  See https://buf.build/plugins for the new naming conventions.

epinzur:
resolves #242

### DIFF
--- a/.github/workflows/ci_client_cli.yml
+++ b/.github/workflows/ci_client_cli.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Install Buf
         uses: bufbuild/buf-setup-action@v1.12.0
         with:
-          version: "1.10.0"
+          version: "1.17.0"
           buf_user: ${{ secrets.BUF_USER }}
           buf_api_token: ${{ secrets.BUF_API_TOKEN }}
           github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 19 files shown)


therapon:
There was a failure on `pylint`. Not 100% sure on the root cause, but it looks like the new generation (and maybe a newer version of `pylint`?) caused more warnings and errors that dropped the code's score below the value we set as the minimum. That caused the task to fail. 

I applied some of the recommendations and added an ignore statement to the files that use protobuf generated code and caused `pylint` to generate errors for "accessing missing fields". this seemed safe since pylint cannot detect the members in those classes due to the use of reflection.

therapon (approved):

epinzur:
thanks for cleaning this up @therapon!

PR #291, opened 2023-04-27
Rewrite the "getting started" page

kerinin:
The goal here is to quickly introduce the concepts needed to understand Kaskada, and provide links to deeper reading as appropriate.

### DIFF
--- a/docs-src/modules/getting-started/pages/index.adoc
+++ b/docs-src/modules/getting-started/pages/index.adoc
@@ -1,14 +1,91 @@
 = Getting Started
 
-Kaskada's query language xref:overview:what-is-kaskada.adoc[focuses on event-based data processing] making it easier to reason about when things happen, the state at specific points in time, and how results change over time.
-Kaskada queries are simple, declarative and composable providing a more expressive way to query event-based data.
+== Reasoning about events with timelines
 
-There are several ways to quickly get started with Kaskada.
+Kaskada is a unified event processing engine that provides stateful stream processing in a high-level query language designed specifically for reasoning about events in bulk and in real time.
+Kaskada is built on the idea of a _timeline_ - the history of how a value changes over time for a specific entity or group.
 
-* xref:getting-started:hello-world-jupyter.adoc[]: How to get up and running with Kaskada inside a Jupyter notebook running locally. 
-This quick tutorial will help you install and run Kaskada from within a Jupyter notebook, load some sample data into Kaskada, use Kaskada queries and Python within Jupyter to explore that data.
-* xref:getting-started:hello-world-cli.adoc[]: How to execute queries from the command line using the Kaskada CLI. 
-This quick tutorial will help you install and run Kaskada from the command line, load some sample data into Kaskada, and use Kaskada queries to explore that data all on the command line.
+[stream_viz,name=basic-sum]
+.Aggregating events as a timeline
+....
+[
+    {
+        "label": "Purchase.amount",
+        "kind": "discrete",
+        "data": [
+            [
+                {"t": 2, "v": "5"},
+                {"t": 4, "v": "2"},
+                {"t": 13, "v": "13"},
+                {"t": 18, "v": "4"}
+            ]
+        ]
+    },
+    {
+        "label": "Purchase.amount | sum()",
+        "kind": "continuous",
+        "data": [
+            [
+                {"t": 2, "v": "5"},
+                {"t": 4, "v": "7"},
+                {"t": 13, "v": "20"},
+                {"t": 18, "v": "24"}
+            ]
+        ]
+    }
+]
+....
+
+Timelines give you declarative queries and the ability to reason about temporal context, time travel, sequencing, timeseries, and more.
+Transforming and combining timelines allow you to intuitively express computations over events.
+
+****
+⭢ Read more about timelines in xref:overview:what-is-kaskada.adoc[]
+****
+
+With Kaskada, it's timelines all the way down - every operation starts from timelines and produces timelines. 
+When you make a query, the resulting timeline is converted to a table in one of two ways: as a value history or as a value snapshot. 
+
+* *Value history* generates a row each time the timeline changes, and each row is associated with a different entity and point in time. 
+* *Value snapshot* generates a row for each entity at the _same_ point in time; each row is associated with a different entity, but all rows are associated with the same time.
+
+****
+⭢ Read more about converting to tables in xref:developing:queries.adoc#configuring-how-queries-are-computed[Queries]
+****
+
+== Kaskada's basic architecture
+
+Kaskada is implemented as a standalone service with two main components.
+
+* The *engine* is a stateful compute engine for transforming event data.
+* The *manager* is a stateful compute engine for transforming event data.
+
+These components are distributed as dependency-free binaries.
+Kaskada is managed using an API, and multiple clients exist to communicate with the API, including Python and a CLI binary.
+Kaskada can be configured to run as a remote service or as a local process.
+
+****
+⭢ Read more about installing Kaskada in xref:ROOT:installing.adoc[]
+****
+
+Kaskada stores data in tables. Tables consist of multiple rows, and each row is a value of the same type.
+Events must be loaded into a table before they can be queried.
+
+****
+⭢ Read more about loading data in xref:ROOT:loading-data.adoc[]
+****
+
+== Quick-starts
+
+The next section of the docs shows how to quickly get started with Kaskada.
+You'll see how to install Kaskada, create tables, load data, and make simple queries.
+You can pick the development environment that's most familiar to you.
+
+****
+⭢ Try out Kaskada using Jupyter in xref:getting-started:hello-world-jupyter.adoc[]
+
+⭢ Try out Kaskada using the CLI in xref:getting-started:hello-world-cli.adoc[]
+****
 
 [TIP]
 .Installing different clients
--- a/docs-src/modules/getting-started/pages/index.adoc
+++ b/docs-src/modules/getting-started/pages/index.adoc
@@ -7,7 +7,7 @@ This document will show you how to quickly get started using Kaskada.
 [TIP]
 ====
 Before jumping in to writing queries in Kaskada, it's a good idea to take a minute understanding "how to think with Kaskada".
-Kaskada is built on simple, powerful abstrctions that may be different than what you're accustomed to.
+Kaskada is built on timelines - a simple, powerful abstraction that may be different than what you're accustomed to.
 
 If you'd prefer to jump right in, scroll down to the xref:#quick-starts[] section
 ====
@@ -47,20 +47,21 @@ Kaskada is built on the idea of a _timeline_ - the history of how a value change
 ]
 ....
 
-Timelines give you declarative queries and the ability to reason about temporal context, time travel, sequencing, time-series, and more.
-Transforming and combining timelines allow you to intuitively express computations over events.
+Timelines allow you to reason about temporal context, time travel, sequencing, time-series, and more. 
+They allow simple, composable, declarative queries over events. 
+Transforming and combining timelines allows you to intuitively express computations over events.
 
 ****
 ⭢ Read more about timelines in xref:overview:what-is-kaskada.adoc[]
 ****
 
-With Kaskada, it's timelines all the way down - every operation starts from timelines and produces timelines. 
-When you make a query, the resulting timeline is interpreted in one of two ways: as a _history_ or as a _snapshot_. 
+With Kaskada, it's timelines all the way down - every operation's inputs and outputs are timelines. 
+When you make a query, the resulting timeline is output in one of two ways: as a _history_ or as a _snapshot_. 
 
-* A timeline *History* generates a value each time the timeline changes, and each row is associated with a different entity and point in time. 
-* A timeline *Snapshot* generates a value for each entity at the _same_ point in time; each row is associated with a different entity, but all rows are associated with the same time.
+* A timeline *History* contains a value each time the timeline changes, and each row is associated with a different entity and point in time. 
+* A timeline *Snapshot* contains a value for each entity at the _same_ point in time; each row is associated with a different entity, but all rows are associated with the same time.
 
-These values may be output in different ways, for example they can be written as rows in a Parquet file, or sent as events to a stream.
+This output may be written in different ways -- for example, they may be written to a Parquet file or sent as events to a stream.
 
 ****
 ⭢ Read more about converting to tables in xref:developing:queries.adoc#configuring-how-queries-are-computed[Queries]
@@ -84,7 +85,8 @@ Kaskada can be configured to run as a remote service or as a local process.
 
 == Kaskada's data model
 
-Kaskada stores events in xref:developing:tables.adoc[tables]. 
+Timelines begin with xref:developing:tables.adoc[tables].
+Tables are how Kaskada stores input events.
 Tables consist of multiple events, and each event is a value of the same xref:fenl:data-model.adoc[type].
 
 [%header,cols=3*]
--- a/docs-src/modules/getting-started/pages/index.adoc
+++ b/docs-src/modules/getting-started/pages/index.adoc
@@ -56,10 +56,13 @@ Transforming and combining timelines allows you to intuitively express computati
 ****
 
 With Kaskada, it's timelines all the way down - every operation's inputs and outputs are timelines. 
-When you make a query, the resulting timeline is output in one of two ways: as a _history_ or as a _snapshot_. 
+The timeline is a flexible abstraction that can be used in different ways depending on your needs.
+In some cases, you may want to know what the timeline's value is at a specific point in time, in other cases you may want to know how the timeline's value changes over time.
 
-* A timeline *History* contains a value each time the timeline changes, and each row is associated with a different entity and point in time. 
-* A timeline *Snapshot* contains a value for each entity at the _same_ point in time; each row is associated with a different entity, but all rows are associated with the same time.
+When making a query, you configure how to use the query's output timeline: you can use the timeline as either a _history_ or as a _snapshot_.
+
+* A timeline *History* contains a value each time the timeline changes, and each row describes a different entity at a different point in time. 
+* A timeline *Snapshot* contains a value for each entity at the _same_ point in time; each row is associated with a different entity, but all rows reflect the same point in time.
 
 This output may be written in different ways -- for example, they may be written to a Parquet file or sent as events to a stream.
 
@@ -124,7 +127,7 @@ Events must be loaded into a table before they can be queried.
 You can load events into a table from various data sources, for example files formatted as Parquet or CSV, or by reading events from a stream.
 
 ****
-⭢ Read more about loading data in xref:ROOT:loading-data.adoc[]
+⭢ Read more about connecting to data in xref:ROOT:loading-data.adoc[]
 ****
 
 == Quick-starts
--- a/docs-src/modules/getting-started/pages/index.adoc
+++ b/docs-src/modules/getting-started/pages/index.adoc
@@ -1,8 +1,19 @@
 = Getting Started
 
-== Reasoning about events with timelines
-
 Kaskada is a unified event processing engine that provides stateful stream processing in a high-level query language designed specifically for reasoning about events in bulk and in real time.
+
+This document will show you how to quickly get started using Kaskada.
+
+[TIP]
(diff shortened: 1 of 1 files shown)


docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,91 @@
 = Getting Started
```
bjchambers, line 1: Feels like it would be good to have an intro paragraph between the "getting started" and the first section (reasoning about veents with timelines).

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,91 @@
...
 
-Kaskada's query language xref:overview:what-is-kaskada.adoc[focuses on event-based data processing] making it easier to reason about when things happen, the state at specific points in time, and how results change over time.
-Kaskada queries are simple, declarative and composable providing a more expressive way to query event-based data.
+== Reasoning about events with timelines
 
-There are several ways to quickly get started with Kaskada.
+Kaskada is a unified event processing engine that provides stateful stream processing in a high-level query language designed specifically for reasoning about events in bulk and in real time.
+Kaskada is built on the idea of a _timeline_ - the history of how a value changes over time for a specific entity or group.
```
bjchambers, line 6: Use of "history" here could be confusing with later "timeline history" and "timeline snapshot"... but maybe not, since the "timeline history" there is the history of how a value changes over time...

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,91 @@
...
+                {"t": 4, "v": "7"},
+                {"t": 13, "v": "20"},
+                {"t": 18, "v": "24"}
+            ]
+        ]
+    }
+]
+....
+
+Timelines give you declarative queries and the ability to reason about temporal context, time travel, sequencing, timeseries, and more.
```
bjchambers, line 39: s/timeseries/time series/ or `time-series`

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,91 @@
...
+Transforming and combining timelines allow you to intuitively express computations over events.
+
+****
+⭢ Read more about timelines in xref:overview:what-is-kaskada.adoc[]
+****
+
+With Kaskada, it's timelines all the way down - every operation starts from timelines and produces timelines.
```
bjchambers, line 46: Terminology "converted to a table" may make people think this doesn't produce streams. Maybe:

When you make a query, the resulting timeline is converted to data in one of two ways -- as a timeline history or as a timeline snapshot.

* The *timeline history* generates a row for every point the timeline changed, and each row is associated with the time it changed and the corresponding entity.
* The *timeline snapshot* generates a row for each entity at the _same_ point in time. Each row is associated with a different entity, but all rows are associated with the same time.

These rows may be output to a stream or table store.
See ... for information on outputting timelines.

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,91 @@
...
+****
+⭢ Read more about converting to tables in xref:developing:queries.adoc#configuring-how-queries-are-computed[Queries]
+****
+
+== Kaskada's basic architecture
+
+Kaskada is implemented as a standalone service with two main components.
+
+* The *engine* is a stateful compute engine for transforming event data.
+* The *manager* is a stateful compute engine for transforming event data.
```
bjchambers, line 61: The manager sounds a lot like the engine...

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,91 @@
...
+
+These components are distributed as dependency-free binaries.
```
bjchambers, line 63: That depend on SSL and glibc, I believe?

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,91 @@
...
+Kaskada is managed using an API, and multiple clients exist to communicate with the API, including Python and a CLI binary.
```
bjchambers, line 64: `The Kaskada service is accessed using a gRPC API ...`

1. "managed" is confusing (given the manager)
2. good to mention gRPC
3. good to call this the "service" or something, in case we introduce more standalone / self-contained options in the future.

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,91 @@
...
+Kaskada can be configured to run as a remote service or as a local process.
+
+****
+⭢ Read more about installing Kaskada in xref:ROOT:installing.adoc[]
+****
+
+Kaskada stores data in tables. Tables consist of multiple rows, and each row is a value of the same type.
```
bjchambers, line 71: - [ ] Confusing to "store data in tables" and convert "timelines to tables". Definitely should do something like suggested above to say "convert timelines to data" or "output timelines" (or "view timelines as data", etc.)
- [ ] We should strongly consider calling this something different (later)
- [ ] AsciiDoc suggests having each sentence on a newline. So `.\n.Tables`.

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,91 @@
...
+Events must be loaded into a table before they can be queried.
```
bjchambers, line 72: Probably worth a sentence pointing out that tables (in this usage) correspond to a timeline, and may be populated from a set of Parquet files, events in a stream, etc.

(something to make it clear that this includes streams)?

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,141 @@
...
+This document will show you how to quickly get started using Kaskada.
 
-* xref:getting-started:hello-world-jupyter.adoc[]: How to get up and running with Kaskada inside a Jupyter notebook running locally. 
-This quick tutorial will help you install and run Kaskada from within a Jupyter notebook, load some sample data into Kaskada, use Kaskada queries and Python within Jupyter to explore that data.
-* xref:getting-started:hello-world-cli.adoc[]: How to execute queries from the command line using the Kaskada CLI. 
-This quick tutorial will help you install and run Kaskada from the command line, load some sample data into Kaskada, and use Kaskada queries to explore that data all on the command line.
+[TIP]
+====
+Before jumping in to writing queries in Kaskada, it's a good idea to take a minute understanding "how to think with Kaskada".
+Kaskada is built on simple, powerful abstrctions that may be different than what you're accustomed to.
```
bjchambers, line 10: "simple, powerful abstractions":

1. Spelling
2. Suggest something like "Kaskada is built on timelines -- a simple, powerful abstraction that ..."

Specifically, "abstractions" makes it sound like there are several things to learn, while "timelines -- a simple, ..." makes it singular. May make it seem friendlier to learn just one thing.

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,141 @@
...
+                {"t": 5, "v": "7"},
+                {"t": 13, "v": "20"},
+                {"t": 18, "v": "24"}
+            ]
+        ]
+    }
+]
+....
+
+Timelines give you declarative queries and the ability to reason about temporal context, time travel, sequencing, time-series, and more.
```
bjchambers, line 50: This feels weird -- we jump from the abstraction to first saying "declarative queries". Suggest maybe:


```
Timelines allow you to reason about temporal context, time travel, sequencing, time-series, and more.
They allow simple, composable, declarative queries over events.
Transforming and combining timelines allow you to intuitively express computations over events.
```

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,141 @@
...
+Transforming and combining timelines allow you to intuitively express computations over events.
+
+****
+⭢ Read more about timelines in xref:overview:what-is-kaskada.adoc[]
+****
+
+With Kaskada, it's timelines all the way down - every operation starts from timelines and produces timelines. 
+When you make a query, the resulting timeline is interpreted in one of two ways: as a _history_ or as a _snapshot_.
```
bjchambers, line 58: s/interpreted/output?

Eg., interpreted made me think it changes the meaning of the timeline (affects the timeline) rather than is just a way of outputting it.

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,141 @@
...
+Transforming and combining timelines allow you to intuitively express computations over events.
+
+****
+⭢ Read more about timelines in xref:overview:what-is-kaskada.adoc[]
+****
+
+With Kaskada, it's timelines all the way down - every operation starts from timelines and produces timelines. 
+When you make a query, the resulting timeline is interpreted in one of two ways: as a _history_ or as a _snapshot_. 
+
+* A timeline *History* generates a value each time the timeline changes, and each row is associated with a different entity and point in time.
```
bjchambers, line 60: "generates" seems weird -- it's pretty imperative.

Maybe `A timeline *History* contains a value ...`

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,141 @@
...
+
+****
+⭢ Read more about timelines in xref:overview:what-is-kaskada.adoc[]
+****
+
+With Kaskada, it's timelines all the way down - every operation starts from timelines and produces timelines. 
+When you make a query, the resulting timeline is interpreted in one of two ways: as a _history_ or as a _snapshot_. 
+
+* A timeline *History* generates a value each time the timeline changes, and each row is associated with a different entity and point in time. 
+* A timeline *Snapshot* generates a value for each entity at the _same_ point in time; each row is associated with a different entity, but all rows are associated with the same time.
```
bjchambers, line 61: same comment re generates

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,141 @@
...
+
+These values may be output in different ways, for example they can be written as rows in a Parquet file, or sent as events to a stream.
```
bjchambers, line 63: Maybe "This output may be written in different ways -- for example, they may be written to a Parquet file or sent as events to a stream."

Trying to set up a "content" (what is output) and "format" (what it's written) to relationship.

So instead of "the timeline snapshot" generating rows, it is a way of outputting a timeline that contains rows (the content). That output (content) may be formatted to different destinations.

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,141 @@
...
+You can access the API using the provided Python client or using the CLI binary.
+Kaskada can be configured to run as a remote service or as a local process.
+
+****
+⭢ Read more about installing Kaskada in xref:ROOT:installing.adoc[]
+****
+
+== Kaskada's data model
+
+Kaskada stores events in xref:developing:tables.adoc[tables].
```
bjchambers, line 87: Connect this to timelines somehow. Ideally, starting with timelines:

```
Input timelines come from tables.
Tables consist of multiple events, and each event is a value of the same type.
```

Alternatively, introduce tables and relate back:

```
...
Each input table describes a source timeline which may be used in queries.
```

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,141 @@
...
+....
+
+Timelines give you declarative queries and the ability to reason about temporal context, time travel, sequencing, time-series, and more.
+Transforming and combining timelines allow you to intuitively express computations over events.
+
+****
+⭢ Read more about timelines in xref:overview:what-is-kaskada.adoc[]
+****
+
+With Kaskada, it's timelines all the way down - every operation starts from timelines and produces timelines.
```
therapon, line 57:
```suggestion
With Kaskada, it's timelines all the way down - every operation's inputs and outputs are timelines. 
```

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -47,20 +47,21 @@ Kaskada is built on the idea of a _timeline_ - the history of how a value change
...
+Transforming and combining timelines allows you to intuitively express computations over events.
 
 ****
 ⭢ Read more about timelines in xref:overview:what-is-kaskada.adoc[]
 ****
 
-With Kaskada, it's timelines all the way down - every operation starts from timelines and produces timelines. 
-When you make a query, the resulting timeline is interpreted in one of two ways: as a _history_ or as a _snapshot_. 
+With Kaskada, it's timelines all the way down - every operation's inputs and outputs are timelines. 
+When you make a query, the resulting timeline is output in one of two ways: as a _history_ or as a _snapshot_.
```
therapon, line 59: This is the part that makes me skeptical. 

I think we went with a timeline is either discrete or continuous. Timeline is the abstraction. So the abstraction has 2 interpretations (I am avoiding the word "type" here). 

Getting the values *out of the timeline* and into something else (like a stream, a table, a data format) then has to make a decision on whether they would like to extract a snapshot or historical 

The way it is phrased here it allows for the reader to reach the conclusion that the result of a query is either historical or snapshot when earlier on we said it's all timelines (inputs and outputs). 

I think we are driving the point that 

1. Once your data is in the "system" it's timelines that you need to think of to operate on your data. 
2. Once you have the timeline that you want now you need to think on how to get the data out of the timeline and the form that you need them (stream, table, file etc.) Now you need to make a decision as to whether you want a snapshot from the timeline or the historical values from the timeline. 


But I am still digesting the names and their meaning.

bjchambers: I think I agree. Previously, we've emphasized something about separating the `query` or `feature definition` or something (the formula or definition of the value to compute) and then `query parameters`. Basically, a way to make it clear that we separate "what to compute" from "how to compute it / output it / etc.".

I'm not sure we need to go that far in this getting started page -- although it may be helpful. But, I was trying to get at it some above -- in terms of saying "1. The result of a query is a timeline. 2. The timeline may be output as a history or a snapshot. This determines the data or content that is output. 3. The timeline history or timeline snapshot may be written to a stream (Kafka, Pulsar, etc.) or a table (Parquet, CSV, C*, etc.)".

We could maybe go a little further and try to clarify that the Fenl query is only the first step -- produces timelines. And the 2nd and 3rd step are "late bound" when you execute the query.

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,143 @@
...
+                {"t": 13, "v": "{amount: 13}"},
+                {"t": 18, "v": "{amount: 4}"}
+            ]
+        ]
+    }
+]
+....
+
+Events must be loaded into a table before they can be queried.
+You can load events into a table from various data sources, for example files formatted as Parquet or CSV, or by reading events from a stream.
```
epinzur, line 124: For streams, I think this works for now, but we may want to change this in the future.  I'm not sure how to word it, but something along the lines of: 

All input data to Kaskada is defined by tables. A table can be backed by files loaded into the system, by a stream source, (or by other external sources like Apache Iceberg).

For a file-backed table, events must be loaded before they can be queried.  For a stream-backed table, Kaskada will automatically include new events in the output calculations as they arrive.

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,143 @@
...
+
+****
+⭢ Read more about loading data in xref:ROOT:loading-data.adoc[]
```
epinzur, line 127: maybe "read more about connecting data" ?

docs-src/modules/getting-started/pages/index.adoc
kerinin: See what you guys think about the revised wording.

I agree with Theo that we should make it more clear that `query -> timeline` in all cases. I tried to focus the "historical vs snapshot" distinction in the different use cases people might have.

So:
- Queries _output_ timelines (to tie back to operations inputs & outputs being timelines)
- Timelines are _used_ in different ways (relate to what a user wants to do with the timelin)
- Histories and Snapshots are _written_ as streams or tables (reflecting the mechanics of serialization)

docs-src/modules/getting-started/pages/index.adoc
epinzur: I think everything should just be a table.  We will probably have tables that aren't manually loaded (like iceberg backed tables).

Also I could see a future where a table is made up both files and a stream: historical files plus a stream of new data.

therapon (approved):

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,146 @@
...
+====
+Before jumping in to writing queries in Kaskada, it's a good idea to take a minute understanding "how to think with Kaskada".
+Kaskada is built on timelines - a simple, powerful abstraction that may be different than what you're accustomed to.
+
+If you'd prefer to jump right in, scroll down to the xref:#quick-starts[] section
+====
+
+== Thinking with Kaskada's timelines
+
+Kaskada is built on the idea of a _timeline_ - the history of how a value changes over time for a specific entity or group.
```
bjchambers, line 17: "for specific entities or groups"

"specific entity" makes me think it only supports a single entity, and "group" is often synonymous with "group by key" which is basically the same as an entity... so should be plural.

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,146 @@
...
+
+****
+⭢ Read more about timelines in xref:overview:what-is-kaskada.adoc[]
+****
+
+With Kaskada, it's timelines all the way down - every operation's inputs and outputs are timelines. 
+The timeline is a flexible abstraction that can be used in different ways depending on your needs.
+In some cases, you may want to know what the timeline's value is at a specific point in time, in other cases you may want to know how the timeline's value changes over time.
+
+When making a query, you configure how to use the query's output timeline: you can use the timeline as either a _history_ or as a _snapshot_.
```
bjchambers, line 62: s/making/running or executing/

Specifically, "making" could be misinterpreted as *writing* the query. Also, maybe re-order to make more active:

`You configure how to use the query's timeline when you execute the query -- you can use it as either a _history_ or _snapshot_."

docs-src/modules/getting-started/pages/index.adoc
bjchambers, line 17: Side observation -- I think I like the use of history here, since the "history" is very similar (basically equivalent) to the history output. This also may be a good argument for why we default to that -- a timeline is (basically) the history, so we output that by default.

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,146 @@
...
+
+* A timeline *History* contains a value each time the timeline changes, and each row describes a different entity at a different point in time.
```
bjchambers, line 64: why capitalize history and snapshot here?

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,146 @@
...
+With Kaskada, it's timelines all the way down - every operation's inputs and outputs are timelines. 
+The timeline is a flexible abstraction that can be used in different ways depending on your needs.
+In some cases, you may want to know what the timeline's value is at a specific point in time, in other cases you may want to know how the timeline's value changes over time.
+
+When making a query, you configure how to use the query's output timeline: you can use the timeline as either a _history_ or as a _snapshot_.
+
+* A timeline *History* contains a value each time the timeline changes, and each row describes a different entity at a different point in time. 
+* A timeline *Snapshot* contains a value for each entity at the _same_ point in time; each row is associated with a different entity, but all rows reflect the same point in time.
+
+This output may be written in different ways -- for example, they may be written to a Parquet file or sent as events to a stream.
```
bjchambers, line 67: `This output` -> `Both history and snapshot outputs`

(specifically, clarify that we're not binding to the most recent output -- snapshot -- but both kinds)

docs-src/modules/getting-started/pages/index.adoc
```diff
@@ -1,14 +1,146 @@
...
+Kaskada can be configured to run as a remote service or as a local process.
+
+****
+⭢ Read more about installing Kaskada in xref:ROOT:installing.adoc[]
+****
+
+== Kaskada's data model
+
+Timelines begin with xref:developing:tables.adoc[tables].
+Tables are how Kaskada stores input events.
```
bjchambers, line 92: `stores` vs. `retrieves` or `accesses`? specifically, with pulsar tables it's not just storage but retrieval, right?

PR #298, opened 2023-05-01
api changes for long-lived materializations in sparrow

epinzur:


proto/kaskada/kaskada/v1alpha/common.proto
```diff
@@ -87,6 +88,9 @@ message TableConfig {
   // TODO: Consider instead defaulting to a name derived from the type of the group
   // key. This would default to allowing implicit grouping amongst tables.
   string grouping = 5;
+
+  // The backing source for the table
+  Source source = 6;
```
epinzur, line 93: this is identical to the `Source` property used `wren`'s API for creating a new table.  For stream-backed tables it would include all the connection and authentication info we need to connect to the stream.

proto/kaskada/kaskada/v1alpha/compute_service.proto
```diff
@@ -304,9 +304,68 @@ message ExecuteResponse {
...
+
+    // Minimum event time present in input.
+    int64 min_event_time = 4;
+    // Maximum event time present in input.
+    int64 max_event_time = 5;
+    // Current event time present in output.
+    int64 output_time = 6;
+
+    // The number of output rows produced so far.
+    int64 produced_output_rows = 7;
```
epinzur, line 352: I wasn't sure what data to put in here, if anything.  Mostly I was thinking this was a way to for wren to check if a sparrow instance is still running an assigned materialization.

proto/kaskada/kaskada/v1alpha/compute_service.proto
```diff
@@ -304,9 +304,68 @@ message ExecuteResponse {
...
+
+message StopMaterializationResponse{
+}
+
 service ComputeService {
   rpc Compile(CompileRequest) returns (CompileResponse);
   rpc Execute(ExecuteRequest) returns (stream ExecuteResponse);
+  rpc StartMaterialization(StartMaterializationRequest) returns (StartMaterializationResponse);
+  rpc GetMaterializationStatus(GetMaterializationStatusRequest) returns (GetMaterializationStatusResponse);
+  rpc StopMaterialization(StopMaterializationRequest) returns (StopMaterializationResponse);
```
epinzur, line 368: could maybe use a different word here? `run`? `execution` ?

```
  rpc StartExecution(StartExecutionRequest) returns (StartExecutionResponse);
  rpc GetExecution(GetExecutionStatusRequest) returns (GetExecutionResponse);
  rpc StopExecution(StopMaterializationRequest) returns (StopExecutionResponse);
```

proto/kaskada/kaskada/v1alpha/compute_service.proto
jordanrfrazier: You could move the `ProgressInformation` (residing in the `ExecuteRequest`) to a common message, then use that

epinzur: I was wondering if the existing `ProgressInformation` applies well to the streaming world? If you think it does, then I'll just pull it out to share.

epinzur: @jordanrfrazier I just made this change, and also pulled out ComputeSnapshot and ComputeSnapshotConfig at the same time, since we will probably need them later in materializations. 

Can you make the necessary sparrow changes to get this merged?

jordanrfrazier (approved):

PR #300, opened 2023-05-02
(ci) updates rust builds for PRs only to use bigger machines  

therapon:
[Build sparrow (rust) / Rust Checks](https://github.com/kaskada-ai/kaskada/actions/runs/4857535060/jobs/8658115221#logs)
succeeded now in 24m 39s

on 8 core machine


epinzur:
amazing!

epinzur (approved):

therapon:
> @therapon are we still okay with this switching sparrow to `ubuntu-latest` ?

We do not need to move to latest unless we want to. There is a way to configure machines with ubuntu 20 instead for now. Let me create the configuration first re-runniung and merging this PR

Issue #202, opened 2023-04-03
internal error while loading sample parquet file


jbellis:
Following the instructions in getting started with jupyter, on a local install:

![image](https://user-images.githubusercontent.com/42158/229601281-ec59bd67-2c35-4a39-82c6-2e3205da347a.png)

Environment is python 3.10.6 on Ubuntu 22.04

therapon:
I did try this on my Ubuntu 22.04 and it did work. 

1. What is the version of jupyter you are running `jupyter --version` on your command line should print out the version, here is mine 
```
> jupyter --version 
Selected Jupyter core packages...
IPython          : 7.34.0
ipykernel        : 6.21.2
ipywidgets       : 8.0.2
jupyter_client   : 8.0.2
jupyter_core     : 5.2.0
jupyter_server   : 2.3.0
jupyterlab       : 3.6.1
nbclient         : 0.7.2
nbconvert        : 7.2.9
nbformat         : 5.7.3
notebook         : 6.5.2
qtconsole        : 5.3.2
traitlets        : 5.9.0

```

2. The logs for the engine and manager should be under `~/.cache/kaskada` 
For each instantiation there should be a folder under this path whose name is a UUID. Hopefully there is only 1 folder :) else a `grep` on the request id that shoes in the error message would tell you which file it is. 

If you can have a look at the logs and paste here any errors that would be helpful.

jbellis:
```
$ cat ~/.cache/kaskada/logs/a0414e39-1d8e-4412-838f-88fd7b9da963/engine_stderr.txt 
/home/jonathan/.cache/kaskada/bin/engine@v0.5.4/kaskada-engine: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory
```

I manually installed libssl1.1 as suggested here https://stackoverflow.com/questions/72133316/libssl-so-1-1-cannot-open-shared-object-file-no-such-file-or-directory and that fixed it.  still, I'm curious

1. did you also get libssl1.1 on ubuntu 22.04, or is your installation not looking for it?
2. what is it that requires libssl1.1?  it looks like ubuntu 22.04 has deprecated that in favor of libssl3, which is why it requires manual installation

here is my jupyter version output
```
$ jupyter --version
Selected Jupyter core packages...
IPython          : 7.34.0
ipykernel        : 6.22.0
ipywidgets       : 8.0.6
jupyter_client   : 8.1.0
jupyter_core     : 5.3.0
[...]

PR #301, opened 2023-05-02
added file existence and accessibility checks on load

epinzur:
related to #202 

Also removed last traces of deprecated `pkg/errors` package from wren's code.

### DIFF
--- a/wren/compute/manager.go
+++ b/wren/compute/manager.go
@@ -12,7 +12,7 @@ import (
 	"time"
 
 	"github.com/google/uuid"
-	"github.com/pkg/errors"
+
 	"github.com/rs/zerolog/log"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/trace"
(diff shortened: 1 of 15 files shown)


wren/compute/manager.go
```diff
@@ -76,12 +76,14 @@ func (m *Manager) CompileQuery(ctx context.Context, owner *ent.Owner, query stri
 	subLogger := log.Ctx(ctx).With().Str("method", "manager.CompileQuery").Logger()
 	formulas, err := m.getFormulas(ctx, owner, requestViews)
 	if err != nil {
-		return nil, errors.WithMessage(err, "getting views")
+		subLogger.Error().Err(err).Msg("issue getting formulas")
+		return nil, err
```
kerinin, line 80: Is removing the error wrapper intentional?

kerinin (approved):

wren/compute/manager.go
epinzur: yes. the error wrapper package has been deprecated for quite some time and it recommended for use anymore.

PR #303, opened 2023-05-02
Track Current Session in Python Client (Closes #267)

### LINKED ISSUE #267: bug: Re-running the cell creating the session leads to race conditions
**Description**
This seems to be because the session is closed when it is garbage collected (probably good) but it uses global state (probably bad) so that instead of closing the *old* session it closes the *new* session (definitely bad).

**To Reproduce**
Have a cell in a notebook like:

```
# Create the Kaskada Session
session = LocalBuilder().build()
```

Note that without the `session =` the session is *immediately* closed because it is garbage collected right away.

Run the cell.

Rerun the cell.

**Actual Behavior**
A new session is started and immediately closed.

**Expected Behavior**
When the cell is rerun the *previous* session is closed and a new session is started.

kevinjnguyen:
# Description 

Updates the Python session management to verify the result of executing the binaries. If the binaries result in a failure, then an error message is displayed.
Screenshot 2023-05-03 at 4 41 54 PM

## Changes
* Re-running when an existing session is running, will result in a failure to spin up the new binaries due to port binding. A `RuntimeError` is thrown suggesting the user to verify the logs.
* Removes the gc for the session which enables the following:
```python
LocalBuilder().build() 
```
_Note: the session variable is not saved._

Screenshot 2023-05-03 at 4 51 06 PM


bjchambers:
I think the processes we start (for Sparrow and Wren) would die when the Python process dies too, right?

If so, I think it may be simpler to keep the session open until it is *explicitly* closed.

Then we can just have a reasonable error message if the port is not available (suggest they close the session first).

It would also mean that the user doesn't need to understand that we garbage collect things, etc.

Similarly, if we detect that there is an existing session, could we just connect to that?

I think that means we'd only need the following:

1. If we can't start a Sparrow or Wren, we produce an error (mentioning that if the port wasn't available, they may need to `close()` their previous session).
2. Optional: Track the Sparrow and Wren processes that we've started. If set, new sessions connect to that rather than restarting the processes.
3. Can remove all the GC and other logic.

epinzur:
I agree with Ben.  I think we should try to use the existing session whenever possible, since wren is storing its db in-memory.   Every time wren is restarted, the user needs to start the notebook over, re-creating tables, etc...

Maybe separately we should adjust wren to store the sqlite db on disk instead of in-memory.

kevinjnguyen:
> If we can't start a Sparrow or Wren, we produce an error (mentioning that if the port wasn't available, they may need to close() their previous session).

I went with this approach. From the Python `Popen` usage, the suggested code will wait up to 5 seconds for a return code from executing the subprocess. If the subprocess returns within the 5 seconds, then the services have failed to launch and a `RuntimeException`.

Screenshot 2023-05-03 at 4 41 54 PM

bjchambers:
> Maybe separately we should adjust wren to store the sqlite db on disk instead of in-memory.

Probably worth filing a ticket for this. Given we put the data files in something like `~/.cache/kaskada/data/*` wouldn't seem unreasonable to have a `~/.cache/kaskada/manager.db`

bjchambers:
> I went with this approach. From the Python `Popen` usage, the suggested code will wait up to 5 seconds for a return code from executing the subprocess. If the subprocess returns within the 5 seconds, then the services have failed to launch and a `RuntimeException`.

Does that mean that starting a session will *always* take 5 seconds to move on to the next cell? I wonder if instead we could do something like use the GRPC health check -- specifically, we could use this to (a) see if the services are still running on the given ports, and (b) check to see if they have started up on the given ports.

kevinjnguyen:
> Does that mean that starting a session will _always_ take 5 seconds to move on to the next cell? I wonder if instead we could do something like use the GRPC health check -- specifically, we could use this to (a) see if the services are still running on the given ports, and (b) check to see if they have started up on the given ports.

Yes, this is correct. The first run successfully runs the process and binds to the ports. The resulting services run until terminated. The proposed changes would wait up to 5 seconds for the processes to finish with a result code. Since the first run never finishes, there is no result code so its inferred the processes started and the client will attempt to connect.

The second run would try to launch and fail to bind to the ports. This usually results in the services failing to come up within 5 seconds and finish with a `1` result code. The run finished with a result code within 5 seconds, so its inferred this run failed and the error message is displayed.

kevinjnguyen:
> I wonder if instead we could do something like use the GRPC health check -- specifically, we could use this to (a) see if the services are still running on the given ports, and (b) check to see if they have started up on the given ports.

Is the suggestion here to use the gRPC healthcheck endpoint prior to starting up the process to verify if an existing session is running?
* Start new local session
  * Check to see if an existing session is running by calling the gRPC healthcheck on the manager/engine.
  * If an existing session exists, then fail to launch a new local session.
  * If no session exists, create the new local session.

kevinjnguyen:
This sounds like a good approach to me. I'll go ahead and close this PR in favor of working towards the health-check solution.

PR #311, opened 2023-05-03
Align the two getting started docs

kerinin:
This uses the same flow and code examples for both Jupyter and the CLI.

### DIFF
--- a/docs-src/modules/getting-started/pages/hello-world-cli.adoc
+++ b/docs-src/modules/getting-started/pages/hello-world-cli.adoc
@@ -85,14 +85,13 @@ You should see output similar to the following:
 10:18AM INF Success!
 ----
 
+== Loading Data into a Table
 
-== Managing resources
+Kaskada stores data in _tables_. Tables consist of multiple rows, and
+each row is a value of the same type.
+When querying Kaskada, the contents of a table are interpreted as a xref:fenl:continuity.adoc[discrete timeline]: the value associated with each event corresponds to a value in the timeline.
 
-The CLI manages Kaskada resources declaratively by managing spec files. 
-A spec file is a YAML file describing a set of Kaskada resources, for examples tables and views.
-
-We'll begin by creating a table.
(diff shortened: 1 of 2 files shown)


docs-src/modules/getting-started/pages/hello-world-cli.adoc
```diff
@@ -234,94 +226,101 @@ This makes it easier to see how a single entity changes over time.
...
 ----
-cat query.fenl | kaskada-cli query run --stdout --response-as csv
+cat query.fenl | ./kaskada-cli query run --stdout --response-as csv
 ----
+In this example, we build a pipeline of functions using the `|` character.
+We begin with the timeline produced by the table `Purchase`, then filter it to the set of times where the purchase's customer is `"patrick"` using the `xref:fenl:catalog.adoc#when[when()]` function.
 
-As you begin to better understand your data you can start using
-aggregations over your data such as the `max()` function:
+Kaskada's query language provides a rich set of operations for reasoning about time.
```
therapon, line 236: maybe link to the catalogue here, same style as "go here to find out more" that we have throughout the docs

docs-src/modules/getting-started/pages/hello-world-cli.adoc
```diff
@@ -234,94 +226,101 @@ This makes it easier to see how a single entity changes over time.
...
-time. Because values can only travel forward in time, Fenl prevents
-information about the future from "leaking" into the past.
+==== Changing how the result timeline is output
 
-Read more in the xref:fenl:fenl-quick-start.adoc[Fenl
-Language Guide]
-====
+When you make a query, the resulting timeline is interpreted in one of two ways: as a history or as a snapshot.
+
+* A timeline *History* generates a value each time the timeline changes, and each row is associated with a different entity and point in time.
```
therapon, line 281:
```suggestion
* A timeline *History* generates a value each time there is a change in the value for the entity, and each row is associated with a different entity and point in time.
```

therapon (approved):

PR #320, opened 2023-05-04
Add execution stream for pulsar 

jordanrfrazier:
Updates the iterators to streams and creates an execution input stream that implements input buffer logic with watermarks.

Next PR will hook this up to the `Scan` operation using pulsar-backed tables.


crates/sparrow-runtime/src/prepare/column_behavior.rs
```diff
@@ -0,0 +1,522 @@
...
+    /// Reference the given column.
+    Reference { index: usize, nullable: bool },
+    /// Hash the given column.
+    ///
+    /// This field is used for preparing the key column during preparation.
+    PrepareEntityKey { index: usize, nullable: bool },
+    /// Hash the given column.
+    ///
+    /// This field is used for preparing the key column during execution.
+    ExecuteEntityKey { index: usize, nullable: bool },
```
jordanrfrazier, line 42: This (and related methods) are the only addition to this file.

crates/sparrow-runtime/src/prepare/column_behavior.rs
```diff
@@ -0,0 +1,522 @@
...
+                let new_keys = arrow::compute::take(column, &indices, None)
+                    .into_report()
+                    .change_context(Error::PreparingColumn)?;
+                metadata
+                    .add_entity_keys(new_hash_keys, new_keys)
+                    .into_report()
+                    .change_context(Error::PreparingColumn)?;
+                Arc::new(entity_column)
+            }
+            ColumnBehavior::ExecuteEntityKey { index, nullable } => {
```
jordanrfrazier, line 350: Differentiated from the `PrepareEntityKey` because in prepare we use metadata files to store the key hash inverse mappings, whereas during execution we have to populate it dynamically.

crates/sparrow-runtime/src/prepare/execute_iter.rs
```diff
@@ -0,0 +1,434 @@
...
+
+/// An iterator over prepare batches. TODO
+///
+/// In addition to iterating, this is responsible for the following:
+///
+/// 1. Inserting the time column, subsort column, and key hash from source to
+///    the batch
+/// 2. Casts required columns
+/// 3. Sorts the record batches by the time column, subsort column, and key hash
+pub struct ExecuteIter<'a> {
```
jordanrfrazier, line 32: Similar to the `PrepareIter`, but includes the input buffer / late data / watermark logic. 

I'll create a `StreamReader`, similar to the `TableReader`, that the `Scan` operation uses depending on what input type it's reading from. The `StreamReader` will initialize a `pulsar_stream` as the `reader`, then create an `ExecuteIter` to read batches from.

crates/sparrow-runtime/src/prepare/execute_iter.rs
```diff
@@ -0,0 +1,434 @@
...
+            key_hash_inverse,
+            table_config: config,
+            bounded_lateness: todo!(),
+            watermark: todo!(),
+            leftovers: todo!(),
+        })
+    }
+
+    /// Convert a read batch to the merged batch format.
+    fn prepare_next_batch(
```
jordanrfrazier, line 218: The input buffer logic is in here.

crates/sparrow-runtime/src/prepare/column_behavior.rs
```diff
@@ -0,0 +1,524 @@
...
+                    }
+                );
+                // 1. Hash the current entity key column to a UInt64Array
+                let key_hashes = sparrow_kernels::hash::hash(keys)
+                    .into_report()
+                    .change_context(Error::PreparingColumn)?;
+                // 2. Update key hash table
+                futures::executor::block_on(key_hash_inverse.add(keys.clone(), &key_hashes))
+                    .into_report()
+                    .change_context(Error::Internal)?;
```
jordanrfrazier, line 368: This was not easily avoidable, afaict. Any suggestions (even if we skip doing them for now)?

crates/sparrow-runtime/src/streams/pulsar_stream.rs
```diff
@@ -193,26 +226,28 @@ impl PulsarReader {
...
-        for avro_value in &mut avro_values {
-            for (field_name, field_value) in avro_value {
-                if field_name == "_publish_time" {
-                    match field_value {
-                        Value::TimestampMillis(publish_time) => {
-                            if *publish_time < self.last_publish_time {
-                                *publish_time = self.last_publish_time;
-                            } else {
-                                self.last_publish_time = *publish_time;
+        if self.require_ordered_publish_time {
```
jordanrfrazier, line 229: Just added this flag so I can reuse the `pulsar_stream` code without pulling a lot of the logic into a separate path.

crates/sparrow-runtime/src/prepare/prepare_metadata.rs
jordanrfrazier, line 1: no changes done here, just moved it out of prepare_iter

crates/sparrow-runtime/src/execute.rs
```diff
@@ -18,16 +18,14 @@ use crate::execute::operation::OperationContext;
 use crate::s3::S3Helper;
 use crate::RuntimeOptions;
 
-mod avro_arrow;
+pub(crate) mod avro_arrow;
```
bjchambers, line 21: What do we need from this module? Should it be in `sparrow-arrow` which already has an `avro` module?

crates/sparrow-runtime/src/execute/key_hash_inverse.rs
```diff
@@ -105,7 +105,7 @@ impl KeyHashInverse {
     /// values are aligned to map from a key to a hash per index. The
     /// current implementation eagerly adds the keys and hashes to the
     /// inverse but can be optimized to perform the addition lazily.
-    fn add(&mut self, keys: ArrayRef, key_hashes: &UInt64Array) -> anyhow::Result<()> {
+    pub fn add(&mut self, keys: ArrayRef, key_hashes: &UInt64Array) -> anyhow::Result<()> {
```
bjchambers, line 108: Does this need to be public? I believe that we had this hidden because users should interact with this via the "thread-safe key hash inverse" (otherwise it's not... thread safe...)

Given the protections on mutability, it's probably *ok* to have this be public... just wanted to make sure it was necessary.

crates/sparrow-runtime/src/metadata/raw_metadata.rs
```diff
@@ -13,10 +13,10 @@ use sparrow_api::kaskada::v1alpha::source_data::{self, Source};
 use sparrow_api::kaskada::v1alpha::PulsarConfig;
 use tracing::info;
 
-use crate::execute::pulsar_schema;
 use crate::metadata::file_from_path;
 use crate::stores::object_store_url::ObjectStoreKey;
 use crate::stores::{ObjectStoreRegistry, ObjectStoreUrl};
+use crate::streams::pulsar_schema;
```
bjchambers, line 19: I wonder if we should have a pattern here of exposing a module for each "kind of stream", so this would be `crate::streams::pulsar::schema`?

crates/sparrow-runtime/src/prepare/column_behavior.rs
```diff
@@ -0,0 +1,525 @@
...
+        nullable: bool,
+    },
+    /// Perform an "order preserving" cast from a primitive number to u64.
+    OrderPreservingCastToU64 { index: usize, nullable: bool },
+    /// Reference the given column.
+    Reference { index: usize, nullable: bool },
+    /// Hash the given column.
+    ///
+    /// This field is used for preparing the key column during preparation.
+    PrepareEntityKey { index: usize, nullable: bool },
```
bjchambers, line 38: In the comments, we should explain why these are different. Otherwise, we're just saying "we have different behavior during preparation and execution, go read the code to find out what that means". Ideally, we could even name these more specifically to how they differ.

crates/sparrow-runtime/src/prepare/column_behavior.rs
```diff
@@ -0,0 +1,525 @@
...
+    /// Hash the given column.
+    ///
+    /// This field is used for preparing the key column during execution.
+    ExecuteEntityKey { index: usize, nullable: bool },
+    /// A random column
```
bjchambers, line 43: Update this comment while here? Seems to be missing some words.

crates/sparrow-runtime/src/prepare/column_behavior.rs
bjchambers: I think it may be better to return the `keys` and `key_hashes` columns, and have this happen outside this method (possibly in the stream scan operation), as noted above. That way we don't need to do the block on, etc.

crates/sparrow-runtime/src/prepare/execute_iter.rs
bjchambers: Haven't looked at this (which seems to be the main thing in this PR). Happy to approve the other pieces in a separate PR, or look at this further in my next pass.

crates/sparrow-runtime/src/prepare/execute_iter.rs
```diff
@@ -0,0 +1,536 @@
...
+///
+/// In practical terms, this allows for items in the stream to be within 1 second
+/// compared to the max timestamp read.
+///
+/// This is hard-coded for now, but could easily be made configurable as a parameter
+/// to the table.
+#[allow(unused)]
+const BOUNDED_LATENESS_NS: i64 = 1_000_000_000;
+
+/// An iterator over prepare batches.
```
bjchambers, line 34: Comments are outdated.

crates/sparrow-runtime/src/prepare/column_behavior.rs
jordanrfrazier: Hm - I could update it in the `execute` method maybe..that's the only async context. Then I can give the `Scan` operation the the `ThreadSafeKeyHashInverse` and the `key_column_name`.

crates/sparrow-runtime/src/execute/key_hash_inverse.rs
jordanrfrazier: You are correct - reverted this change.

crates/sparrow-runtime/src/execute.rs
jordanrfrazier: We convert the avro values into the arrow arrays. Yeah that's a good idea. I'll move that over

crates/sparrow-runtime/src/metadata/raw_metadata.rs
jordanrfrazier: Yeah, I figured we'd move to that pattern once we get another stream. I can go ahead and do that now

crates/sparrow-runtime/src/prepare/column_behavior.rs
jordanrfrazier: Updated

crates/sparrow-runtime/src/prepare/column_behavior.rs
jordanrfrazier: Added a few comments

crates/sparrow-runtime/src/prepare/execute_iter.rs
jordanrfrazier: Prefer looking at this in this PR, we can go over together if that's faster to follow up too

bjchambers:
I think the PR name / description need to be updated. This is still listed as a draft, which doesn't seem accurate.

crates/sparrow-main/tests/e2e/fixture/local_table.rs
```diff
@@ -82,11 +82,13 @@ impl LocalTestTable {
             //
             // TODO: Simulate the actual interaction with prepare (eg., collect raw files
             // and run prepare in response to analysis).
-            let mut iter = prepared_batches(source_data, &self.config, &None)
+            let mut prepared_iter = prepared_batches(source_data, &self.config, &None)
                 .await
                 .map_err(|e| e.into_error())?;
+            let mut iter = prepared_iter.stream();
```
bjchambers, line 88: From a naming perspective, it's not clear why we *first* create an iterator *then* change it to a stream. It seems like either (a) we should create the stream directly, possibly with the iterator as a "hidden" implementation detail or (b) we should create the iterator and read from it directly.

crates/sparrow-runtime/src/execute/output/pulsar.rs
```diff
@@ -66,9 +66,9 @@ pub(super) async fn write(
     };
 
     let topic_url = format_topic_url(&pulsar)?;
-    let output_schema =
-        pulsar_schema::get_output_schema(schema).change_context(Error::SchemaSerialization)?;
-    let formatted_schema = pulsar_schema::format_schema(output_schema.clone())
+    let output_schema = streams::pulsar::schema::get_output_schema(schema)
+        .change_context(Error::SchemaSerialization)?;
+    let formatted_schema = streams::pulsar::schema::format_schema(output_schema.clone())
```
bjchambers, line 71: What is a "formatted" schema? It seems like this could probably have a more descriptive name.

crates/sparrow-runtime/src/prepare/column_behavior.rs
bjchambers: Why is that a different column behavior? If I remember correctly -- metadata needs the key and key hash columns to update the metadata. I think instead of having the column behavior add to the metadata, we should be able to do that *after* preparing. Specifically:

1. Use the non-metadata populating column and prepare the columns.
2. The result of this should be a batch that contains a `_key_hash` column (prepared using `ExecuteEntityKey`) and a `<key>` column (for whatever the name of the underlying key was).
3. Thus, the batch can be added to the metadata *after* being prepared, by extracting those two columns.

In addition to getting rid of the need to distinguish between these two cases, it moves the maintenance of the metadata outside of the "column behavior" which seems a clearer place to have it.

crates/sparrow-runtime/src/prepare/execute_iter.rs
```diff
@@ -0,0 +1,538 @@
...
+    watermark: i64,
+    /// The leftovers are events that have been read from the stream but have not
+    /// been produced, as the watermark has not advanced far enough.
+    ///
+    /// Note this implies that if a period of inactivity in the stream occurs, we
+    /// cannot produce the last n events until the watermark advances.
+    leftovers: Option<RecordBatch>,
+}
+
+// impl<'a> Stream for ExecuteIter<'a> {
```
bjchambers, line 72: Remove the large commented out blocks?

crates/sparrow-runtime/src/prepare/execute_iter.rs
```diff
@@ -0,0 +1,538 @@
...
+            debug_assert!(self.leftovers.is_none());
+            arrow::compute::concat_batches(&self.prepared_schema, &[leftovers, record_batch])
+                .into_report()
+                .change_context(Error::PreparingColumn)?
+        } else {
+            // The only time there should be no leftovers is the first batch, but we still need to handle this case.
+            record_batch
+        };
+
+        // 5. Pull out the time, subsort, and key hash columns to sort the record batch
```
bjchambers, line 306: I wonder if this could be implemented as a wrapper around the existing prepare iterator, rather than as an alternative. What I'm picturing:

1. Run prepare to produce batches
2. Use a custom iterator, or something like `fold` to maintain state, and produce batches using the logic here.

The benefit is that (1) we wouldn't duplicate the prepare logic -- only one place to fix bugs (2) it would be easier to understand what *this* does different.

crates/sparrow-runtime/src/prepare/column_behavior.rs
jordanrfrazier: Yeah, I noted that in the comment 😢 . I can make that change.

crates/sparrow-runtime/src/prepare/execute_iter.rs
jordanrfrazier: The `execute_iter` contains the input buffer logic. Currently, part of that logic is "dropping late data", where late data is defined as "data that arrives with an event time prior to the current watermark". The watermark is defined as the current largest event time minus an allowed lateness delta. 

The problem if we use `prepare` then fold over batches produced by that iterator is we lose _ordering_, and we cannot tell if we should've dropped data. For example, if the following messages are read by the stream:
```
(allowed lateness=5)
[0, 2, 3, 5, 7, 10] - no late data, don't drop any
[0, 10, 2, 3, 5, 7] - drop [2, 3]. 
```
But since `prepare` sorts the batch of messages, the logic around the iterator would only see the ordered batch, and not know it should've dropped `[2, 3`. This is a problem because we then get non-deterministic results based on batch size.

jordanrfrazier: I think even once we're iterated on how we handle late data and have the ability to walk back entities, we'll still want a cutoff time that drops late data. For example, if we're reading events in 2023, and we see an event from 2022, it likely is no longer relevant for making a prediction right now, and we shouldn't re-process a year's worth of events for such an outlier.

crates/sparrow-main/tests/e2e/fixture/local_table.rs
jordanrfrazier: Hm, yeah. It was left as is since that's how prepare used it, but I can make a change to make this clearer now that we're using async streams.

jordanrfrazier: Okay made this change for the `prepare_iter` -> `prepare_input_stream`, let me know what you think and I'll do the same for the `execute_iter` in that case.

bjchambers: That works for me.

crates/sparrow-runtime/src/prepare/execute_iter.rs
bjchambers: That's a good catch. What it seems like is we actually have three things happening -- we probably don't need to implement it now, but:

1. The underlying *reader*. Reading batches from Parquet, or reading rows from Pulsar and assembling batches. This is probably where the late data handling *should* happen. Maybe watermark too.
2. The general *preparation*. This could be a map or fold over the batches from the reader. In this case, it would do things like prepare the columns, sort things batches if not done by the reader, etc.
3. Some special stuff. Not sure we'd still need execute/prepare to be separate.

But, I think you're argument is convincing for why handling late data / watermarks needs to happen *before* the more general things.

crates/sparrow-runtime/src/prepare/prepare_input_stream.rs
```diff
@@ -0,0 +1,144 @@
...
+/// An stream over prepare batches and corresponding key hash metadata.
+///
+/// This stream reads unordered, unprepared batches from the `reader`, and is
+/// responsible for:
+/// 1. Inserting the time column, subsort column, and key hash from source to
+///    the batch
+/// 2. Casting required columns
+/// 3. Sorting the record batches by the time column, subsort column, and key hash
+/// 4. Computing the key-hash and key batch metadata.
+pub async fn try_new<'a>(
```
bjchambers, line 29: Would suggest calling this something different -- like `prepare_input(...)`. Basically:

1. We no longer need a `struct` so there isn't a type, so `try_new` is less appropriate.
2. The type reflects this is a stream.

Then, we can write `crate::prepare::prepare_input(reader_stream, ...)` which would be pretty clear.

crates/sparrow-runtime/src/prepare/execute_iter.rs
jordanrfrazier: Added a comment for this

crates/sparrow-runtime/src/execute/output/pulsar.rs
jordanrfrazier: Yeah, this is the `avro_schema` actually. Changed

crates/sparrow-runtime/src/prepare/column_behavior.rs
jordanrfrazier, line 1: No changes in this file anymore

crates/sparrow-runtime/src/prepare/execute_input_stream.rs
```diff
@@ -0,0 +1,468 @@
...
+                .change_context(Error::PreparingColumn)?;
+
+            // 2. Slicing may reduce the number of entities to operate and sort on.
+            let record_batch = slice_preparer.slice_batch(record_batch)?;
+
+            // 3. Prepare each of the columns by getting the column behavior result
+            let mut prepared_columns: Vec<ArrayRef> = Vec::new();
+            for c in columns.iter_mut() {
+                let result = c.get_result(&record_batch).await?;
+                update_key_inverse(c, &record_batch, &result, key_hash_inverse.clone()).await?;
```
bjchambers, line 195: This seems wrong -- isn't this updating the key hash inverse *for every column*?

I think we'd want something like the following, but outside of the loop:

```
let mut prepared_columns: Vec<ArrayRef> = Vec::new();
let key_column = prepared_columns.get_index(key_column);
let key_hash_column= prepared_columns.get_index(key_hash_column);
update_key_hash_inverse(...)
```

crates/sparrow-runtime/src/prepare/prepare_input_stream.rs
```diff
@@ -0,0 +1,280 @@
...
+            // 1. Slicing may reduce the number of entities to operate and sort on.
+            let read_batch = slice_preparer.slice_batch(batch)?;
+            // 2. Prepare each of the columns by getting the column behavior result
+            let mut prepared_columns = Vec::new();
+            for c in columns.iter_mut() {
+                let result = c
+                    .get_result(&read_batch)
+                    .await?;
+
+                update_key_metadata(c, &read_batch, &result, &mut metadata)?;
```
bjchambers, line 102: Same comment here... we shouldn't update key metadata for all columns... only the key hash column.

bjchambers (approved):

crates/sparrow-runtime/src/prepare/execute_input_stream.rs
jordanrfrazier: How it currently is is mimicking what was there - we're matching on the column behavior and updating when we see the `ColumnBehavior::EntityKey`. I suppose we could change it, since the only metadata that's added is from the key

crates/sparrow-runtime/src/prepare/prepare_input_stream.rs
jordanrfrazier: same

crates/sparrow-runtime/src/prepare/execute_input_stream.rs
bjchambers: I think that would feel a lot more straightforward. Even if we don't use the name/index to look it up, it would be clearer if that happened outside the loop. This makes it really easy to read this as updating on every column.

jordanrfrazier: Cool, yeah. Moved this out

Issue #318, opened 2023-05-04
bug: Columns dropped from Parquet


kerinin:
**Description**
When I create a table and load the attached Parquet file, the `transfer_value` column is not present in the loaded data.

Here's the file's contents
```python
import pandas as pd
pd.read_parquet("nft_df.parquet")
```
This displays the file's contents - includes a final column named `transfer_value`

```
                                          entity_id  transfer_block_number                            contract_address token_id       transfer_value
0    0xcb45dce241a6f6f1e874e9925bea5840f3fd7d3a|314               17183150  0xcb45dce241a6f6f1e874e9925bea5840f3fd7d3a      314  1041900000000000000
1   0xcb45dce241a6f6f1e874e9925bea5840f3fd7d3a|2095               17181436  0xcb45dce241a6f6f1e874e9925bea5840f3fd7d3a     2095                    0
2    0xcb45dce241a6f6f1e874e9925bea5840f3fd7d3a|271               17180562  0xcb45dce241a6f6f1e874e9925bea5840f3fd7d3a      271                    0
3    0xcb45dce241a6f6f1e874e9925bea5840f3fd7d3a|493               17174850  0xcb45dce241a6f6f1e874e9925bea5840f3fd7d3a      493  1037900000000000000
4   0xcb45dce241a6f6f1e874e9925bea5840f3fd7d3a|1538               17173914  0xcb45dce241a6f6f1e874e9925bea5840f3fd7d3a     1538                    0
..                                              ...                    ...                                         ...      ...                  ...
95  0xdbc61a1685c6f70c11cbb7da70338352b1fa4d51|7059               17180319  0xdbc61a1685c6f70c11cbb7da70338352b1fa4d51     7059                    0
96  0xdbc61a1685c6f70c11cbb7da70338352b1fa4d51|1357               17180319  0xdbc61a1685c6f70c11cbb7da70338352b1fa4d51     1357                    0
97  0xdbc61a1685c6f70c11cbb7da70338352b1fa4d51|1335               17180319  0xdbc61a1685c6f70c11cbb7da70338352b1fa4d51     1335                    0
98   0xdbc61a1685c6f70c11cbb7da70338352b1fa4d51|767               17178469  0xdbc61a1685c6f70c11cbb7da70338352b1fa4d51      767                    0
[...]

PR #323, opened 2023-05-05
draft: Fail Unsupported Columns (#318) 

kevinjnguyen:
# Fail Unsupported Columns (#318)

Updates the `FileService` and `PrepareService` to fail for unsupported columns (Decimal) rather than silently drop the columns.


crates/sparrow-runtime/src/metadata/raw_metadata.rs
```diff
@@ -246,36 +248,43 @@ impl RawMetadata {
...
+fn convert_field(field: &Field) -> error_stack::Result<Field, Error> {
+    match field.data_type() {
+        DataType::Timestamp(time_unit, Some(tz)) => {
+            // TODO: We discard this because the conversion from an Arrow
+            // schema to the Schema protobuf currently fails on such timestamp columns.
+            info!(
+                "Discarding time zone {:?} on timestamp column '{}'",
+                tz,
+                field.name()
+            );
```
jordanrfrazier, line 270: I'd make this `tracing::warn!("Time zones are unsupported. Interpreting column '{}' with time zone '{}' as UTC")`

crates/sparrow-runtime/src/metadata/raw_metadata.rs
```diff
@@ -246,36 +248,43 @@ impl RawMetadata {
...
+            Ok(Field::new(
+                field.name(),
+                DataType::Timestamp(time_unit.clone(), None),
+                field.is_nullable(),
+            ))
+        }
+        DataType::Decimal128(_, _) | DataType::Decimal256(_, _) => {
+            // TODO: Support decimal columns (fix)
+            info!("Discarding decimal column '{}'", field.name());
```
jordanrfrazier, line 279: Same, make this a warning, say unsupported

crates/sparrow-runtime/src/prepare.rs
```diff
@@ -426,7 +426,8 @@ fn reader_from_parquet<'a, R: parquet::file::reader::ChunkReader + 'static>(
     let batch_size = get_batch_size(num_rows, num_files);
     let parquet_reader = parquet_reader.with_batch_size(batch_size);
 
-    let raw_metadata = RawMetadata::from_raw_schema(parquet_reader.schema().clone());
+    let raw_metadata = RawMetadata::from_raw_schema(parquet_reader.schema().clone())
+        .change_context_lazy(|| Error::DetermineMetadata)?;
```
jordanrfrazier, line 430: Should this be `Error::ReadSchema`? same for below

jordanrfrazier: same for below

jordanrfrazier: i.e. `DetermineMetadata` doesn't make much sense from a user's POV. The cause is from an invalid schema

jordanrfrazier (approved):
The sparrow changes look good to me. Whether this actually produces the error in the client-side, I'm not sure. I'd add an integration test and try that quickly (which you can do easily now with Eric's changes).

PR #340, opened 2023-05-08
feat: add fenlmagic extension loaded message (#305)

### LINKED ISSUE #305: feat: Return some message to tell the user that the fenlmagic extension loaded successfully
**Summary**
I'd like the fenlmagic extension to output something when it loads successfully.

```
%load_ext fenlmagic
```
(currently returns nothing)

**Is your feature request related to a problem? Please describe.**
I watched a user get very confused when there was no output after loading the extension. I think there will be other users that feel the same.

**Use Case**
Reduce confusion for new users of jupyter & kaskada.

**Describe the solution you'd like**
on succesful load of the extension, return a response like `fenlmagic extension loaded.` 

**Describe alternatives you've considered**

kevinjnguyen:
Closes #305.

Screenshot 2023-05-08 at 5 45 43 PM

### DIFF
--- a/clients/python/src/fenlmagic/__init__.py
+++ b/clients/python/src/fenlmagic/__init__.py
@@ -6,7 +6,9 @@
 # pylint: disable=no-member
 from __future__ import print_function
 
+import logging
 import os
+import sys
 
 import IPython
 import pandas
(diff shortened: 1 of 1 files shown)


clients/python/src/fenlmagic/__init__.py
```diff
@@ -40,6 +45,7 @@ class FenlMagics(Magics):
     def __init__(self, shell, client):
         super(FenlMagics, self).__init__(shell)
         self.client = client
+        logger.info("fenlmagic extension loaded")
```
epinzur, line 48: maybe change to just `"extension loaded"` so that the output reads without the repeated "fenlmagic"?

epinzur (approved):

PR #341, opened 2023-05-08
Update the Dockerfile Notebook Version to 0.1.7

kevinjnguyen:
Updates the Kaskada client to version 0.1.7 in the dockerfile notebook.

Q: Do we need to push a new image somehow to ghcr?

### DIFF
--- a/Dockerfile.notebook
+++ b/Dockerfile.notebook
@@ -18,5 +18,4 @@ WORKDIR ${HOME}
 RUN pip install pulsar-client
 
 # Install the kaskada python client
-RUN pip install kaskada==0.1.1a9
-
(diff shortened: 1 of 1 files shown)


epinzur:
I don't think we are using this dockerfile anymore externally.  Maybe it is better to just delete it?

kerinin:
I definitely don't want to be constantly updating hard-coded library versions. Can the be folded into the jupyter environment Eric has on a branch and all the hard-coded versions removed?

kevinjnguyen:
Sure. There are currently at least 6 dockerfiles in the root of the repository and maybe more if we are building docker images from other branches. I'll file another issue and close out this PR.

PR #345, opened 2023-05-09
(docs) fix integrations menu

therapon:
* make `integrations` menu item a link to the index page 
* fix unclosed example block


kevinjnguyen (approved):

docs-src/modules/integrating/nav.adoc
```diff
@@ -1,4 +1,4 @@
-* xref:index.adoc[]
+* xref:index.adoc[Integrations]
```
bjchambers, line 1: I don't think this is necessary.

> When Antora converts the xref macro and its value into a link, the text entered between the square brackets is displayed in the current page when it’s published to your site. If no link text is specified between the macro’s square brackets, and the value of the xref macro doesn’t contain a fragment, then the target page’s [default reference text](https://docs.antora.org/antora/latest/page/link-content/#default) or text assigned to the [reftext attribute](https://docs.antora.org/antora/latest/page/reftext-and-navtitle/) is displayed in the current page when it’s published.

Further, note that even going to the new `index` doesn't work: https://kaskada.io/docs-site/kaskada/main/integrating/index.html. If the problem was related to the nav, then I would expect it to only affect the menu, not the actual page.

bjchambers: See also https://raw.githubusercontent.com/kaskada-ai/kaskada/main/docs-src/modules/getting-started/nav.adoc which works.

PR #355, opened 2023-05-10
Update Default Data Location to User Directory Cache (#253)

### LINKED ISSUE #253: feat: Default to storing data in `~/.cache/data`, rather than `data`
**Summary**
The current data location is often user visible, and users interpret as part of the system's public API. We should move it to an internal location so it's clear that the contents aren't intended to be poked at.

kevinjnguyen:
Closes #253.


therapon (approved):

epinzur:
can we also move the flag`db-path` in wren also to the new path by default.

epinzur (approved):

Issue #342, opened 2023-05-09
feat: remove redundant or unused dockerfiles


kevinjnguyen:
**Summary**
In the repository now, there are currently 6 dockerfiles that may or may not be used. There may also be additional dockerfiles used in other branches to build images for specific use cases that need to be maintained.

We should consolidate our use of docker into a single location rather than single use files.

**Is your feature request related to a problem? Please describe.**
As we upgrade our dependencies, we need to upgrade the docker images to use the latest binaries/clients but without a centralized location or knowledge of where they live, the images quickly become obsolete.

**Use Case**
A single location of dockerfiles that are used/expected to maintain even by the pipeline.

**Describe the solution you'd like**
Single use for POC or demos should be removed or on a separate branch.

**Describe alternatives you've considered**
N/A

PR #357, opened 2023-05-12
cleaned up dockerfiles and added jupyter build

epinzur:
removed unused `Dockerfile.notebook`

trying to see if we can use `Dockerfile.release` in place of `Dockerfile.integration` since they are so similar.

Added new `Dockerfile.jupyter` to create a mult-arch release of our system based off a Jupyter notebook environment with pre-installed kaskada services and clients.


Dockerfile.jupyter
```diff
@@ -0,0 +1,31 @@
...
+
+# Install the pulsar python client and the kaskada python client
+RUN pip install pulsar-client kaskada --no-cache-dir
+
+RUN mkdir -p /home/jovyan/.cache/kaskada/bin
+COPY --chmod=755 ./release/${TARGETOS}/${TARGETARCH}/cli            /home/jovyan/.cache/kaskada/bin/kaskada-cli
+COPY --chmod=755 ./release/${TARGETOS}/${TARGETARCH}/wren           /home/jovyan/.cache/kaskada/bin/kaskada-manager
+COPY --chmod=755 ./release/${TARGETOS}/${TARGETARCH}/sparrow-main   /home/jovyan/.cache/kaskada/bin/kaskada-engine
+
+ENV KASKADA_DISABLE_DOWNLOAD=true
```
epinzur, line 31: should look very similar to `Dockerfile.release`

Dockerfile.jupyter.dockerignore
```diff
@@ -0,0 +1,7 @@
+*
+!release/linux/amd64/sparrow-main
+!release/linux/amd64/wren
+!release/linux/amd64/cli
+!release/linux/arm64/sparrow-main
+!release/linux/arm64/wren
+!release/linux/arm64/cli
```
epinzur, line 7: should look very similar to `Dockerfile.release.dockerignore`

epinzur:
closes #342

.github/workflows/ci_engine.yml
```diff
@@ -66,14 +66,14 @@ jobs:
...
-          path: ci-bin/wren/
+          name: manager-binary
+          path: release/linux/amd64/
 
       - name: Get Sparrow binary
         uses: actions/download-artifact@v3
         with:
           name: engine-debug
-          path: ci-bin/engine/
+          path: release/linux/amd64/
```
epinzur, line 76: i realize it is a bit odd copying a debug build to a release path, but this works.  Maybe in the future we can drop the `release` copy path in `Dockerfile.release`

.github/workflows/release_engine.yml
```diff
@@ -289,6 +290,24 @@ jobs:
             org.opencontainers.image.revision=${{ github.sha }}
             org.opencontainers.image.licenses=Apache-2.0
 
+      - name: Build and push Docker images for Jupyter
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          platforms: linux/amd64, linux/arm64
+          file: Dockerfile.release
```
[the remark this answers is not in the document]
epinzur: yes, good catch!

therapon (approved):

Issue #256, opened 2023-04-20
feat: Render Wren's logs better


kerinin:
**Summary**
@stinkymatt has been running Kaskada from Python and the log output is hard to read, for example newlines are ignored. This may be related to log coloring.

Options that have been discussed:
* Don't use colored logs
* Use JSON logging

bjchambers:
This may also apply to the Sparrow logs.

bjchambers:
If we went the route of JSON logging, it could also be a path towards better interactions with the logs. For instance, we could use `jq` to filter to the logs that have a given `request_id`, etc.

therapon:
+1 on JSON 

It seems to be the common format used for structured logging which will allow easy ingestion into existing tools (Sumo comes to mind)

epinzur:
Both wren and sparrow are capable of producing json logs, although I think they output in different formats.

However, I generally think text logs are easier for humans to read.  

Do we think the wren & sparrow logs in the notebook will be more often read by humans or other processing systems/tools?

I think out decisions here should be based on the answer to this question.

kerinin:
Our logs as they exist today aren't very human-friendly, but I mostly find myself using them to try to diagnose a failures, either by looking at the most recent logs (local execution) or searching for a request ID (remote execution). 

I can't think of any ways that I would setup an automated system to consume the logs, although some of the tools we've used in the past for interacting over logs in bulk took advantage of structured logging to allow querying logs in a richer way.

PR #358, opened 2023-05-12
updated logging formats of both processes to be more similar

epinzur:
Added flags and `NO_COLOR` environment variable support to both wren & sparrow to disable ANSI color args in log output when set.

see: https://no-color.org/

Also 
* Disabled wren logging database calls by default in debug mode.  Added new flag to re-enable when desired.
* Updated python client to disable color log output for services


epinzur:
fixes #256

PR #370, opened 2023-05-22
added mock generation to wren

epinzur:
Also updated existing tests to use the new mocks, so that we don't need to manually update the mocks in tests when interfaces change.

this is mostly prep for upcoming changes related to materializations


kevinjnguyen:
This works for me. Do we intend to keep the mocks as part of our source control or was this just to get it to run? Usually generated code should not be committed as part of code.

kevinjnguyen (approved):

Issue #254, opened 2023-04-20
feat: Improve Python's subprocess management


kerinin:
**Summary**
The way that the Python library manages the kaskada subprocesses could be improved in a couple ways:

* Raise an exception in Python if the subprocess can’t be started (ie if it can’t bind the port b/c another process is still running), * Raise an exception if the subprocess exits prematurely
* (maybe?) Create the subprocess on-demand if it has been terminated, for example by IPython

The last point comes out of some issues @stinkymatt was having where the process seems to be terminating for no reason.

epinzur:
probably should move wren db storage to be local and not in-memory

kevinjnguyen:
The latest PR #373 iterates on the next implementation of the Local Session to automatically recover from local session failure. Please take a look at this [example notebook](https://github.com/kaskada-ai/kaskada/blob/main/examples/Local%20Session%20(Next%20Iteration).ipynb) for a documented feature rundown.

PR #373, opened 2023-05-23
feat: Kaskada Health Checks + Auto Recovery for Python (#254 / #267) 

kevinjnguyen:
# Description

Adds health checks to the Python client and improves overall code syntax.

## Changes

* Adds `grpcio-health-checking` to the Python `pyproject.tml`.
* Updates the pylint max line to 200 (easier to read)
* Replaces `api_utils.run_subprocess` with a `KaskadaLocalService` and a `SubprocessFactory`.
* Replaces `api_utils.check_socket` with an actual health check.
* Updated subprocess runs to synchronously terminate with the `stop()` method.
* Add a keep alive watcher to automatically recover from failure.
* Update the connection retry logic from the session to the client.
* Create a `LocalSession` object to wrap around the Kaskada services to enable restarting.
* Updates client docs and organization
* Adds a `HealthCheckClient`.
* Adds a `HealthCheckServicer`.
* Adds a `HealthCheckWatcher`.


epinzur:
This looks wonderful.  

I do want to ask about remote services though.  Are they still supported? 
* I saw that you can still specify an endpoint when creating a client, so it seems that it should still be supported, but I couldn't exactly tell from the code.
* I also saw that there is a new `should_connect` param when creating a client.  Maybe that should be named `is_local`, or something similar?  The heath-check endpoints for manager and engine are unlikely to be available on remote running services.

epinzur (approved):

kevinjnguyen:
> I do want to ask about remote services though. Are they still supported?

In theory, remote services are supported. Remote services will need to expose the gRPC health check endpoint for the Engine and Manager. The Engine uses the same port (50052) and the Manager uses (6666). At the moment, they default the local host versions. We will need to revise this once a remote implementation comes up and running. There will be some configuration but the `RemoteSession` is a future implementation.

> I saw that you can still specify an endpoint when creating a client, so it seems that it should still be supported, but I couldn't exactly tell from the code.

The client will attempt to use the defaults which are localhost for the Engine and Manager. In the future, we will need to customize this.

> I also saw that there is a new should_connect param when creating a client. Maybe that should be named is_local, or something similar? The heath-check endpoints for manager and engine are unlikely to be available on remote running services.

The remote services use case is an interesting story since it differs from our normal local session quite differently. In a traditional microservice, they expose out the health services publicly to ensure clients check the status. Are they usually only for internal load balancers and not externally accessible? That'd be a consideration for changing this. But I would like to defer the change until needed for remote sessions.

The original intent of `should_connect` was to allow me to write a unit test around the client without it attempting to connect.

epinzur:
>  In a traditional microservice, they expose out the health services publicly to ensure clients check the status. Are they usually only for internal load balancers and not externally accessible? That'd be a consideration for changing this. But I would like to defer the change until needed for remote sessions.

Yes, health checks are almost never externally available.  They are generally only used by internal services like kubernetes pod health checks and load balancers.

When we used to host wren externally, we would only expose the 50051 service outside of our kubernetes environment. Sparrow was not exposed externally at at all. Any service that is exposed externally needs to implement authentication and authorization, so many services like sparrow are generally never exposed.

Issue #280, opened 2023-04-25
feat: Improve user experience when a query doesn't produce a record.


kerinin:
**Summary**
Currently, when a user runs a simple query like `count(Foo)`, an error is returned to the effect of "output type must be a record". This is the result of a query returning a non-record type, which isn't currently supported. This error is confusing to new users who don't know how to construct records, and aren't accustomed to this type of restriction in other systems.

Improve the error message to suggest a remedy (ie, link to a docs section explaining records, or explain more clearly the problem and how to solve it).

(As noted below, changes to support arbitrary types are pare of #273)

bjchambers:
This seems like a duplicate of #273

kerinin:
If we want to go that route I think you're right, but there's still the possibility of just changing the error message. We could also stage these - change message now while the #273 is pending.

PR #385, opened 2023-05-25
Fenl Diagnostic HTML Rendering with Protos (#280) (#376)

kevinjnguyen:
# Fenl Diagnostic HTML Rendering with Protos 

A second pass at implementing rendering of Fenl Diagnostics from the Engine upward. The UX is the same but now the source of truth is the Engine.

## Changes

* Removes the custom python client approach
* Updates fenl diagnostic to have a `web_link` as another field
* Updates python client rendering to use the `web_link` field


kevinjnguyen:
@therapon as a follow up, maybe we can work on documentation together. What are your thoughts on the documentation steps?

The proposed changes here will automatically render the error code with the web link if they are provided in the fenl diagnostic proto message. Meaning if we add a web link to the Engine error definition, then it'll automatically render the link. Like:
```rust
InvalidOutputType(E0013, Error, "Invalid output type", "https://kaskada.io/docs-site/kaskada/main/fenl/fenl-diagnostic-codes.html#e0013"),
```

We need to create the documentation properly first and then link them to the error code.

You also mentioned not rendering all the errors on a single page. It may be helpful to pair on the Antora document structure.

kevinjnguyen:
@cla-bot check

kevinjnguyen:
@cla-bot check

PR #386, opened 2023-05-25
refactored the compute package into several interfaces

epinzur:
now have:
* CompileManager
* ComputeManager
* FileManager
* PrepareManager
* MaterializationManager


kerinin (approved):
LGTM, although I don't have a ton of familiarity with the code at this point.

PR #387, opened 2023-05-25
added long-running materialization support to wren

epinzur:


wren/compute/materialization_manager.go
```diff
@@ -110,6 +119,77 @@ func (m *materializationManager) GetMaterializationStatus(ctx context.Context, m
...
+
+		isRunning := false
+		// check to see if the materialization was running in the previous iteration
+		if _, found := m.runningMaterializations[materializationID]; found {
+			//verify that the materialization is still running
+			progressInfo, err := m.GetMaterializationStatus(ctx, materializationID)
+			if err != nil {
+				log.Error().Err(err).Str("id", materializationID).Msg("failed to get materialization status")
+			}
+			isRunning = progressInfo != nil
```
jordanrfrazier, line 149: Let's make a note to look back at this logic later once Sparrow is done. We may want the ability to `pause` a materialization, or it may have `failed` and we want to report the failure reason (but we'll still want the progress to see what's been produced up til failure). 

This probably leads me to think we'll have a `MaterializationStatusResponse` that wraps the `ProgressInfo` and includes some materialization-specific state as well.

wren/service/materialization.go
```diff
@@ -181,6 +189,26 @@ func (s *materializationService) createMaterialization(ctx context.Context, owne
...
+			newSourceType = materialization.SourceTypeStreams
+		default:
+			log.Error().Msgf("unknown source type %T", table.Source.Source)
+			return nil, customerrors.NewInternalError("unknown table source type")
+		}
+		if sourceType == materialization.SourceTypeUnspecified {
+			sourceType = newSourceType
+		} else if sourceType != newSourceType {
+			return nil, customerrors.NewInvalidArgumentErrorWithCustomText("cannot materialize tables from different source types")
+		}
```
jordanrfrazier, line 209: Another good spot to remind ourselves to update once we can read directly from files

jordanrfrazier (approved):

wren/main.go
```diff
@@ -351,6 +351,18 @@ func main() {
 		return nil
 	})
 
+	// peridocally reconcile materializations to ensure the ones that are supposed to be running are running
+	g.Go(func() error {
+		for {
+			time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
epinzur: changed to once a minute.  really we should probably tie it into wren watching the sparrow health-check, and just do it after a sparrow restart.

wren/service/materialization.go
epinzur: lots of changes when we can read directly from files... would rather not leave todo's through the codebase for each changed required.

PR #388, opened 2023-05-26
feat: materialization API sparrow impl

jordanrfrazier:


crates/sparrow-runtime/src/execute.rs
```diff
@@ -236,3 +240,128 @@ pub async fn execute(
         request.compute_snapshot_config,
     ))
 }
+
+pub async fn materialize(
```
jordanrfrazier, line 244: Some of the setup code is similar to the `execute` method above, and could be refactored.

therapon:
@cla-bot check

crates/sparrow-main/src/serve/compute_service.rs
```diff
@@ -221,6 +277,27 @@ async fn execute_impl(
         .map(|item| item.into_status()))
 }
 
+fn materialize_impl(
```
bjchambers, line 280: This should be renamed -- it doesn't actually do the materializing, it just creates one.
It looks like the reason the materializations are in an `Arc` is so they can be passed here, but that is only used for inserting in the map. I'd suggest instead having this return the `control_handle` and then having the caller insert into the materializations. This will in turn mean we don't need to clone the materializations to pass them in here.

Also, not sure splitting this method out at all benefits that much, but your call.

crates/sparrow-materialize/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod materialize;
```
bjchambers: Why is this a public module and not just stored here? This means that logically, people use `sparrow_materialize::materialize::Materialization`, which seems to stutter (eg., what else is going to be in `sparrow_materialize` that isn't also in `materialize`)?

We could maybe have a `materialization.rs` and a `materialization_control.rs` and then expose both of them with `pub use`? Other thoughts?

crates/sparrow-materialize/src/materialize.rs
```diff
@@ -0,0 +1,235 @@
...
+    ) -> Self {
+        Self {
+            id: id.clone(),
+            plan,
+            tables,
+            destination,
+        }
+    }
+
+    // TODO: Move some of the `start` over to here, such that this is for a standalone process
```
bjchambers: . (just capturing this todo)

crates/sparrow-materialize/src/materialize.rs
```diff
@@ -0,0 +1,235 @@
...
+    // that I can call and not consider when trying to manage handles, threads.
+}
+
+pub struct MaterializationControl {
```
bjchambers, line 71: Rustdoc on the struct

crates/sparrow-materialize/src/materialize.rs
```diff
@@ -0,0 +1,235 @@
...
+    CreateMaterialization,
+    #[display(fmt = "failed to read progress")]
+    ReadingProgress,
+    #[display(fmt = "internal error")]
+    Internal,
+}
+
+impl error_stack::Context for Error {}
+
+pub struct Materialization {
```
bjchambers, line 24: Rustdoc on the struct. What is it and how does it relate to materialization control, etc.

crates/sparrow-materialize/src/materialize.rs
```diff
@@ -0,0 +1,235 @@
...
+    pub id: String,
+    pub plan: ComputePlan,
+    pub tables: Vec<ComputeTable>,
+    pub destination: Destination,
+}
+
+impl Materialization {
+    pub fn new(
```
bjchambers, line 32: Rustdoc on public methods please.

crates/sparrow-materialize/src/materialize.rs
```diff
@@ -0,0 +1,235 @@
...
+        let (stop_tx, _stop_rx) = tokio::sync::watch::channel(false);
+        let (progress_tx, progress_rx) =
+            tokio::sync::watch::channel(MaterializationProgress::Uninitialized);
+
+        let handle = tokio::spawn(async move {
+            let plan = materialization.plan.clone();
+            let destination = materialization.destination.clone();
+            let tables = materialization.tables.clone();
+
+            // TODO: pass stop signal receiver
```
bjchambers: .

crates/sparrow-materialize/src/materialize.rs
```diff
@@ -0,0 +1,235 @@
...
+            let mut progress_stream = sparrow_runtime::execute::materialize(
+                plan,
+                destination,
+                tables,
+                s3_helper,
+                bounded_lateness_ns,
+            )
+            .await
+            .change_context(Error::CreateMaterialization)
+            .expect("TODO") // TODO: Report error
```
bjchambers: .

crates/sparrow-materialize/src/materialize.rs
```diff
@@ -0,0 +1,235 @@
...
+            while let Some(message) = progress_stream
+                .try_next()
+                .await
+                .change_context(Error::ReadingProgress)?
+            {
+                if let Some(progress) = message.progress {
+                    last_progress = progress.clone();
+                    progress_tx
+                        .send(MaterializationProgress::Running(progress))
+                        .expect("TODO");
```
bjchambers, line 109: .

crates/sparrow-materialize/src/materialize.rs
```diff
@@ -0,0 +1,235 @@
...
+                }
+            }
+
+            progress_tx
+                .send(MaterializationProgress::Stopped(last_progress))
+                .expect("TODO");
```
bjchambers, line 115: .

crates/sparrow-materialize/src/materialize.rs
```diff
@@ -0,0 +1,235 @@
...
+    pub async fn stop(&mut self) -> error_stack::Result<(), Error> {
+        self.stop_signal_sender
+            .send(true)
+            .into_report()
+            .change_context(Error::Internal)?;
+
+        let handle = self.handle.take().expect("handle");
+        match &self.status {
+            MaterializationProgress::Uninitialized => {
+                // TODO: It's possible that the materialization process has started
```
bjchambers, line 139: .

bjchambers: good catch -- I guess always returning progress at the end would be a bit more correct? but probably OK either way for now.

crates/sparrow-materialize/src/materialize.rs
```diff
@@ -0,0 +1,235 @@
...
+                tracing::info!("cannot stop; materialization already failed")
+            }
+        };
+
+        Ok(())
+    }
+
+    pub fn get_status(&self) -> MaterializationStatus {
+        let progress = self.progress_receiver.borrow();
+        match &*progress {
```
bjchambers, line 188: ? weird pattern

crates/sparrow-materialize/src/materialize.rs
```diff
@@ -0,0 +1,235 @@
...
+            MaterializationProgress::Running(progress) => MaterializationStatus {
+                id: self.id.clone(),
+                progress: progress.clone(),
+                state: State::Running,
+                error: None,
+            },
+            MaterializationProgress::Stopped(progress) => MaterializationStatus {
```
bjchambers, line 201: ```MaterializationStatus { state: State::Stopped, ..previous }``` can be helpful for some of these to reduce repetition.

crates/sparrow-main/src/serve/compute_service.rs
```diff
@@ -14,28 +17,34 @@ use sparrow_api::kaskada::v1alpha::{
...
 
 use crate::serve::error_status::IntoStatus;
 use crate::BuildInfo;
 
-#[derive(Debug)]
 pub(super) struct ComputeServiceImpl {
     flight_record_path: &'static Option<S3Object>,
     s3_helper: S3Helper,
+    /// Thread-safe map containing the materialization id to control handles.
+    materializations: Arc<DashMap<String, MaterializationControl>>,
```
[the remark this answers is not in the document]
jordanrfrazier: You're correct - I can pass a reference to this in the materialize_impl method and it works.

crates/sparrow-main/src/serve/compute_service.rs
jordanrfrazier: Done.

The benefit of splitting it into this method gives me one place to handle the error, which I like.

crates/sparrow-materialize/src/materialize.rs
jordanrfrazier: Yeah, is there a better way to do this? This was the only way I could deference the `MutexGuard` to get the value out, but I couldn't do it without the reference because it was moving values out of `self`, which is a reference.

crates/sparrow-materialize/src/materialize.rs
jordanrfrazier: By that, do you mean: 
```
        let base_status = MaterializationStatus {
            id: self.id.clone(),
            progress: ProgressInformation::default(),
            state: State::Uninitialized,
            error: None,
        };

        match &*progress {
            MaterializationProgress::Uninitialized => MaterializationStatus {
                state: State::Uninitialized,
                ..base_status
            },
...
```

crates/sparrow-materialize/src/materialize/materialization_control.rs
```diff
@@ -0,0 +1,219 @@
...
+            },
+            MaterializationProgress::Failed(progress, e) => MaterializationStatus {
+                id: self.id.clone(),
+                progress: progress.clone(),
+                state: State::Failed,
+                error: Some(e.to_string()),
+            },
+        }
+    }
+}
```
bjchambers, line 180: Thought experiment -- may be easier / cleaner to clone the progress info and then divide it up to make the result.

```suggestion
    pub fn get_status(&self) -> MaterializationStatus {
        // Review note: While excess cloning can be a concern, it doesn't seem too bad here.
        // We need to make a copy of the progress to return anyway.
        let progress = self.progress_receiver.borrow().clone();
        let error = match progress {
          MaterializationProgress::Failed(_, e) => Some(e.to_string()),
          _ => None
        };
 
        let progress = match progress {
          MaterializationProgress::Uninitialized => ProgressInformation::default(),
          MaterializationProgress::Running(progress) => progress,
          MaterializationProgress::Stopped(proggres) => progress,
          MaterializationProgress::Failed(progress, _) => progress,
        };
        MaterializationStatus {
          id: self.id.clone(),
          progress,
          state: ???,
          error,
        }
    }
}
```

crates/sparrow-materialize/src/materialize/materialization_control.rs
```diff
@@ -0,0 +1,219 @@
...
+
+impl Default for State {
+    fn default() -> Self {
+        Self::Uninitialized
+    }
+}
+
+/// Internal progress information for a materialization.
+#[derive(Debug)]
+enum MaterializationProgress {
```
bjchambers, line 214: May be better to line this up with the status a bit further. This would let us eliminate some duplication of code from things like get status.

For instance:

```
struct MaterializationProgress {
  state: State,
  progress: ProgressInformation,
  error: Option<error_stack::Report<Error>>,
}
```

crates/sparrow-materialize/src/materialize/materialization_control.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// Unique identifier of the materialization
+    pub id: String,
+    /// Progress information
+    pub progress: ProgressInformation,
+    /// Current state of the materialization
+    pub state: State,
+    /// Failure reason
+    ///
+    /// Only populated if the materialization has failed
+    pub error: Option<String>,
```
bjchambers, line 194: Why `String` -- it seems like it would be better for the consumer of this to receive the actual error, and then it can choose to convert to a string if desired. For instance, this discards any information about error codes that may be attached to the error, etc.

crates/sparrow-materialize/src/materialize/materialization_control.rs
```diff
@@ -0,0 +1,219 @@
...
+            },
+        }
+    }
+}
+
+/// Publish status of a materialization.
+#[derive(Debug, Default)]
+pub struct MaterializationStatus {
+    /// Unique identifier of the materialization
+    pub id: String,
```
bjchambers, line 186: Why does this need the ID? If I recall, the only place this ID is used is as part of the service -- we could let that attach the ID. This, along with other suggestions then mean that `MaterializationStatus` is the same as `MaterializationProgress`, which lets you collapse those two structs and eliminate the boilerplate converting between the two.

jordanrfrazier: It doesn't - felt like more of an API convention to include the ID as part of the response, but happy to remove if it saves us that clone and we feel it's not needed.

crates/sparrow-main/src/serve/compute_service.rs
```diff
@@ -101,21 +108,79 @@ impl ComputeService for ComputeServiceImpl {
...
+            };
+
+            Ok(Response::new(GetMaterializationStatusResponse {
+                materialization_id: id,
+                state: status.state as i32,
+                progress: Some(status.progress),
+                error,
+            }))
+        } else {
+            Err(tonic::Status::internal(format!(
```
bjchambers, line 180: If we want this to be clean, should probbaly be `tonic::Status::not_found` or some such... but probably OK for now.

crates/sparrow-materialize/src/materialize/materialization_control.rs
```diff
@@ -0,0 +1,221 @@
...
+
+use crate::{Error, Materialization};
+
+/// The control for a materialization process.
+///
+/// This handles communication with the materialization process,
+/// allowing a user to manipulate the process via the service layer.
+pub struct MaterializationControl {
+    /// Unique identifier of the materialization
+    pub id: String,
```
bjchambers, line 17: Does this need the ID? I wonder if we could keep the ID a purely service concept:

1. Materialization: How we actually do materialization
2. Materialization Control: How we manage an asynchronous materialization
3. Service: How we implement the gRPC interface, including the map from IDs to materializations.

In that model, it seems like the ID doesn't matter here.

bjchambers (approved):

crates/sparrow-main/src/serve/compute_service.rs
jordanrfrazier: Good catch, updated

crates/sparrow-materialize/src/materialize/materialization_control.rs
jordanrfrazier: Hm, we could. But I currently have it for logging within the control API. I'll leave it for debugging purposes, but added a note to remove it if decide it's unnecessary

PR #389, opened 2023-05-26
example: Add SVG to PNG conversion script

bjchambers:


examples/timelines/convert.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
```
therapon, line 2: I would add in a comment a sample invocation as a simple way to document how to use the script.

therapon: Spoke too soon. There are not options for the script it's all in the same dir. Neverming

therapon (approved):

therapon:
@cla-bot check

Issue #257, opened 2023-04-20
feat: Numbers should cast to timestamps as Unix timestamps


kerinin:
**Summary**
Loading a CSV containing timestamps encoded as unix timestamps results in times that are all in 1970. We should treat integers as seconds since the unix epoch when converting to timestamps. We should also allow casting decimals to timestamps. 

Example CSV that should be parseable:

```
t,k,v
1682019069.0,foo,bar
```

bjchambers:
I suspect this is because of how prepare and casting use arrow's cast kernel. Specifically, if we go from `i64` to `timestamp_ns`, I believe it just does a cast interpreting the number as nanoseconds. We probably need to go from `i64 -> timestamp_ms -> timestamp_ns`, or otherwise do math on the `i64` before converting.

When we do this, we should make sure to do it everywhere:
1. In prepare, when using an `i64` column as the timestamp
2. In Fenl when writing `number_column as timestamp_ns`.
3. Etc.

Specifically, we should make sure that `shift_to(number_column as timestamp_ns)` is basically the same as using `number_column` as the time (assuming ordering is correct, etc.).

bjchambers:
Started looking at this. I'm not sure it's as straightforward as described, especially the case of casting.

The intention of "casting" is to convert between different types. `number as timestamp_ns` is indicating "interpret the number as a nanosecond timestamp". It doesn't seem like this should "treat the number as milliseconds and then convert that to milliseconds". And there is a way to write that -- `(number as timestamp_ms) as timestamp_ns`.

For the case of the time column, it may be better to allow specifying a `time_unit` that defaults to `ms` and informs how we convert the time in prepare. We could even have the `time_unit` option default to `ms` which would change the default behavior as desired.

bjchambers:
One issue with the above, is that it could potentially contradict the schema in the Parquet file, etc. It may be better to handle this as part of a "richer" schema for CSV files.

PR #394, opened 2023-05-30
proposed: time column format support

rltvty:
for #257 

proposed new common TimestampFormat message that we can use for various future configuration.  Use this new message to specify the `time_column_format` both for manager tables and for compute/prepare.

### DIFF
--- a/proto/kaskada/kaskada/v1alpha/common.proto
+++ b/proto/kaskada/kaskada/v1alpha/common.proto
@@ -52,6 +52,29 @@ message FileResults {
   repeated string paths = 2;
 }
 
+enum TimestampPrecision {
+  // defaults to nanoseconds if unspecified
+  // note: this behavior may change in the future
+  TIMESTAMP_UNSPECIFIED = 0;
+  TIMESTAMP_SECONDS = 1;
+  TIMESTAMP_MILLISECONDS = 2;
+  TIMESTAMP_MICROSECONDS = 3;
+  TIMESTAMP_NANOSECONDS = 4;
+}
+
+message TimestampFormat {
+  oneof format {
+
+    // This format indicates the amount of time since the unix epoch.
+    // For example "TIMESTAMP_MILLISECONDS" indicates that values describe the number of milliseconds since the unix epoch.
+    TimestampPrecision unix_timestamp = 1;
+
+    // This format describes a string-formatted timestamp.
+    // For example, "%Y-%m-%d %H:%M:%S" would correspond to the timestamp string "2015-09-05 23:56:04"
+    string format_string = 2;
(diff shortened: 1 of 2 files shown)


proto/kaskada/kaskada/v1alpha/common.proto
```diff
@@ -52,6 +52,29 @@ message FileResults {
...
+    TimestampPrecision unix_timestamp = 1;
+
+    // This format describes a string-formatted timestamp.
+    // For example, "%Y-%m-%d %H:%M:%S" would correspond to the timestamp string "2015-09-05 23:56:04"
+    string format_string = 2;
+  }
+}
+
 // General configuration for a table.
 message TableConfig {
```
rltvty, line 79: this message type is used for Prepare, Compute, Materialize.  To support this, there are probably numerous changes required in sparrow.

bjchambers:
> > I think it may be better to focus this on CSV support. Arrow allows us to read CSV in two ways:
> > 
> > 1. Infer the schema, and then read the CSV file using the inferred schema.
> > 2. Read the CSV file using a specified schema.
> > 
> > A cleaner solution here might be to allow the CSV-backed tables to specify a schema. This would allow them to say "this column (`timestamp`) is `timestamp_ms`" which would then cause it to parse as expected.
> 
> If this is focused on only CSV support, how do we handle different timestamp formats from parquet and pulsar?

Parquet schemas already have an internal precision (the [unit](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#timestamp)). If this includes Parquet, how do we handle cases where the Parquet Schema specifies a precision which differs from this format?

In Pulsar, we currently use Avro schemas, which also already have a way of encoding the [unit](https://avro.apache.org/docs/1.8.0/spec.html#Timestamp+%28millisecond+precision%29).

I think in both of those cases we should focus on supporting the established standard for specifying the precision, since this avoids cases where there is disagreement, and also means that the user doesn't need to specify anything to use properly formatted files.

This *only* affects CSV because that is the only format that doesn't have a schema (and/or the CSV schema doesn't include precision information).

bjchambers:
> that makes sense for supporting unix timestamp fields. what do we do for string-based timestamps?

That still seems to be a mostly CSV issue -- if the user writes data containing timestamps to Parquet or Avro, it would use the timestamp data type. So, this means that either (a) the format is CSV which didn't have a schema, so timestamps were written as strings for clarity or (b) the user stored the data as a string intentionally.

That latter case seems unlikely in a scenario where a user is investing in managing the data -- data lake, etc. -- and seems to be outside the focus of getting this working for demos / notebooks / etc. with CSV.

PR #399, opened 2023-05-30
build: Don't ignore `main`

bjchambers:
This is a dangerous name to ignore -- basically any Java/Scala code will be in a directory like `src/main` which will be entirely ignored by this pattern.

We should either ignore `/main` (if we really want to ignore it only at the root level) or better yet change the configuration of `wren` to not compile to `main`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -286,3 +286,4 @@ debug/
 .ionide
 
 # End of https://www.toptal.com/developers/gitignore/api/go,git,osx,rust,python,visualstudiocode
+wren/main
--- a/.gitignore
+++ b/.gitignore
@@ -286,6 +286,3 @@ debug/
 .ionide
 
 # End of https://www.toptal.com/developers/gitignore/api/go,git,osx,rust,python,visualstudiocode
-
-# Wren compiles to `main`?
-main
--- a/Makefile
+++ b/Makefile
@@ -95,26 +95,26 @@ test/int/run-api-s3:
 ## CI related targets
 ####
 ci/integration/tests/docker-compose-up:
-	export DOCKER_BUILDKIT=1 
+	export DOCKER_BUILDKIT=1
 	docker compose -f ./tests/integration/docker-compose-ci-integration.yml up --build --detach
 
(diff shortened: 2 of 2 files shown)


therapon (approved):

kevinjnguyen (approved):

kevinjnguyen:
> @kevinjnguyen @epinzur Do you know where we produce this name? I think it's because `wren` is called `main.go`. We could maybe change that to `wren.go` (not sure about Go patterns), and/or compile with `-o wren` to change the output name. I'm not sure of the full history, so want to make sure there aren't other places we should change to make this less problematic.

You are correct. The default output name of a go binary is the name of the main file. This is most likely the only reason for adding the `main` to the git ignore.

therapon:
> @kevinjnguyen @epinzur Do you know where we produce this name? I think it's because `wren` is called `main.go`. We could maybe change that to `wren.go` (not sure about Go patterns), and/or compile with `-o wren` to change the output name. I'm not sure of the full history, so want to make sure there aren't other places we should change to make this less problematic.

We do use the `-o` option with the `go` compiler 
https://github.com/kaskada-ai/kaskada/blob/a76693cad49a27b5cf5affaad050fbc40f2ce63e/.github/workflows/reusable_ci_wren.yml#L81

I am guessing the entry in the `.gitignore` file was leftover from a previous setup?

.gitignore
```diff
@@ -285,7 +285,4 @@ debug/
 .history
 .ionide
 
-# End of https://www.toptal.com/developers/gitignore/api/go,git,osx,rust,python,visualstudiocode
-
-# Wren compiles to `main`?
-main
\ No newline at end of file
+# End of https://www.toptal.com/developers/gitignore/api/go,git,osx,rust,python,visualstudiocode
```
therapon, line 288:
```suggestion
# End of https://www.toptal.com/developers/gitignore/api/go,git,osx,rust,python,visualstudiocode
wren/main
```

bjchambers:
I did find one place we didn't use `-o wren` (in the Makefile) and fixed it.

Issue #409, opened 2023-06-01
feat: Partitioned execution


bjchambers:
**Summary**

The current "execution plan" is close, but not quite correctly adapted for use when describing the steps necessary to perform a computation.
Ideally, we would introduce a physical plan that is more similar to relational query engines, allowing us to leverage existing techniques and creating the options to run on existing systems.

For now, the plan is to introduce these and move execution towards running them directly, and then (separately) work towards compiling queries directly to physical plans.

- [x] Introduce serde compatible physical plan structs.
- [x] Introduce a "pipeline scheduler" putting schedule information in the physical plans
- [x] Introduce new expression executors
- [ ] Introduce a partitioned, pipeline executor
- [ ] Add additional transforms (select, lookup, etc.)
- [ ] Add merge based pipelines (repartition, merge)
- [ ] Figure out how to incorporate aggregations
- [ ] Connect to "real" sources
- [ ] Connect to real sinks
- [ ] Convert the existing execution plan to a physical plan.
- [ ] Update tests to run (with a flag) using the converted plans
- [ ] Introduce a protobuffer encoding for the physical plan. It may be possible to share this with a serialization of logical plans since they will have similar structure. Ideally, this would allow for source/sink extensibility.
- [ ] "partitioned" parquet reader + "partitioned" prepare

bjchambers:
#407 was a refactoring to move the ScalarValue into a more accessible location for the physical plans. For now, the intention is to use the ScalarValue within the physical plan to represent literal values. In the long term we may want to revisit that and use a better encoding that would be more aligned with logical plans, but can revisit once the basic plumbing is laid out a bit better.

PR #410, opened 2023-06-01
feat: introducing physical plans

bjchambers:
This is part of #409.


crates/sparrow-physical/src/expr.rs
```diff
@@ -0,0 +1,79 @@
...
+    enum_as_inner::EnumAsInner,
+)]
+#[serde(rename_all = "snake_case")]
+pub enum ExprKind {
+    /// Apply the named instruction to the given children.
+    Call(Cow<'static, str>),
+    /// Reference an input column by name.
+    Column(String),
+    /// A boolean literal.
+    BooleanLiteral(bool),
```
bjchambers, line 41: I was originally going to have a `Literal(ScalarValue)` but that is a very implementation specific option. My current thinking is that it may be better to use a combination of these 3 simple literals (booleans, strings, numerics) and maybe `null` literals, in combination with the `DataType` of the expression to create the scalar value if we need it.

I'd propose we see how it plays out when we go to generate and/or execute these and see what we think. Thoughts?

crates/sparrow-physical/src/plan.rs
```diff
@@ -0,0 +1,12 @@
+use crate::Step;
+
+/// A plan is a directed, acyclic graph of steps.
+///
+/// The plan is represented as an array of steps, with each step referencing
+/// it's children (inputs) by index. The array is topologically sorted so that
+/// all children have indices less than the step that references them.
+#[derive(Debug)]
```
bjchambers, line 8: My thinking for now that (a) supporting `serde` is good for plans/steps/expressions since that lets us use JSON or YAML as a way to read/write plans locally and (b) we can create a protobuf that captures the structure of the plan and write a conversion.

I suspect the physical and logical plans will have basically the same structure, with some differences in terms of what kinds of expressions and steps they support. We could probably use the same protobuf with either a "superset enum" or some extension points. Will wait to see when we have the logical and are implementing the protobuf representation.

crates/sparrow-physical/src/step.rs
```diff
@@ -0,0 +1,59 @@
+use arrow_schema::SchemaRef;
+
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+#[repr(transparent)]
+#[serde(transparent)]
+pub struct StepId(usize);
+
+/// A single step in the physical plan.
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub struct Step {
```
bjchambers, line 10: From a naming perspective:

"steps" describe a "table" or "batch" (set of columns) and "expressions" describe a column. Not sure if anyone has a better / clearer suggestion here.

crates/sparrow-physical/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+#![warn(
```
bjchambers, line 1: Please especially look at naming / comments to see if the general structure here is clear. I'll flesh these out further as I work on executing them and converting existing plans to physical plans for execution.

crates/sparrow-physical/src/step.rs
therapon: Personally, slight preference towards batch than table

crates/sparrow-physical/src/step.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub id: StepId,
+    /// The kind of step being performed.
+    pub kind: StepKind,
+    /// Steps which act as input (children) of this step.
+    pub children: Vec<StepId>,
```
therapon, line 15: I am not 100% sure why we prefer `children` rather than `input`, `upstream` or some other word that does not denote parent-child relationship.

crates/sparrow-physical/src/step.rs
```diff
@@ -0,0 +1,59 @@
...
+        /// The last expression should be the boolean predicate.
+        exprs: Vec<crate::Expr>,
+    },
+    /// A step that repartitions the output.
+    Repartition {
+        num_partitions: usize,
+        /// Expressions to apply to compute columns which may be referenced by `keys`.
+        exprs: Vec<crate::Expr>,
+        /// Indices of expression columns representing the keys.
+        keys: Vec<usize>,
```
therapon, line 56: I am not really sure how these two fields work? For if the keys are only used by the exprs would it be better to place them in one field that is a tuple holding the exprs and the keys, or even a Map?

crates/sparrow-physical/src/expr.rs
bjchambers: Most types I believe would use some of these -- for instance, `type=TimestampMillis` with a `NumberLiteral` would interpret the number as the number of millis.

crates/sparrow-physical/src/step.rs
bjchambers: Agreed on that -- but the naming question here is `step` and `expression` (vs some alternative).

crates/sparrow-physical/src/expr.rs
```diff
@@ -0,0 +1,79 @@
...
+/// input columns. Expressions appear in a variety of places within the steps
+/// that make up a physical plan.
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+
+pub struct Expr {
+    pub kind: ExprKind,
+    /// Arguments to the expression.
+    ///
+    /// These are indices referencing earlier expressions.
+    pub children: Vec<usize>,
```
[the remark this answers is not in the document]
therapon: Thanks. My vote is wrapping `usize` and have `StepId` and `ExpId`

crates/sparrow-physical/src/step.rs
bjchambers: I think we adopted it from `egg`. I'm OK calling this `inputs` (in `Step`) and maybe `args` (in `Expression`).

crates/sparrow-physical/src/step.rs
bjchambers: The difficulty is if we want to key by something like `(a + b, a + c)` as the columns. The way it worked in the prototype was that the expressions create columns and then keys reference some of those columns. This lets us do:

```
exprs: 
  - a
  - b
  - c
  - add [0, 1]
  - add [0, 2]
keys: [3, 4]
```

So we don't need to repeat the common sub-expressions (can reference them by index in later expressions), etc.

We *could* (although I'm not sure if it would be clearer or not) create a struct which was basically a vec of expressions and a vec of outputs, then use that in each stepkind which computes things using expressions.

crates/sparrow-physical/src/step.rs
therapon: `step` / `expression` looks good to me

crates/sparrow-physical/src/expr.rs
```diff
@@ -0,0 +1,91 @@
...
+
+/// A physical expression which describes how a value should be computed.
+///
+/// Generally, each expression computes a column of values from zero or more
+/// input columns. Expressions appear in a variety of places within the steps
+/// that make up a physical plan.
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+
+pub struct Expr {
+    pub kind: ExprKind,
```
therapon, line 25: Maybe add a one line comment here to explain what we mean with `kind`. I would throw the other names like `opcode` or `function`

crates/sparrow-physical/src/plan.rs
```diff
@@ -0,0 +1,12 @@
+use crate::Step;
+
+/// A plan is a directed, acyclic graph of steps.
+///
+/// The plan is represented as an array of steps, with each step referencing
+/// it's children (inputs) by index. The array is topologically sorted so that
```
therapon, line 6: topology here being the order of visitation (or evaluation?).

bjchambers: Rather than trying to explain "order of visitation" just stated: "every step appears after the inputs to that step". This seems to explain what we mean by "topologically sorted" by stating the exact property, which seems preferable. Otherwise, we'd need to explain what we mean by "order of visitation" (DFS or BFS? bottom up or top down? etc.) Thoughts?

crates/sparrow-physical/src/step.rs
bjchambers: Thoughts on whether it would be clearer to have something like:

```
struct Expressions {
  columns: Vec<crate::Expr>,
  outputs: Vec<usize>,
}
```

Then, we could have things like Project and Repartition use that struct. It would basically say "the pattern for computing values is to use this struct.

Does that add to much indirection, or does it seem to give us a better way to systematize this pattern?

therapon: This looks like a good idea to me but I am not 100% sure about pros and cons--extra indirection etc. 

Let's try it. The pattern should show its benefits/issues early on and then we can decide on keeping it or removing it.

therapon (approved):

PR #412, opened 2023-06-02
ref: create an Exprs

bjchambers:


crates/sparrow-physical/src/expr.rs
```diff
@@ -2,6 +2,35 @@ use std::borrow::Cow;
...
+/// Represents 1 or more values computed by expressions.
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub struct Exprs {
+    /// The expressions computing the intermediate values.
+    pub exprs: Vec<Expr>,
+    /// The indices of columns to output.
+    pub outputs: Vec<ExprId>,
+}
+
+impl Exprs {
```
bjchambers, line 14: If we add a lot, I may move this to a separate exprs.rs file... but that can wait.

therapon (approved):

PR #413, opened 2023-06-02
feat: introduce pipeline scheduler

bjchambers:
This is part of #409.

Introduces `Pipeline` information to the physical plan. This indicates which steps are part of a linear sequence, and should (ideally) be executed together.

Also implements a pipeline "scheduler" to determine the pipeline for each step, in a new `sparrow-backend` crate. As the physical plan is built-up, the code should go in this "compiler backend" package, which can own optimization and conversion of logical plans to physical plans.

### DIFF
--- a/crates/sparrow-backend/src/pipeline_schedule.rs
+++ b/crates/sparrow-backend/src/pipeline_schedule.rs
@@ -29,8 +29,7 @@ pub fn pipeline_schedule(steps: &IndexSlice<StepId, [Step]>) -> Vec<Pipeline> {
 
     let mut pipelines = Vec::new();
     for (index, step) in steps.iter_enumerated() {
-        let index: StepId = index.into();
-        let assignment = if is_pipeline_breaker(index, &step, &references) {
+        let assignment = if is_pipeline_breaker(index, step, &references) {
             // A step with no input (such as a scan) starts a new pipeline.
             // A step with multiple inputs (such as a merge) is a separate pipeline.
             let index = pipelines.len();
--- a/crates/sparrow-backend/src/pipeline_schedule.rs
+++ b/crates/sparrow-backend/src/pipeline_schedule.rs
@@ -1,5 +1,6 @@
+use index_vec::{IndexSlice, IndexVec};
 use sparrow_core::debug_println;
-use sparrow_physical::{Pipeline, Step, StepKind};
+use sparrow_physical::{Pipeline, Step, StepId, StepKind};
 
 const DEBUG_SCHEDULING: bool = false;
 
@@ -13,47 +14,23 @@ const DEBUG_SCHEDULING: bool = false;
 /// 3. Any other operation with a single input is part of the
 ///    the same pipeline as the input.
 /// 4. Any other operation is a separate pipeline.
-pub fn pipeline_schedule(steps: &[Step]) -> Vec<Pipeline> {
+pub fn pipeline_schedule(steps: &IndexSlice<StepId, [Step]>) -> Vec<Pipeline> {
     // Compute how many times each step is referenced.
     // Any step referenced multiple times must end a pipeline.
-    let mut references = vec![0; steps.len()];
+    let mut references = index_vec::index_vec![0; steps.len()];
     for step in steps {
         for input in &step.inputs {
-            references[input.0] += 1;
+            references[*input] += 1;
         }
     }
 
     // Then compute the assignments for each step.
-    let mut assignments = Vec::with_capacity(steps.len());
+    let mut assignments = IndexVec::with_capacity(steps.len());
 
     let mut pipelines = Vec::new();
-    for (index, step) in steps.iter().enumerate() {
-        let break_pipeline = if step.inputs.len() != 1 {
-            debug_println!(
-                DEBUG_SCHEDULING,
-                "Step {index} is new pipeline since it has multiple inputs"
-            );
-            true
-        } else if references[step.inputs[0].0] > 1 {
-            debug_println!(
-                DEBUG_SCHEDULING,
-                "Step {index} is new pipeline since it's only input ({}) is referenced {} times",
-                step.inputs[0].0,
-                references[step.inputs[0].0]
-            );
-            true
-        } else if is_pipeline_breaker(&step.kind) {
-            debug_println!(
-                DEBUG_SCHEDULING,
-                "Step {index} is new pipeline based on kind {:?}",
-                step.kind
-            );
-            true
-        } else {
-            false
-        };
-
-        let assignment = if break_pipeline {
+    for (index, step) in steps.iter_enumerated() {
+        let index: StepId = index.into();
+        let assignment = if is_pipeline_breaker(index, &step, &references) {
             // A step with no input (such as a scan) starts a new pipeline.
             // A step with multiple inputs (such as a merge) is a separate pipeline.
             let index = pipelines.len();
@@ -61,11 +38,11 @@ pub fn pipeline_schedule(steps: &[Step]) -> Vec<Pipeline> {
             index
         } else {
             // The step has exactly 1 input and is not a pipeline breaker.
-            assignments[step.inputs[0].0]
+            assignments[step.inputs[0]]
         };
 
         // Add this step to the assigned pipeline and record the assignment.
-        pipelines[assignment].steps.push(index.into());
+        pipelines[assignment].steps.push(index);
         assignments.push(assignment);
     }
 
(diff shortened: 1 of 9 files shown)


crates/sparrow-backend/src/pipeline_schedule.rs
```diff
@@ -0,0 +1,163 @@
...
+///
+/// This is computed in a bottom-up manner:
+///
+/// 1. Each leaf (input) is the start of a separate pipeline.
+/// 2. Certain step kinds are "pipeline breaking", which means
+///    they start a new pipeline. For instance, `with_key`.
+/// 3. Any other operation with a single input is part of the
+///    the same pipeline as the input.
+/// 4. Any other operation is a separate pipeline.
+pub fn pipeline_schedule(steps: &[Step]) -> Vec<Pipeline> {
```
bjchambers, line 16: This is a simple analysis, showing the benefits (simplicity) of using the topologic order.

crates/sparrow-backend/src/pipeline_schedule.rs
```diff
@@ -0,0 +1,163 @@
...
+            references[input.0] += 1;
+        }
+    }
+
+    // Then compute the assignments for each step.
+    let mut assignments = Vec::with_capacity(steps.len());
+
+    let mut pipelines = Vec::new();
+    for (index, step) in steps.iter().enumerate() {
+        let break_pipeline = if step.inputs.len() != 1 {
```
bjchambers, line 31: This was helpful for debugging, but makes the code a lot longer. Really, the criteria is:

```
step.inputs.len() != 1 || references[step.inputs[0.0] > 1 || is_pipeline_breaker(&step.kind)
```

My thought was to keep the longer form around, since it *is* useful to explain why a pipeline was broken... but I could be convinced to drop this for simplicity.

crates/sparrow-backend/src/pipeline_schedule.rs
```diff
@@ -0,0 +1,163 @@
...
+            debug_println!(
+                DEBUG_SCHEDULING,
+                "Step {index} is new pipeline since it has multiple inputs"
+            );
+            true
+        } else if references[step.inputs[0].0] > 1 {
+            debug_println!(
+                DEBUG_SCHEDULING,
+                "Step {index} is new pipeline since it's only input ({}) is referenced {} times",
+                step.inputs[0].0,
```
bjchambers, line 41: The use of `.0` (to get the `usize` out of the newtype wrappers) and `.into()` to convert `usize` to the newtype wrappers) is kind of a pain. I'm partway inclined to say we should just use `usize` in the physical plan for simplicity. Thoughts?

crates/sparrow-physical/src/plan.rs
```diff
@@ -9,4 +9,19 @@ use crate::Step;
...
+
+/// Information about a specific "pipeline" within the plan.
+///
+/// Pipelines take a single input through a linear sequence of
+/// steps. Identifying pipelines within the plan provides units
+/// of work that ideally run on the same core.
+///
+/// See [Morsel-driven Parallelism](https://dl.acm.org/doi/10.1145/2588555.2610507).
+#[derive(Debug, serde::Serialize, serde::Deserialize, Default, PartialEq)]
+pub struct Pipeline {
```
bjchambers, line 24: Adds "pipeline information" to the plan.

crates/sparrow-backend/src/pipeline_schedule.rs
jordanrfrazier: I imagine we'll want the debugging. Could refactor into a `breaks_pipeline()` method to shorten this method.

therapon (approved):

crates/sparrow-backend/src/pipeline_schedule.rs
therapon: I thought there was a way to get the wrapper to be "transparent" with some `repr` trick. Does that not work in this case?

jordanrfrazier (approved):

crates/sparrow-backend/src/pipeline_schedule.rs
```diff
@@ -0,0 +1,163 @@
...
+///
+/// 1. Each leaf (input) is the start of a separate pipeline.
+/// 2. Certain step kinds are "pipeline breaking", which means
+///    they start a new pipeline. For instance, `with_key`.
+/// 3. Any other operation with a single input is part of the
+///    the same pipeline as the input.
+/// 4. Any other operation is a separate pipeline.
+pub fn pipeline_schedule(steps: &[Step]) -> Vec<Pipeline> {
+    // Compute how many times each step is referenced.
+    // Any step referenced multiple times must end a pipeline.
```
jordanrfrazier, line 18: I don't quite understand this rule. Wouldn't the `sum` step be referenced multiple times, but still possible to be part of the same pipeline? 
```
let sum = sum(Foo)
in { f1: sum, f2: sum }
```

crates/sparrow-backend/src/pipeline_schedule.rs
bjchambers: So, `repr(transparent)` just means that in assembly, there is no difference between the struct and the field -- so it's a "zero cost" wrapper, at runtime.

But, the whole value of the type wrapper is that it requires you to explicitly convert in/out. So we *could* add a `From<ExprId> for usize`, which would let both conversions be `x.into()`, but there will always need to be an explicit conversion.

The alternative would be a type alias `type ExprId = usize`, but that is basically "erased" at compile time, I believe, so it provides no safety benefits.

crates/sparrow-backend/src/pipeline_schedule.rs
bjchambers: I don't think `pipeline::schedule` makes sense, since I don't think we'd have anything else in that module.

I can see either:

1. `schedule_pipeline` name it a "verb" -- I tend to prefer this for "imperative" things, where it makes sense to describe what the function *does*
2. `pipeline_schedule` name it a "noun phrase" -- something that *describes* what it returns. I tend to prefer this for things that are more "functional" (executed for the result rather than the side-effects).

That said, I'm fine with either `schedule_pipeline` or `pipeline_schedule`.

crates/sparrow-backend/src/pipeline_schedule.rs
therapon: Ah got it, thanks

bjchambers: So I guess, we have the following options:

1. Keep the wrappers. We'll always need to do `2.into()` and `some_array[2.into()]`, unless we create our own "wrapper aware vector" so we can override the indexing to do the conversion. Seems to be heavyweight, but doable. For instance `ExprId` and `ExprIdVec`, etc.
2. Drop the wrappers. Use `usize`. Use good naming and trust ourselves.
3. Use type aliases. Keeps a unique name, but doesn't provide any protections.

I'm generally torn between 1 and 2 -- I don't think 3 buys us much. Thoughts?

crates/sparrow-backend/src/pipeline_schedule.rs
bjchambers: The `sum` is an aggregation -- it's an expression within one step. In the above, we'd actually only have one step performing a aggregations.

```
Step {
  kind: StepKind::Aggregation {
    aggregates: vec![
      0="inputs[0]",
      1="sum(0)",
      2="last(1)",
      3="first(1)",
  ],
  inputs: vec![scan],
  schema: ...
}
```

crates/sparrow-backend/src/pipeline_schedule.rs
jordanrfrazier: Cool, no makes sense to leave as is then

crates/sparrow-backend/src/pipeline_schedule.rs
```diff
@@ -0,0 +1,163 @@
...
+        // Add this step to the assigned pipeline and record the assignment.
+        pipelines[assignment].steps.push(index);
+        assignments.push(assignment);
+    }
+
+    pipelines
+}
+
+/// Return true if the step is "pipeline breaking".
+fn is_pipeline_breaker(index: StepId, step: &Step, references: &IndexVec<StepId, usize>) -> bool {
```
bjchambers, line 53: Split out and used the `IndexVec`.

crates/sparrow-physical/src/step.rs
```diff
@@ -2,23 +2,30 @@ use arrow_schema::SchemaRef;
...
+/// to specify columns -- for instance the columns to compute in
+/// a projection or the condition to use for a selection.
+/// Conceptually, steps describe how batches are produced while
+/// expressions describe columns inside a batch.
+///
+/// During execution, each step receives an partitioned stream
+/// of ordered batches and produces a partitioned stream of
+/// ordered batches. Steps never operate between partitions --
+/// instead, operations like `with_key` for a given partition
+/// produce output to destined for multiple partitions based
```
therapon, line 25:
```suggestion
/// produce output destined for multiple partitions based
```

crates/sparrow-physical/src/step.rs
```diff
@@ -2,23 +2,30 @@ use arrow_schema::SchemaRef;
...
 /// A single step in the physical plan.
+///
+/// Each step corresponds to a specific relational operator.
+/// Many [kinds of steps](StepKind) include [expressions](Exprs)
+/// to specify columns -- for instance the columns to compute in
+/// a projection or the condition to use for a selection.
+/// Conceptually, steps describe how batches are produced while
+/// expressions describe columns inside a batch.
+///
+/// During execution, each step receives an partitioned stream
```
therapon, line 21:
```suggestion
/// During execution, each step receives a partitioned stream
```

therapon (approved):

PR #416, opened 2023-06-05
docs(diagnostic): moves diagnostic message to individual html page

therapon:
Diagnostic Pages changes 

* moves under Troubleshooting nav item 
* diagnostic page from nav is a table of error code and message that links to each diagnostic page 
* diagnostic pages is a single HTML page with the error message and text on the diagnosis and possible fixes.


crates/sparrow-compiler/src/diagnostics/code.rs
```diff
@@ -58,7 +58,7 @@ DuplicateFieldNames(E0009, Error, "Duplicate field names in record expression",
 InvalidArgumentType(E0010, Error, "Invalid argument type(s)", ""),
 SyntaxError(E0011, Error, "Invalid syntax", ""),
 CyclicReference(E0012, Error, "Circular dependency", ""),
-InvalidOutputType(E0013, Error, "Invalid output type", "https://kaskada.io/docs-site/kaskada/main/fenl/fenl-diagnostic-codes.html#e0013"),
+InvalidOutputType(E0013, Error, "Invalid output type", "https://kaskada.io/docs-site/kaskada/main/troubleshooting/diagnostic-pages/e0013.html"),
```
bjchambers, line 61: Ideally, this would be something like a template here, and we'd just have a boolean on the cases that are error codes.

Also note I'd like to rework a lot of these error codes... so they're likely to change. Not sure how we'll revision things... maybe that will just need to be a one time thing... or I could probably find a way to generate a slightly different parallel URL.

therapon: Agreed about templating (or macro?). 

This PR is mainly to showcase a different setup for the html pages. We had each error be an anchor on 1 page and this PR makes each error an individual page. We link all of them on a "landing page" inside a table. 

The reason I would like to have them as individual pages 

* easier pattern, root URL and append the page name (e.g. `e0013.html`) 
* easier to redirect/alter individual pages without having one page that keeps updating 
* the 1 with anchors will eventually get too large

therapon: > Not sure how we'll revision things... maybe that will just need to be a one time thing... or I could probably find a way to generate a slightly different parallel URL.

I have a few ideas. 

1. add the version to the URL path. We can pick the version number from the toml file and generate redirects for the landing page 
1. we can create sub-folders for each version and softlink unchanged pages and add new pages accordingly 
1. we can make each page "clever" and pass in the version number that will then render the appropriate text 

I like something close to 1 above personally--we can hash out the implementation

bjchambers: Thoughts then:

1. Have a version in the directory path (`v0/e0013.html`) so that we can easily create a new set of error numbers.
2. Suggest having a page for *every* error code, even if it is just "nothing is documented for this error code; file a bug <here> or ask in <discussion> or <slack">"

bjchambers:
what's the status of this?

therapon:
> what's the status of this?

I was looking for a way to implement the idea in the comments (if a url to an error code does not exist show a generic page and a link to create the page/issue) but it looks like antora does not seem to do this. There is one way to configure error pages for the whole antora site and I have not found any antora plugin to help. Some solutions rely on an nginx running along with antora and dealing with redirects from nginx. 

At this point I am thinking of shelving this PR, create an issue and see if I can come up with another way to implement the idea. If there are any thoughts or known solutions to this please share.

bjchambers:
> > what's the status of this?
> 
> I was looking for a way to implement the idea in the comments (if a url to an error code does not exist show a generic page and a link to create the page/issue) but it looks like antora does not seem to do this. There is one way to configure error pages for the whole antora site and I have not found any antora plugin to help. Some solutions rely on an nginx running along with antora and dealing with redirects from nginx.
> 
> At this point I am thinking of shelving this PR, create an issue and see if I can come up with another way to implement the idea. If there are any thoughts or known solutions to this please share.

Hmm. Assuming that the error codes reference things in the documentation, and there is a degree of versioning, I wonder if we could just have a unit test that iterates over the cases of the error code (this is possible with `strum`) and verify that we have an error page in the Antora doc-src.

In the future (if/when we want to version things), we can actually do something like link to `<doc_version>/errors/<error_code>`, and use Antora to manage versioned docs and embed a `doc_version` in the binary so it knows which version to link against.

That seems like a minimal fix for now -- just make sure we created a placeholder for each error code. Thoughts?

therapon:
> > > what's the status of this?
> > 
> > 
> > I was looking for a way to implement the idea in the comments (if a url to an error code does not exist show a generic page and a link to create the page/issue) but it looks like antora does not seem to do this. There is one way to configure error pages for the whole antora site and I have not found any antora plugin to help. Some solutions rely on an nginx running along with antora and dealing with redirects from nginx.
> > At this point I am thinking of shelving this PR, create an issue and see if I can come up with another way to implement the idea. If there are any thoughts or known solutions to this please share.
> 
> Hmm. Assuming that the error codes reference things in the documentation, and there is a degree of versioning, I wonder if we could just have a unit test that iterates over the cases of the error code (this is possible with `strum`) and verify that we have an error page in the Antora doc-src.
> 
> In the future (if/when we want to version things), we can actually do something like link to `<doc_version>/errors/<error_code>`, and use Antora to manage versioned docs and embed a `doc_version` in the binary so it knows which version to link against.
> 
> That seems like a minimal fix for now -- just make sure we created a placeholder for each error code. Thoughts?

Seems reasonable. I'll give it a shot.

Issue #418, opened 2023-06-05
cleanup: Replace custom time logic with Arrow built-ins


bjchambers:
When we first supported timestamps and durations, arrow-rs didn't have very much support for operations (time between two dates, adding intervals to dates, etc.). This has come a long way. We should look to eliminate our custom logic in favor of standard implementations. As part of this we should see how our functionality maps to Arrow -- it may be OK to change some of the time functionality to better match the available functions.

Issue #417, opened 2023-06-05
bug: Take advantage of FieldRef to avoid copying


bjchambers:
**Description**
Recent versions of Arrow have introduced a `FieldRef = Arc<Field>` type. When upgrading, we continued cloning the fields and creating new `Arc`s around them for simplicity. However, in many places we should be able to hold a single `Arc<[FieldRef]>` like they do in `StructArray` and clone the `Arc`, allowing any operation on struct arrays to avoid allocating new fields for each field and the array of fields.

Filing this bug as part of the upgrade so that I can link to it in the relevant TODOs.

PR #419, opened 2023-06-05
build: Update to arrow 40

bjchambers:
Recent versions of Arrow make it much easier to create custom
operations -- things like `array.unary(...)` and working with
buffers more easily.

This upgraed was a bit painful since we were going through so
many versions. The biggest changes were the low level
operations to work with bits -- we can just iterate directly
now and no longer need our helper.

Fields also got wrapped in an `Arc` to avoid inefficient deep
copies. Filed #417 and #418 to take advantage of some of these
new capabilities.


crates/sparrow-main/tests/e2e/cast_tests.rs
```diff
@@ -37,7 +37,7 @@ async fn cast_data_fixture() -> DataFixture {
 
 #[tokio::test]
 async fn test_implicit_cast_i64_to_f64_add() {
-    insta::assert_snapshot!(QueryFixture::new("{ i64_field: Input.i64, f64_field: Input.f64, add: Input.i64 + Input.f64 }").run_to_csv(&cast_data_fixture().await).await.unwrap(), @r###"
+    insta::assert_snapshot!(QueryFixture::new("{ add: Input.i64 + Input.f64 }").run_to_csv(&cast_data_fixture().await).await.unwrap(), @r###"
```
therapon, line 40: how was this test passing? We seem to have removed fields from the output and yet we did not alter the expected result. Am I reading this wrong?

crates/sparrow-arrow/src/scalar_value.rs
```diff
@@ -64,13 +64,13 @@ impl From<bool> for ScalarValue {
 pub struct ScalarTimestamp {
     value: Option<i64>,
     unit: TimeUnit,
-    tz: Option<String>,
+    tz: Option<Arc<str>>,
```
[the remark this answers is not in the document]
bjchambers: I don't think so: `sparrow-arrow` is part of the implementation, and I don't think it is exposed. So it is public within the crate, but not part of the library surface (if we were to define one).

crates/sparrow-main/tests/e2e/cast_tests.rs
bjchambers: This may behave been a "accidentally changed the wrong thing and saved". I don't believe this test is passing. Fixing.

crates/sparrow-runtime/src/prepare.rs
```diff
@@ -463,23 +464,39 @@ const BATCH_SIZE: usize = 1_000_000;
...
+    let csv_reader = ReaderBuilder::new(raw_schema.clone())
         .has_header(true)
         .with_batch_size(BATCH_SIZE);
     let reader = csv_reader
         .build(reader)
         .into_report()
         .change_context(Error::CreateCsvReader)?;
     let raw_metadata =
-        RawMetadata::from_raw_schema(reader.schema()).change_context_lazy(|| Error::ReadSchema)?;
+        RawMetadata::from_raw_schema(raw_schema.clone()).change_context(Error::ReadSchema)?;
```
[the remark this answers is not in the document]
bjchambers: Lazy is preferred if (and only if) computing the new context is "expensive". For instance:

* `e.change_context(Error::ReadSchema)` is basically free -- it is creating a literal value of an enum.
* `e.change_context_lazy(|| Error::InternalString(format!("some thing happened {...} {}", ...)))` is preferred because it actually needs to assemble the string.

bjchambers:
There is a failing compiler test related to record simplification. I need to figure out the root cause before merging this.

crates/sparrow-instructions/src/evaluators/aggregation/boolean/last_boolean_evaluator.rs
```diff
@@ -161,25 +160,24 @@ impl LastBooleanEvaluator {
...
-                    })
-                    .collect::<anyhow::Result<BooleanArray>>()?
-            } else {
-                izip!(key_indices.values(), 0..)
-                    .map(|(entity_index, input_index)| {
-                        Self::update_accum(token, *entity_index, true, input.value(input_index))
-                    })
-                    .collect::<anyhow::Result<BooleanArray>>()?
-            };
+        let result: BooleanArray = if let Some(input_valid_bits) = input.nulls() {
```
jordanrfrazier, line 163: https://docs.rs/arrow/latest/arrow/array/trait.Array.html#tymethod.nulls

Wouldn't this be the...non-valid bits then? I still don't know.

crates/sparrow-runtime/src/execute/operation/shift_to.rs
```diff
@@ -333,9 +333,7 @@ impl ShiftToColumnOperation {
             shifted_times_primitive,
         )?;
 
-        let num_filtered = is_forward_shift.len()
-            - is_forward_shift.data().buffers()[0]
-                .count_set_bits_offset(is_forward_shift.offset(), is_forward_shift.len());
+        let num_filtered = is_forward_shift.len() - is_forward_shift.values().count_set_bits();
```
jordanrfrazier, line 336: Does this not need the  `offset` anymore?

crates/sparrow-runtime/src/execute/operation/shift_to.rs
```diff
@@ -451,8 +449,8 @@ impl ShiftToColumnOperation {
             let group_indices: &UInt32Array = downcast_primitive_array(group_indices.as_ref())?;
             // TODO: We shouldn't need to clone the array. Instead, we should be able
             // to pass an owned reference to the grouping indices. But... the clone
-            // shouldn't be too expensive, and that would require API hanges.
-            let group_indices: UInt32Array = UInt32Array::from(group_indices.data().clone());
+            // shouldn't be too expensive, and that would require API changes.
+            let group_indices: UInt32Array = UInt32Array::from(group_indices.to_data());
```
jordanrfrazier, line 453: Does the comment still apply?

crates/sparrow-runtime/src/merge/input.rs
```diff
@@ -297,9 +297,12 @@ mod tests {
         (1..max_len)
             .prop_flat_map(|len| (arb_key_triples(len), arb_i64_array(len)))
             .prop_map(|((time, subsort, key_hash), values)| {
-                let schema =
-                    TableSchema::from_data_fields([Field::new("data", DataType::Int64, true)])
-                        .unwrap();
+                let schema = TableSchema::from_data_fields([Arc::new(Field::new(
```
jordanrfrazier, line 300: I'm surprised you can't call `FieldRef::new(..)` as an alias for `Arc::new(Field::new(..))`.

crates/sparrow-runtime/src/execute/operation/shift_to.rs
bjchambers: Nope -- the new version of arrow returns a `NullBuffer` which has an internal offset and length. This is part of making these operations less error prone and easier for us to work with.

crates/sparrow-runtime/src/execute/operation/shift_to.rs
bjchambers: Yes it does -- `to_data` creates a copy of the data, and we create a new `UInt32Array` from it. It shouldn't be too expensive, but it there should still be possibilities of using an `ArrayRef` or something if it *was* expensive.

crates/sparrow-runtime/src/merge/input.rs
bjchambers: Yep. Not sure why Rust doesn't let you use the aliased name like that.  I think it's because:

1. You can't define a method on `FieldRef`, since that is really a set of types.
2. You could maybe call `FieldRef::new(...)` instead of `Arc::new(...)`, but that only works for the "outer-most thing" in the alias, which maybe could be confusing. In cases like this you may want to "`new`" both parts. Probably was easier to do the straightforward thing of just having you used the underlying / aliased type operations.

crates/sparrow-runtime/src/execute/operation/merge.rs
```diff
@@ -690,19 +690,19 @@ mod tests {
...
+        1970-01-01T00:00:00.000002000,0,1,1
+        1970-01-01T00:00:00.000003000,1,1,2
+        1970-01-01T00:00:00.000004000,0,2,3",
             None,
         )
         .unwrap();
         let operation_1 = batch_from_csv(
             "
         _time,_subsort,_key_hash,e0,e2
-        2500,0,2,1,3
```
bjchambers, line 702: The failing tests seem to have been caused by the CSV parser no longer accepting integers as timestamps. I don't believe this will affect user CSV files, since prepare should parse these as numbers and then cast the numbers, which will be a different path than trying to parse straight to timestamps.

jordanrfrazier (approved):

PR #420, opened 2023-06-06
Update aggregation-and-windowing.adoc

YoshiyukiKono:
I noticed that the table was rendered differently than you should intend because of the unnecessary cell, which I removed.


YoshiyukiKono:
I close this request so that I will create another one that should meet your intention more.

PR #424, opened 2023-06-06
feat: remove ack from pulsar reader

jordanrfrazier:
This removes the `ack` when reading messages from pulsar. The reason being that if we run a query twice, we want it to run over the same data in the stream instead of consuming the data. 

For materializations, we should rely on the subscription and broker to decide the offset at which to start reading (once we start saving state). Otherwise, each materialization should start at the beginning of the stream and produce deterministic results. Of course, this does depend on the configured lifetime of events in the stream, but that is expected. 

However, there is a small chance this breaks the pulsar demo that worked by preparing messages into parquet files then querying over those - since we're not acking, if we try to prepare multiple times, I wonder if we would be duplicated input data into the prepared files. But, unlikely this demo is used anymore, and we most likely want to phase this code path out anyway in favor of reading directly from the stream. Preparing streams to files has its place, but there's more work involved there regardless to have that a user-ready feature.

### DIFF
--- a/crates/sparrow-runtime/src/streams/pulsar/stream.rs
+++ b/crates/sparrow-runtime/src/streams/pulsar/stream.rs
@@ -196,13 +196,13 @@ impl PulsarReader {
 
             match msg {
                 Some(msg) => {
-                    // Note: Intentionally not acking here. 
-                    // In the materialization case, we should be using subscriptions to start from 
-                    // a specific offset. We do not want to ack messages in case we need to re-read 
-                    // messages from a time in the past to handle late data or backfills. 
+                    // Note: Intentionally not acking here.
+                    // In the materialization case, we should be using subscriptions to start from
+                    // a specific offset. We do not want to ack messages in case we need to re-read
+                    // messages from a time in the past to handle late data or backfills.
                     //
-                    // TODO: It's possible this breaks the specialized prepare demo - 
-                    // `prepare` may duplicate input data into the same spot. 
+                    // TODO: It's possible this breaks the specialized prepare demo -
+                    // `prepare` may duplicate input data into the same spot.
 
                     let result: error_stack::Result<AvroWrapper, DeserializeError> =
                         msg.deserialize();
--- a/crates/sparrow-runtime/src/streams/pulsar/stream.rs
+++ b/crates/sparrow-runtime/src/streams/pulsar/stream.rs
@@ -196,10 +196,14 @@ impl PulsarReader {
 
             match msg {
                 Some(msg) => {
-                    self.consumer
-                        .ack(&msg)
-                        .await
-                        .map_err(|e| ArrowError::from_external_error(Box::new(e)))?;
+                    // Note: Intentionally not acking here. 
+                    // In the materialization case, we should be using subscriptions to start from 
+                    // a specific offset. We do not want to ack messages in case we need to re-read 
+                    // messages from a time in the past to handle late data or backfills. 
+                    //
+                    // TODO: It's possible this breaks the specialized prepare demo - 
+                    // `prepare` may duplicate input data into the same spot. 
+
                     let result: error_stack::Result<AvroWrapper, DeserializeError> =
                         msg.deserialize();
                     let aw = match result {


epinzur:
agree that we should not worry about the prepare from pulsar path.

epinzur (approved):

bjchambers:
> This removes the `ack` when reading messages from pulsar. The reason being that if we run a query twice, we want it to run over the same data in the stream instead of consuming the data.
> 
> For materializations, we should rely on the subscription and broker to decide the offset at which to start reading (once we start saving state). Otherwise, each materialization should start at the beginning of the stream and produce deterministic results. Of course, this does depend on the configured lifetime of events in the stream, but that is expected.
> 
> However, there is a small chance this breaks the pulsar demo that worked by preparing messages into parquet files then querying over those - since we're not acking, if we try to prepare multiple times, I wonder if we would be duplicated input data into the prepared files. But, unlikely this demo is used anymore, and we most likely want to phase this code path out anyway in favor of reading directly from the stream. Preparing streams to files has its place, but there's more work involved there regardless to have that a user-ready feature.

This seems weird to me. For cases like this, what I generally expect:

1. If we need to start over, we start a new subscription, which starts at a given offset, etc.
2. If we restart (after reading a message, but before committing to state), we should reprocess the same message.

To accomplish this, systems like Beam and Flink send an `ack` for a given message once results derived from that message are persisted in state. But, it is generally important to send the `ack`...

jordanrfrazier:
> > This removes the `ack` when reading messages from pulsar. The reason being that if we run a query twice, we want it to run over the same data in the stream instead of consuming the data.
> > For materializations, we should rely on the subscription and broker to decide the offset at which to start reading (once we start saving state). Otherwise, each materialization should start at the beginning of the stream and produce deterministic results. Of course, this does depend on the configured lifetime of events in the stream, but that is expected.
> > However, there is a small chance this breaks the pulsar demo that worked by preparing messages into parquet files then querying over those - since we're not acking, if we try to prepare multiple times, I wonder if we would be duplicated input data into the prepared files. But, unlikely this demo is used anymore, and we most likely want to phase this code path out anyway in favor of reading directly from the stream. Preparing streams to files has its place, but there's more work involved there regardless to have that a user-ready feature.
> 
> This seems weird to me. For cases like this, what I generally expect:
> 
>     1. If we need to start over, we start a new subscription, which starts at a given offset, etc.
> 
>     2. If we restart (after reading a message, but before committing to state), we should reprocess the same message.
> 
> 
> To accomplish this, systems like Beam and Flink send an `ack` for a given message once results derived from that message are persisted in state. But, it is generally important to send the `ack`...

I'm not sure if it's different in Beam or Flink, but if we `ack` a message on a particular Pulsar stream, that means that we are forever removing that message from the stream, preventing other queries/consumers from accessing items from that stream.

This certainly shouldn't be done if we are to support regular queries (exploration/training) from a stream, and I don't think it makes sense in the materialization case either, as we may want multiple materializations producing different results from the same input stream.

bjchambers:
I believe that the `ack` only applies to your consumer. Eg., when a pulsar stream has N consumers, each message will be delivered to each consumer. If a consumer `acks` a message, the other consumers will still see the message. It is only after all N consumers have `acked` the message that the message will be considered `ack`ed. Additionally, Pulsar has options for retaining messages so that *future* consumers will still see a message.

We shouldn't be thinking of this as "acking on the stream", but rather "this consumer acknowledge that it has consumed the message from the stream", but it doesn't speak to other consumers.

bjchambers:
From https://pulsar.apache.org/docs/3.0.x/concepts-messaging/#acknowledgment

> The consumer sends an acknowledgment request to the broker after it consumes a message successfully. Then, this consumed message will be permanently stored, and deleted only after all the subscriptions have acknowledged it.

Each materialization should be a separate subscription. Our materialization consumes a message (and acknowledges it) once it has been consumed successfully -- for us that means the result of that message is persisted in state. The message will continue to be stored until **"all other subscriptions have acknowledged it."**

jordanrfrazier:
> From https://pulsar.apache.org/docs/3.0.x/concepts-messaging/#acknowledgment
> 
> > The consumer sends an acknowledgment request to the broker after it consumes a message successfully. Then, this consumed message will be permanently stored, and deleted only after all the subscriptions have acknowledged it.
> 
> Each materialization should be a separate subscription. Our materialization consumes a message (and acknowledges it) once it has been consumed successfully -- for us that means the result of that message is persisted in state. The message will continue to be stored until **"all other subscriptions have acknowledged it."**

https://pulsar.apache.org/docs/3.0.x/concepts-messaging/#message-retention-and-expiry

"By default, Pulsar message brokers:
    immediately delete all messages that have been acknowledged by a consumer"

There is a message retention policy that allows the topic to keep messages that are acknowledged. This means that if there is only a single subscriber on a topic at a time, then we'll delete the messages once they're acknowledged, based on the documentation. Doesn't this read as:

1. Create topic
2. Create materialization reading from topic
3. Ack messages 1-10 from topic
4. Messages 1-10 are acked by only existing subscriber, so delete them
5. Create materialization 2 reading from same topic
6. Materialization 2 sees messages 10...n

bjchambers:
I don't think we plan to rely on our first subscription existing and thus causing retention.

Consider a few things:

1. Any message sent before we start a materialization will be sent when there are no subscriptions, so will be deleted / dropped.
2. If we want Pulsar to retain things forever, we should use Pulsar tiered storage, which is *designed* for this. Not sending the ACK will cause Pulsar to keep it in the "active" path forever, rather than being able to reclaim that memory, etc.

jordanrfrazier:
> I don't think we plan to rely on our first subscription existing and thus causing retention.
I'm not sure what you mean by this. 

> Any message sent before we start a materialization will be sent when there are no subscriptions, so will be deleted / dropped

You mean if it is deleted/dropped due to it not being read before its lifetime expires?

> we should use Pulsar tiered storage

if all users enabled this, we can query over historical data that is stored in long-term storage as if it were still in the stream. As long as we're okay with the current `ack and drop` behavior. I think that I'm unclear on to what your main pushback on this change is - is it because we're not acting as a "consumer" on a stream should act like if we don't ack messages when we read them? 

We won't be able to support querying over pulsar streams unless tiered storage (or some other mechanism) is enabled, if we leave it as is.

epinzur:
I'm also confused by Ben's most recent comments, but I'll start a different thread...

Everytime we make a new materialization, we should use a new subscription-id (ideally based off the materialization id), and then starting reading from the start of the topic.  If we need to restart the materialization, we should restore state from the previous compute snapshot, and then start re-consuming messages.

I think this means that we only want to ack messages that have been successfully persisted to a compute snapshot.

Until we this code path supports resuming from snapshots, it might be best to not ack read messages.  Therefore when we restart processing a materialization, we can re-read from the beginning of the topic to restore state.

epinzur:
every query needs to use a different subscription id.

jordanrfrazier:
So, closing this for now. I agree now that we shouldn't be removing the `ack`, since that's contrary to the behavior of a consumer from a stream. But, I do still believe that we would get the behavior we desire by removing the ack (i.e. queries can read from the streams independently and read all messages without removing any from the stream)

epinzur:
i think maybe you misunderstood: i think we should remove the ack for
now... but probably put it back when we start snapshotting compute state.

bjchambers:
I'm fine either way for now, but I think in general the above discussion is aligned.

1. The expectation for consumer behavior is to acknowledge messages.
2. Every query should use a different subscriber ID -- so other queries acknowledging a message or not shouldn't matter (beyond the affect it has on retention).
3. Until the first query is run (and thus the first subscriber ID is created) and/or if all subscribers disappear, message won't be retained. As a result, it is *not* correct for us to rely on "because we don't acknowledge the message will still be there". That is only true of messages that appear (or exist) after our subscriber ID is created, and until our subscriber ID is deleted.

That said -- it's OK to remove the ACK for now if we need for demos, etc. but we should figure out something that lets us implement computation as a "proper" consumer.

PR #430, opened 2023-06-10
Bump engine version to 0.8.0

kerinin:


epinzur:
We should first make sure we at least one integration test covering the API side of pulsar materializations before we create a release.

kerinin:
I don’t think that should block release given that this is new functionality that is unlikely to break existing users and we have users who are blocked waiting for a resolution to Pulsar streaming. If we can get an integration test written quickly that’s fine but supporting pulsar is high priority.

PR #433, opened 2023-06-14
Update Streaming Avro Record Output

kevinjnguyen:
PR with the progress made towards debugging the streaming implementation.

Changes:
* Reverted the revert for Publish Time. The implementation was correct, just missing one more step.
* Added the missing step for publish time. When reading the stream, a flag is passed to consider adding the publish time. 
  * In the execution phase, no publish time (false).
  * In the preparation phase, publish time (true).

### DIFF
--- a/crates/sparrow-runtime/src/metadata/raw_metadata.rs
+++ b/crates/sparrow-runtime/src/metadata/raw_metadata.rs
@@ -80,7 +80,17 @@ impl RawMetadata {
             }
             source_data::Source::PulsarSubscription(ps) => {
                 let config = ps.config.as_ref().ok_or(Error::PulsarSubscription)?;
-                Ok(Self::try_from_pulsar(config).await?.sparrow_metadata)
+                // The `_publish_time` is metadata on the pulsar message, and required
+                // by the `prepare` step. However, that is not part of the user's schema.
+                // The prepare path calls `try_from_pulsar` directly, so for all other cases
+                // we explicitly set the schema to not include the `_publish_time` column.
+                //
+                // The "prepare from pulsar" step is an experimental feature, and will
+                // likely change in the future, so we're okay with this hack for now.
+                let should_include_publish_time = false;
+                Ok(Self::try_from_pulsar(config, should_include_publish_time)
+                    .await?
+                    .sparrow_metadata)
             }
         }
     }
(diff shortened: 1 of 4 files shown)


crates/sparrow-arrow/src/avro/avro_arrow.rs
```diff
@@ -13,8 +13,13 @@ use std::sync::Arc;
 /// represents a row and contains tuples with a string field name and the corresponding Avro Value,
 /// and returns a Result containing a vector of Arrow arrays (Vec<ArrayRef>) if the
 /// conversion is successful, or an ArrowError if an error occurs during conversion.
-pub fn avro_to_arrow(values: Vec<Vec<(String, Value)>>) -> Result<Vec<ArrayRef>, ArrowError> {
-    (0..values[0].len())
+pub fn avro_to_arrow(values: Vec<Vec<(String, Value)>>, should_include_publish_time: bool) -> Result<Vec<ArrayRef>, ArrowError> {
+    let upper_limit = if should_include_publish_time {
```
bjchambers, line 17: For long term -- this logic seems error-prone. Rather than relying on offsets (and order of columns) and boolean flags, it seems like it would be better to use the schema and the presence/absence of a given field.

epinzur (approved):

crates/sparrow-runtime/src/execute/compute_executor.rs
```diff
@@ -245,7 +243,6 @@ impl ComputeExecutor {
 
                 Ok(ProgressUpdate::ExecutionComplete { compute_snapshots })
             };
-
```
bjchambers, line 248: Looks like a lot of unrelated changes -- removing empty lines? Probably don't want to include those changes.

crates/sparrow-runtime/src/metadata/raw_metadata.rs
```diff
@@ -171,6 +181,7 @@ impl RawMetadata {
     /// Create a `RawMetadata` from a Pulsar topic.
     pub(crate) async fn try_from_pulsar(
         config: &PulsarConfig,
+        should_include_publish_time: bool,
```
bjchambers, line 184: I think this can just be `include_publish_time` since it controls the behavior. I don't think `should_` adds much.

bjchambers (approved):
Seems fine for getting things moving.

Ideally, we'd either introduce an enum for the boolean or we'd simplify things so we didn't need to pass it (possibly deciding based on the presence of the `_publish_time` field in the schema that we need to add it). If we did the enum, it would let us write something `PublishTime::Add` or `PublishTime::None` rather than `true` and `false`, which would make it clearer what the booleans mean, especially in the cases of the functions that take multiple booleans.

PR #450, opened 2023-06-22
Update Pulsar Output Per Batch

kevinjnguyen:
Updates the pulsar output module to have the producer send the internal batch after processing every batch. See: https://docs.google.com/document/d/1gU8NV65ATliGd2-eG2G-mCX2mu8S3jY9m9x1wOVqsaw/edit?usp=sharing (June 22, 2023) for more info.

### DIFF
--- a/tests/integration/api/mat_pulsar_to_pulsar_test.go
+++ b/tests/integration/api/mat_pulsar_to_pulsar_test.go
@@ -26,7 +26,7 @@ type pulsarToPulsarTestSchema struct {
 	Count    int `json:"count"`
 }
 
-var _ = FDescribe("Materialization from Pulsar to Pulsar", Ordered, Label("pulsar"), func() {
+var _ = PDescribe("Materialization from Pulsar to Pulsar", Ordered, Label("pulsar"), func() {
 	var (
 		ctx                   context.Context
 		cancel                context.CancelFunc
--- a/tests/integration/api/mat_pulsar_to_pulsar_test.go
+++ b/tests/integration/api/mat_pulsar_to_pulsar_test.go
@@ -26,7 +26,7 @@ type pulsarToPulsarTestSchema struct {
 	Count    int `json:"count"`
 }
 
-var _ = PDescribe("Materialization from Pulsar to Pulsar", Ordered, Label("pulsar"), func() {
+var _ = FDescribe("Materialization from Pulsar to Pulsar", Ordered, Label("pulsar"), func() {
 	var (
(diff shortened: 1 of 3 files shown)


tests/integration/api/mat_pulsar_to_pulsar_test.go
```diff
@@ -26,7 +26,7 @@ type pulsarToPulsarTestSchema struct {
 	Count    int `json:"count"`
 }
 
-var _ = PDescribe("Materialization from Pulsar to Pulsar", Ordered, Label("pulsar"), func() {
+var _ = FDescribe("Materialization from Pulsar to Pulsar", Ordered, Label("pulsar"), func() {
```
jordanrfrazier, line 29: remove f

tests/integration/api/mat_pulsar_to_pulsar_test.go
```diff
@@ -200,11 +200,21 @@ var _ = PDescribe("Materialization from Pulsar to Pulsar", Ordered, Label("pulsa
...
+
+				msg2 := receivePulsarMessageWithTimeout(pulsarConsumer, ctx)
+				g.Expect(msg2).ShouldNot(BeNil())
+
+				var data2 pulsarToPulsarTestSchema
+				err = json.Unmarshal(msg2.Payload(), &data2)
+				g.Expect(err).ShouldNot(HaveOccurred())
+				g.Expect(data2.LastId).Should(Equal(4))
+				g.Expect(data2.LastTime).Should(Equal(1687303805000000000))
+				g.Expect(data2.Count).Should(Equal(1))
```
jordanrfrazier, line 216: Hm. We're getting a second message? I would've expected not, since we shouldn't be processing the second input row with the current late data handling logic. I wonder if this is because of our `send_batch` addition? But if the values here are correct, we are getting new results (although..still 1 count?)

kevinjnguyen: I believe expecting the second message is correct. When we start the test, we load two messages and the first is read at "should output initial results to pulsar". The second event is in the buffer. Then we load two more (3 + 4) events in the "load more data into the table" test. Since each event is over 1 second apart, the events 2 and 3 should be processed while 4 is left in the buffer.

jordanrfrazier: Ah, you're correct - didn't realize this was under the `load additional data` part

jordanrfrazier (approved):

PR #451, opened 2023-06-23
bug: fix materialization stop signal 

jordanrfrazier:
Fixes the stop signal bug by using a stream in scan operation that takes until the signal is received. 

Still keeps the pulsar to pulsar tests ignored, since I have this log line I haven't figured out yet:
```
2023-06-26T14:20:32.562464Z ERROR Error sending end event to channel - send failed because receiver is gone    

```

### DIFF
--- a/crates/sparrow-runtime/src/execute/operation/scan.rs
+++ b/crates/sparrow-runtime/src/execute/operation/scan.rs
@@ -28,7 +28,7 @@ pub(super) struct ScanOperation {
     projected_schema: SchemaRef,
     /// The input stream of batches.
     ///
-    /// Takes until the stop signal is received
+    /// Takes until the stop signal is received or the stream is empty.
     input_stream: Pin<Box<dyn Stream<Item = error_stack::Result<Batch, Error>> + Send>>,
     key_hash_index: KeyHashIndex,
     progress_updates_tx: tokio::sync::mpsc::Sender<ProgressUpdate>,
--- a/crates/sparrow-runtime/src/execute/operation/scan.rs
+++ b/crates/sparrow-runtime/src/execute/operation/scan.rs
@@ -27,10 +27,11 @@ pub(super) struct ScanOperation {
     /// The schema of the scanned inputs.
     projected_schema: SchemaRef,
     /// The input stream of batches.
+    ///
+    /// Takes until the stop signal is received
     input_stream: Pin<Box<dyn Stream<Item = error_stack::Result<Batch, Error>> + Send>>,
     key_hash_index: KeyHashIndex,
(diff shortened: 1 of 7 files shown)


epinzur (approved):

crates/sparrow-runtime/src/execute/operation/scan.rs
```diff
@@ -233,16 +222,38 @@ impl ScanOperation {
                 .change_context(Error::internal_msg("failed to create stream reader"))?
                 .map_err(|e| e.change_context(Error::internal_msg("failed to read batch")))
                 .boxed();
+
                 input_stream
             }
         };
 
+        // Streams until the stop signal has been received
```
kevinjnguyen, line 230: Does this mean that there are two usages here:
* Stop Signal is provided -> Streaming until stop signal
* Stop Signal is empty ->  Non-streaming scenario so read until input stream finishes

Do you mind updating the doc on this if so? This is the key piece that stops the stream so it's better to be over descriptive.

crates/sparrow-runtime/src/execute/output/pulsar.rs
```diff
@@ -154,14 +154,13 @@ pub(super) async fn write(
         // sends the current batch to avoid waiting indefinitely until BATCH_SIZE is achieved.
         //
         // This is not optimal and may raise performance concerns to send small batches.
+        tracing::debug!("Success. Buffered {num_rows} messages to pulsar");
```
kevinjnguyen, line 157: Good catch on fixing the log line to be above the producer send batch.

crates/sparrow-runtime/src/execute/operation/scan.rs
```diff
@@ -233,16 +222,38 @@ impl ScanOperation {
...
+        let input_stream = if let Some(mut stop_signal_rx) = stop_signal_rx {
+            input_stream
+                .take_until(async move {
```
kevinjnguyen, line 233: What are your thoughts between `take_until` vs. `take_while`? Here it may not matter as much since we are reading from the event stream, but the difference would be between inclusive or exclusive of the last event. 

Are we unable to use `take_while` because of the async move?

kevinjnguyen (approved):

crates/sparrow-runtime/src/execute/operation/scan.rs
jordanrfrazier: yeah `take_while` didn't work because a) it requires a `fut` and b) we're using a `watch` channel, where the only future available is on `changed()`. So we need to wait _until_ the watch channel sees a new value

PR #454, opened 2023-06-27
ref: Move hashing to sparrow-arrow

bjchambers:
Also add a `Hasher` that hashes into a re-usable buffer. This tends to speed up the hash operation. The `Hasher` supports hashing of structs and other types that we didn't previously support.

### DIFF
--- a/crates/sparrow-arrow/src/hasher.rs
+++ b/crates/sparrow-arrow/src/hasher.rs
@@ -1,0 +1,322 @@
+// Based on hashing code from datafusion.
+
+use ahash::RandomState;
+use arrow_array::cast::{
+    as_boolean_array, as_generic_binary_array, as_largestring_array, as_primitive_array,
+    as_string_array, as_struct_array,
+};
+use arrow_array::types::{ArrowDictionaryKeyType, Decimal128Type, Decimal256Type};
+use arrow_array::{
+    downcast_dictionary_array, downcast_primitive_array, Array, ArrayAccessor, DictionaryArray,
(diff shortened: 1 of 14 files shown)


crates/sparrow-arrow/src/hasher.rs
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    #[test]
+    fn test_hash_string_stability() {
+        let mut hasher = Hasher::default();
+        let array = StringArray::from(vec![Some("hello"), None, Some("world")]);
+        let array: &dyn Array = &array;
+
+        let hashes = hasher.hash_array(array).unwrap();
+        assert_eq!(hashes, &[1472103086483932002, 0, 8057155968893317866]);
```
bjchambers, line 281: Ideally, I'd delete the old hash code and just always use the `hasher`. We can then move to figuring out how to best manage the "hasher" to share state (I think ideally, we'd have a separate hasher for each thread, or a pool of allocated hashers, etc.).

Unfortunately, since the hashing is slightly different (hashing to `0` when null) this isn't immediately possible.

therapon (approved):

crates/sparrow-arrow/src/hasher.rs
bjchambers: I could make the `hash` kernel just use a thread-local `Hasher`. That would change the hashing a bit -- but that is likely OK. Thoughts on whether I should do that in this PR so that we only have one way of hashing things (and pick-up support for hashing structs)?

PR #457, opened 2023-06-27
feat: use explicit type variable in signatures

jordanrfrazier:
Step 1 of many to support generic types for collections. 

Adds type variable struct to signatures. 

Next step supports nested generics, like `Map<K,V>`

### DIFF
--- a/crates/sparrow-compiler/src/functions/general.rs
+++ b/crates/sparrow-compiler/src/functions/general.rs
@@ -4,15 +4,15 @@ use crate::functions::{Implementation, Registry};
 
 pub(super) fn register(registry: &mut Registry) {
     registry
-        .register("is_valid(input: any) -> bool")
+        .register("is_valid<T: any>(input: T) -> bool")
         .with_implementation(Implementation::Instruction(InstOp::IsValid));
 
     registry
-        .register("hash(input: key) -> u64")
+        .register("hash<K: key>(input: K) -> u64")
         .with_implementation(Implementation::Instruction(InstOp::Hash));
 
     registry
-        .register("with_key(key: key, value: any, const grouping: string = null) -> any")
+        .register("with_key<K: key, T: any>(key: K, value: T, const grouping: string = null) -> T")
         .with_implementation(Implementation::new_pattern({
             const MERGED_OP: &str = "(merge_join ?key_op ?value_op)";
             const MERGED_KEY: &str = const_format::formatcp!("(transform ?key_value {MERGED_OP})");
@@ -31,7 +31,7 @@ pub(super) fn register(registry: &mut Registry) {
         ));
 
     registry
-        .register("lookup(key: key, value: any) -> any")
+        .register("lookup<K: key, T: any>(key: K, value: T) -> T")
         .with_implementation(Implementation::new_pattern({
             // The operation of the lookup request depends on the key value.
             //
@@ -87,6 +87,6 @@ pub(super) fn register(registry: &mut Registry) {
         }));
 
     registry
-        .register("coalesce(values+: any) -> any")
+        .register("coalesce<T: any>(values+: T) -> T")
         .with_implementation(Implementation::Instruction(InstOp::Coalesce));
 }
--- a/crates/sparrow-compiler/src/types/inference.rs
+++ b/crates/sparrow-compiler/src/types/inference.rs
@@ -6,7 +6,7 @@ use arrow::datatypes::{DataType, Field, TimeUnit};
 use hashbrown::hash_map::Entry;
 use hashbrown::HashMap;
 use itertools::{izip, Itertools};
-use sparrow_syntax::{FenlType, Located, Resolved, Signature, TypeConstraint};
+use sparrow_syntax::{FenlType, Located, Resolved, Signature, TypeConstraint, TypeVariable};
 
 use crate::{DiagnosticBuilder, DiagnosticCode};
 
@@ -25,8 +25,7 @@ pub(crate) fn instantiate(
         "Arguments being instantiated should have been resolved against the signature"
     );
 
-    let mut types_for_constraints: HashMap<TypeConstraint, Vec<&Located<FenlType>>> =
-        HashMap::new();
+    let mut types_for_variable: HashMap<TypeVariable, Vec<&Located<FenlType>>> = HashMap::new();
 
     // Make sure the number of arguments are correct.
     match arguments.len().cmp(&parameters.types().len()) {
@@ -69,10 +68,12 @@ pub(crate) fn instantiate(
         izip!(arguments.iter(), &parameter_types).enumerate()
     {
         match (parameter_type.inner(), argument_type.inner()) {
-            (FenlType::Generic(constraint), _) => types_for_constraints
-                .entry(*constraint)
-                .or_default()
-                .push(argument_type),
+            (FenlType::Generic(type_var), _) => {
+                types_for_variable
+                    .entry(type_var.clone())
+                    .or_default()
+                    .push(argument_type);
+            }
             (_, FenlType::Error) => {
                 // No problem here -- the argument is an error, but we already
                 // reported it. Don't hide the actual error.
@@ -109,22 +110,39 @@ pub(crate) fn instantiate(
         }
     }
 
-    let solutions: HashMap<TypeConstraint, FenlType> = types_for_constraints
-        .into_iter()
-        .filter_map(|(constraint, argument_types)| {
+    let solutions: HashMap<TypeVariable, FenlType> = signature
+        .type_parameters
+        .iter()
+        .filter_map(|p| {
+            // This expect should never fail -- it would mean no arguments used the type variable
+            // Which would have been an invalid signature.
+            let argument_types = types_for_variable
+                .remove(&p.name)
+                .expect("unused type variable");
+
             if argument_types.is_empty() {
                 None
             } else {
+                debug_assert!(
+                    p.constraints.len() == 1,
+                    "only one constraint currently supported"
+                );
+                let constraint = p.constraints[0];
+
                 // Can't use `?` because we want Option<Result<...>>.
                 // This allows `filter_map` to filter out the `None` before the errors are
                 // collected.
                 Some(
                     solve_constraint(call, &constraint, &argument_types)
-                        .map(|unified| (constraint, unified)),
+                        .map(|unified| (p.name.clone(), unified)),
                 )
             }
         })
         .try_collect()?;
+
+    // types_for_variables should be empty, asserting that each parameter was assigned a defined
+    // type variable.
+    debug_assert!(types_for_variable.is_empty(), "unassigned type variables");
 
     let instantiated_arguments = parameter_types
         .iter()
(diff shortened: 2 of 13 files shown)


crates/sparrow-syntax/src/syntax/signature.rs
```diff
@@ -39,9 +44,44 @@ impl Signature {
...
+                if !constraint_to_names.contains_key(constraint) {
+                    constraint_to_names.insert(*constraint, type_var);
+                    type_var += 1;
+                }
+            }
+        }
+
+        // Check result for type constraint
+        if let FenlType::Generic(constraint) = result {
+            constraint_to_names.entry(constraint).or_insert(type_var);
```
bjchambers, line 70: This assigns a type variable for a constraint that appears only in the result (which currently shouldn't happen), but it doesn't actually change the result to use that type variable, nor does it change the parameters.

I think as part of this we can change `FenlType` to include a `TypeVariable`, but not allow `TypeConstraint` to be used as `FenlType` (so generics in a `FenlType` are always a reference to a type variable).

Then, we need to change the parameters and result to reference the assigned type parameter.

That in turn will impact how we do inference.

crates/sparrow-main/tests/e2e/basic_error_tests.rs
```diff
@@ -204,7 +204,7 @@ async fn test_illegal_cast_to_generic() {
           - "  --> Query:1:19"
           - "  |"
           - "1 | { m: Numbers.m as number }"
-          - "  |                   ^^^^^^ Unable to cast to type 'number'"
+          - "  |                   ^^^^^^ Unable to cast to type 'num'"
```
jordanrfrazier, line 207: This is obviously wrong. But shows that the user can write the same name as our built-in constraints. It's hard to determine whether it's a valid or invalid use of the string though, so I'm not sure what the best solution is for now. Perhaps just write out the full name of the constraint in the `Display` for the `TypeParameter`, and we work on better error reporting later.

crates/sparrow-compiler/src/types/inference.rs
```diff
@@ -25,6 +25,15 @@ pub(crate) fn instantiate(
         "Arguments being instantiated should have been resolved against the signature"
     );
 
+    let mut constraints_for_type_var: HashMap<TypeVariable, Vec<TypeConstraint>> = HashMap::new();
```
bjchambers, line 28: I'm not sure we need a mutable list of constraints for each type variable. Isn't that information directly avaliable in the type parameters? Similarly, I'm not sure why we still collect types for each type *constraint*.

What I'd expect this code to do:

1. Collect argument types used with each type variable. So `HashMap<TypeVariable, Vec<&Located<FenlType>>>`.
2. For each type variable, determine the concrete type that is the least-upper-bound of those types. So `HashMap<TypeVariable, FenlType>`.
3. For each type variable, make sure that solution satisfies the constraints. This can be done by iterating over the type parameters (since every variable should appear there) and checking the corresponding type against the constraint.

So, we need neither of the HashMaps that are here.

crates/sparrow-compiler/src/functions/general.rs
```diff
@@ -4,15 +4,15 @@ use crate::functions::{Implementation, Registry};
...
         .with_implementation(Implementation::Instruction(InstOp::IsValid));
 
     registry
-        .register("hash(input: key) -> u64")
+        .register("hash<K: key>(input: K) -> u64")
         .with_implementation(Implementation::Instruction(InstOp::Hash));
 
     registry
-        .register("with_key(key: key, value: any, const grouping: string = null) -> any")
+        .register("with_key<K: key, T: any>(key: K, value: T, const grouping: string = null) -> T")
```
therapon, line 15: I thought for this change we could only do 1 type parameter. This has 2

crates/sparrow-plan/src/inst.rs
```diff
@@ -237,7 +235,7 @@ pub enum InstKind {
     Cast(DataType),
     /// Create a record with the given field names.
     ///
-    /// The number of arguments should match the number of fields.
+    /// The N of arguments should match the N of fields.
```
therapon, line 238:
```suggestion
    /// The N arguments should match the N fields.
```

crates/sparrow-syntax/src/syntax/fenl_type.rs
```diff
@@ -101,12 +103,14 @@ pub enum TypeConstraint {
...
     Signed,
-    /// Any floating point numeric ytpe.
+    /// Any floating point numeric type.
     Float,
     /// Any time delta.
     TimeDelta,
     /// Any ordered type. This includes numbers and timestamps.
     Ordered,
+    /// Error type
+    Error,
```
therapon, line 113: This is for erroring out not that we have an Error Type in the fenl right?

crates/sparrow-syntax/src/syntax/fenl_type.rs
```diff
@@ -101,12 +103,14 @@ pub enum TypeConstraint {
...
     /// Any signed numeric type.
     Signed,
-    /// Any floating point numeric ytpe.
+    /// Any floating point numeric type.
     Float,
     /// Any time delta.
     TimeDelta,
     /// Any ordered type. This includes numbers and timestamps.
     Ordered,
+    /// Error type
```
therapon, line 112:
```suggestion
    /// Error -- variant to capture errors 
```

crates/sparrow-syntax/src/syntax/signature.rs
```diff
@@ -39,9 +88,23 @@ impl Signature {
             )
         }
 
+        // check that no duplicates exist in the type_parameters
```
therapon, line 91: Maybe I missed it, is there a check for using a type variable in the return position only? 
e.g., 

```
 foo<I: number, T: any>(in: I): T
```

crates/sparrow-syntax/src/syntax/fenl_type.rs
```diff
@@ -16,8 +18,8 @@ pub enum FenlType {
     // TODO: Rename concrete?
     /// A specific Arrow DataType.
     Concrete(DataType),
-    /// A generic type with the given type constraint.
-    Generic(TypeConstraint),
+    /// A generic type with the given type variable.
+    Generic(TypeVariable),
```
bjchambers, line 22: For a follow-up fixing naming, I think the `Generic` should be a `TypeRef` perhaps?

crates/sparrow-syntax/src/syntax/fenl_type.rs
```diff
@@ -90,7 +92,7 @@ impl<'a> std::fmt::Display for FormatStruct<'a> {
 /// an instantiation of the signature. This leads to a relatively simple
 /// instantiation strategy, where a least upper bound of the types of the
 /// actual arguments is chosen for constrained type.
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Serialize)]
+#[derive(PartialOrd, Ord, Clone, Copy, Debug, PartialEq, Eq, Hash, Serialize)]
 #[repr(u8)]
 pub enum TypeConstraint {
```
bjchambers, line 97: For follow up re: naming: I think this should be renamed to `TypeClass` to be more correct.

crates/sparrow-syntax/src/syntax/fenl_type.rs
bjchambers: This is for collecting multiple errors -- we need a way of recording "we've reported a failure, so don't report more issues with this".

crates/sparrow-syntax/src/syntax/fenl_type.rs
```diff
@@ -187,7 +202,8 @@ impl FromStr for FenlType {
             "duration_ns" => Ok(DataType::Duration(TimeUnit::Nanosecond).into()),
             "window" => Ok(FenlType::Window),
             "json" => Ok(FenlType::Json),
-            _ => Err(FenlType::Error),
+            // catch-all assumes the type is a type variable, which will be validated in the signature creation
```
bjchambers, line 205: is this still true?

crates/sparrow-syntax/src/syntax/signature.rs
```diff
@@ -1,7 +1,53 @@
...
+
+/// Contains the type variable and the constraints for
+/// that variable.
+///
+/// e.g. N(type variable): Eq(constraint) + Hash(constraint)
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(test, derive(serde::Serialize))]
+pub struct TypeParameter {
+    pub name: TypeVariable,
+    pub constraints: Vec<TypeConstraint>,
```
bjchambers, line 26: Looks like we already support multiple constraints?

crates/sparrow-syntax/src/syntax/signature.rs
```diff
@@ -1,7 +1,53 @@
...
+
+impl TypeParameter {
+    pub fn new(name: String, constraints: Vec<TypeConstraint>) -> Self {
+        Self {
+            name: TypeVariable(name),
+            constraints,
+        }
+    }
+
+    pub fn try_new(name: &str, constraint: &TypeConstraint) -> anyhow::Result<Self> {
```
bjchambers, line 44: What's the purpose of this? When would it ever fail?

crates/sparrow-syntax/src/syntax/signature.rs
```diff
@@ -39,9 +88,23 @@ impl Signature {
...
+        for type_param in &type_parameters {
+            if !names.insert(&type_param.name.0) {
+                anyhow::bail!(
+                    "Duplicate type parameter '{}' in signature for '{}'",
+                    type_param.name.0,
+                    name
+                )
+            }
+        }
+        std::mem::drop(names);
```
bjchambers, line 102: I tend to avoid `std::mem::drop` for this "eagerly drop it". Why not put it in a block? Also, it will be dropped as soon as this function returns, which is the next thing that happens, so ti doesn't seem to add much.

crates/sparrow-syntax/src/syntax/signature.rs
```diff
@@ -39,9 +88,23 @@ impl Signature {
             )
         }
 
+        // check that no duplicates exist in the type_parameters
+        let mut names = HashSet::new();
```
bjchambers, line 92:
```
let duplicates: Vec<_> = type_parameters.iter().map(|p| p.name.0).duplicates().collect();
if !duplicates.is_empty() {
  anyhow::bail!("Duplicate type parameters: {}" in signature for '{name}'",
    duplicates.iter().format_with(",", |elt, f| f(&format_args!("'{elt}'")));
}
```

crates/sparrow-compiler/src/types/inference.rs
```diff
@@ -25,8 +25,16 @@ pub(crate) fn instantiate(
         "Arguments being instantiated should have been resolved against the signature"
     );
 
-    let mut types_for_constraints: HashMap<TypeConstraint, Vec<&Located<FenlType>>> =
-        HashMap::new();
+    let mut constraints_for_type_var: HashMap<TypeVariable, Vec<TypeConstraint>> = HashMap::new();
```
bjchambers, line 28: I still don't understand why we need to collect this as a hash map. Why not just wait until we have the types for each variable, and then later iterate over the `type_parameters` themselves. As far as I can tell, this loop just creates a copy of the type parameters keyed by variable (duplicating the assertions you have elsewhere).

crates/sparrow-compiler/src/types/inference.rs
```diff
@@ -109,18 +119,39 @@ pub(crate) fn instantiate(
         }
     }
 
-    let solutions: HashMap<TypeConstraint, FenlType> = types_for_constraints
+    let solutions: HashMap<TypeVariable, FenlType> = types_for_variable
```
bjchambers, line 122: Rather than running find over the parameters every time, I think you could iterate over the parameters and pair that with the type variables.

```
let solutions = signature.type_parameters.iter().map(|p| {
  // This expect should never fail -- it would mean no arguments used the type variable
  // Which would have been an invalid signature.
  let argument_types = types_for_variable.remove(p.name).expect("unused type variable");
  ...
}).try_collect()?;

// Then check to see if types_for_variable is empty -- if not, then there is a different error.
// Not sure it needs a diagnostic though since it means we had a type variable in the
// signature that wasn't defined.
```

crates/sparrow-syntax/src/syntax/signature.rs
bjchambers: On the same vein, a check for using an undefined type variable:

```
foo<I: number>(in: N): N
```

crates/sparrow-compiler/src/types/inference.rs
jordanrfrazier: You're correct - I missed removing this. This is unused.

crates/sparrow-main/tests/e2e/aggregation_tests.rs
```diff
@@ -9,7 +9,8 @@ use crate::fixtures::{
 };
 use crate::QueryFixture;
 
-#[tokio::test]
+#[tokio::main]
+#[test]
```
jordanrfrazier, line 13: fix

bjchambers (approved):

crates/sparrow-compiler/src/functions/general.rs
jordanrfrazier: We can do any number for this change, just not nested, like `C: Collection<K, V>`

crates/sparrow-syntax/src/syntax/signature.rs
jordanrfrazier: We allow the ability to have multiple here, but we don't have the logic for finding the LUB for multiple at the moment. That will come in a follow up

crates/sparrow-syntax/src/syntax/signature.rs
jordanrfrazier: This is used in `grammar.lalrpop`. It used to be able to fail when I was passing `FenlType` but now that I explicitly use `TypeConstraint` I suppose I can just remove this.

PR #458, opened 2023-06-28
feat: Kafka Backed Sources API Proposal

kevinjnguyen:
# API Proposal

Adds Kafka backed sources to the API. Most of the configurations are based on the Rust [client](https://docs.rs/kafka/latest/kafka/consumer/struct.Builder.html).


proto/kaskada/kaskada/v1alpha/kafka.proto
```diff
@@ -0,0 +1,64 @@
...
+
+  // TODO: Additional security config constraints
+  message SecurityConfig {
+    // Should perform hostname validation. Defaults to true.
+    bool hostname_validation = 1;
+  }
+
+  // The connection configuration
+  // TODO: Define these defaults and document them here.
+  message ConnectionConfig {
```
epinzur, line 27: doesn't seem like this is used.

proto/kaskada/kaskada/v1alpha/kafka.proto
```diff
@@ -0,0 +1,64 @@
...
+// Configuration for a single source of data from a Kafka topic.
+message KafkaSubscription {
+  // The Kafka config
+  KafkaConfig config = 1;
+
+  // The group on whose behalf to maintain consumed message offsets.
+  string group = 2;
+
+  // The specific offset to use. Defaults to LATEST.
+  FetchOffset offset = 3;
```
epinzur, line 52: maybe omit this, and other options until they are asked for? to start only include the minimum config properties?

proto/kaskada/kaskada/v1alpha/sources.proto
```diff
@@ -2,12 +2,14 @@ syntax = "proto3";
```
epinzur, line 2: I would have expected to see a similar update to `destinations.proto`

proto/kaskada/kaskada/v1alpha/kafka.proto
epinzur: also i think for our use, we would want to default to EARLIEST, so we can build up history from the start of any existing data on the topic.

proto/kaskada/kaskada/v1alpha/common.proto
```diff
@@ -24,6 +25,8 @@ message SourceData {
     string csv_data = 3;
 
     PulsarSubscription pulsar_subscription = 4;
+
+    KafkaSubscription kafka_subscription = 5;
```
epinzur, line 29: it is a bit odd that we need a `Subscription` message for the `getMetadata()` call.  

I think this is because the `SourceData` message is used for both `getMetadata()` and `prepareFile()`

Maybe instead of adding the KafkaSubscription here, we change `GetMetadataRequest` to be a one-of `SourceData`, `PulsarConfig`, or `KafkaConfig`?

proto/kaskada/kaskada/v1alpha/kafka.proto
kevinjnguyen: good catch, updated.

proto/kaskada/kaskada/v1alpha/kafka.proto
kevinjnguyen: Sure, we can use earliest.

proto/kaskada/kaskada/v1alpha/sources.proto
kevinjnguyen: I'm aiming for read only from kafka first and destinations can come after.

proto/kaskada/kaskada/v1alpha/common.proto
epinzur: thoughts on this @kevinjnguyen ?

epinzur (approved):

proto/kaskada/kaskada/v1alpha/common.proto
kevinjnguyen: I added this in the subsequent changes.

proto/kaskada/kaskada/v1alpha/file_service.proto
```diff
@@ -10,7 +12,11 @@ message SourceMetadata {
 }
 
 message GetMetadataRequest {
-  SourceData source_data = 1;
+  oneof source {
+    SourceData source_data = 1;
+    PulsarSubscription pulsar_subscription = 2;
+    KafkaSubscription kafka_subscription = 3;
```
epinzur, line 18: I don't think these should be subscriptions... these should be configs:
```
PuslarConfig ...
Kafka Config ...
```

epinzur (changes requested):
change from using subscriptions to configs

proto/kaskada/kaskada/v1alpha/preparation_service.proto
```diff
@@ -2,10 +2,16 @@ syntax = "proto3";
...
+import "kaskada/kaskada/v1alpha/pulsar.proto";
 
 message PrepareDataRequest {
   // The source of information for the prepare data
-  SourceData source_data = 1;
+  oneof source {
+    SourceData source_data = 1;
+    PulsarSubscription pulsar_subscription = 6;
+    KafkaSubscription kafka_subscription = 7;
+  }
```
epinzur, line 14: for prepare, I don't think we need to support Puslar or Kafka.  I would just undo this change.

epinzur (approved):

Issue #459, opened 2023-06-28
bug: `make wren/run` doesn't work


bjchambers:
**Description**
As stated in CONTRIBUTING.md, the command `make wren/run` fails with the following error:
"unable to open database file: no such file or directory"
exit status 1
make: *** [wren/run] Error 1

This seems to be because the directory it creates the SQLite DB in doesn't exist. I fixed this by running `mkdir tests/integration/data`.

We should probably fix the documentation and/or ensure the directory exists before attempting to create the DB.

bjchambers:
Also instructions for running integration tests using docker seem broken.

```make test/int/docker-up``` 

produces the following errors:

```
kaskada  | /bin/wren: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /bin/wren)
kaskada exited with code 1
```

bjchambers:
Also, once the local binaries are running, things still fail since a Pulsar is necessary as well.

jeffbanks:
Thanks @bjchambers for opening this issue. I ran into the same Pulsar missing issues today trying to setup a local `dev` environment.

epinzur:
#490 should fix the creation of the sqlite and objectStore directories, and the GLIBC issue on Apple Silicon.

PR #461, opened 2023-06-28
bug: manager make directory for sqlite

kevinjnguyen:
Fixes one issue in #459. Updates the dbpath logic to merge the provided db path with the current path to produce an absolute path using the `abs` ([docs](https://pkg.go.dev/path/filepath#Abs)).

Previously, we were only checking for the home directory and not necessarily worrying about any other use cases. The logic proposed here should merge the two flows.

### DIFF
--- a/wren/client/ent_client.go
+++ b/wren/client/ent_client.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"os/user"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


wren/client/ent_client.go
```diff
@@ -67,13 +66,11 @@ func NewEntConfig(dbDialect string, dbName *string, dbHost *string, inMemory *bo
 		log.Fatal().Msg("Flag: `db-dialect` must be set to either `postgres` or `sqlite`.")
 	}
 
-	if dbPath != nil && strings.HasPrefix(*dbPath, "~/") {
-		usr, err := user.Current()
+	if dbPath != nil {
+		path, err := filepath.Abs(*dbPath)
```
bjchambers, line 70: I'm not sure this will fix the problem -- for me, the directory was `tests/integration/data`, but only `tests/integration` existed. `filepath.Abs` doesn't seem to do anything about creating the directory.

Or is the problem that since the default path didn't start with `~/`, the `strings.HasPrefix(...)` above caused us not to run the normalization logic (the `MkdirAll` below)?

bjchambers: If it's the latter -- this looks good but update the PR description to explain the actual fix (eg., always creating the directory fixes the problem, not calling `Abs`).

bjchambers (approved):

wren/client/ent_client.go
kevinjnguyen: Its the latter. I'll update the PR. Thanks for taking a look!

PR #463, opened 2023-06-29
feat: expression evaluation for new plans

bjchambers:
This introduces the basic evaluator mechanisms for the new physical plans. As part of this, some aspects of the physical expression representation changed. Specifically, allowing a function call to have zero or more literal arguments allowed every expression to be "simple".

This also copied most of the existing evaluators into the new API. This was done by copying to avoid overhead involved in refactoring. The plan is to get the new runtime path based on these working (and passing the same tests) and then delete the old path.

This doesn't copy over some of the scalar functions (comparisons) or all of the time functions. Those can be ported over in the future once the runtime is in place.

This also doesn't copy over any stateful operations (aggregations) yet since aggregation may need to use a different approach to interface with the physical plans.


bjchambers:
Suggested review order:

1. Review the change to `expr.rs` -- this changes how expressions are represented in physical plans, specifically to allow literal arguments.
2. Review everything in `sparrow-expressions` *except* the contents of the `evaluators` directory. This represents the machinery to actually create the expression evaluator for a physical plan and run it.
3. Review some (or all) of the evaluators. Each file should be self contained for a single instruction.
4. Everything outside of `sparrow-expressions` is just a few typos I noticed while making these changes.

crates/sparrow-expressions/src/evaluators.rs
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+}
+
+/// Create the evaluators for the given expressions.
+pub(super) fn create_evaluators(
+    input_schema: &Schema,
+    exprs: &[Expr],
+) -> error_stack::Result<Vec<Box<dyn Evaluator>>, Error> {
+    let mut expressions = IndexVec::with_capacity(exprs.len());
```
therapon, line 145: Maybe comment here on what `expressions` does.

crates/sparrow-expressions/src/evaluators/cast.rs
```diff
@@ -0,0 +1,112 @@
...
+use crate::values::ArrayRefValue;
+use crate::work_area::WorkArea;
+use crate::Error;
+
+inventory::submit!(EvaluatorFactory {
+    name: "cast",
+    create: &create,
+});
+
+/// Evaluator for unary negation.
```
therapon, line 22: I think this comment is wrong? 
This is for casting right?

crates/sparrow-expressions/src/evaluators/column.rs
```diff
@@ -0,0 +1,32 @@
...
+use crate::evaluator::Evaluator;
+use crate::work_area::WorkArea;
+use crate::Error;
+
+inventory::submit!(crate::evaluators::EvaluatorFactory {
+    name: "column",
+    create: &create
+});
+
+/// Evaluator for addition.
```
therapon, line 12:
```suggestion
/// Evaluator for column reference `table.c`.
```

crates/sparrow-expressions/src/evaluators/comparison/lt_primitive_scalar.rs
therapon, line 1: Left over file?

crates/sparrow-expressions/src/evaluators/json_field.rs
```diff
@@ -0,0 +1,154 @@
...
+    name: "json_field",
+    create: &create
+});
+
+/// Evaluator for `json_field` expressions.
+///
+/// This evaluator expects a `string` input and `field` name.
+/// It parses the `string` into a `json` object, then outputs the
+/// value of the `field` name as a `string`, or `null` if the `field`
+/// does not exist.
```
therapon, line 18: From what I can tell this code deals with a single JSON object and a path of length 1. So the for example following is *not* supported 

```
// JSON, path
{ a: { b : 1}, c: 10} , a.b
```

But I think this works? 

```
// JSON, path
{ a: { b : 1}, c: 10} , a
```

crates/sparrow-expressions/src/evaluators/math/clamp.rs
```diff
@@ -0,0 +1,172 @@
...
+use crate::values::PrimitiveValue;
+use crate::work_area::WorkArea;
+use crate::Error;
+
+inventory::submit!(crate::evaluators::EvaluatorFactory {
+    name: "clamp",
+    create: &crate::evaluators::macros::create_primitive_evaluator!(0, create, ordered)
+});
+
+/// Evaluator for Round.
```
therapon, line 17:
```suggestion
/// Evaluator for clamp instruction.
```

crates/sparrow-expressions/src/evaluators/math/greatest.rs
```diff
@@ -0,0 +1,93 @@
...
+use crate::values::PrimitiveValue;
+use crate::work_area::WorkArea;
+use crate::Error;
+
+inventory::submit!(crate::evaluators::EvaluatorFactory {
+    name: "greatest",
+    create: &crate::evaluators::macros::create_primitive_evaluator!(0, create, number)
+});
+
+/// Evaluator for unary negation.
```
therapon, line 17:
```suggestion
/// Evaluator for greatest (max).
```

crates/sparrow-expressions/src/evaluators/math/least.rs
```diff
@@ -0,0 +1,90 @@
...
+use crate::values::PrimitiveValue;
+use crate::work_area::WorkArea;
+use crate::Error;
+
+inventory::submit!(crate::evaluators::EvaluatorFactory {
+    name: "least",
+    create: &crate::evaluators::macros::create_primitive_evaluator!(0, create, number)
+});
+
+/// Evaluator for unary negation.
```
therapon, line 17:
```suggestion
/// Evaluator for least (min) negation.
```

crates/sparrow-expressions/src/evaluators.rs
bjchambers:
```
    // Static information (index in expressions, type, etc.) for each expression in `exprs`.
    // This is used to locate the information about arguments to the remaining expressions.
    //
    // It is only needed while instantiating the evaluators.
```

crates/sparrow-expressions/src/evaluators/comparison/lt_primitive_scalar.rs
bjchambers: placeholder for future... I'll remove for now

crates/sparrow-expressions/src/evaluators/json_field.rs
bjchambers: Yeah, you can do `json_field(json_field(..., "a"), "b")` to do nesting. Ideally we'd do more with paths and/or parsing these into real JSON objects for manipulation instead of going back and forth with strings, but this was initial / rudimentary support.

bjchambers: I think @jordanrfrazier may be able to comment and/or we should import some more notes on this -- there was a plan for how this could evolve to reduce the parsing. I think it was something like if we had a JSON object representation, then this could be `json_field(s, "field") = parse_json(s).field`, but I could be mistaken.

crates/sparrow-expressions/src/evaluators.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    pub fn string(&self) -> error_stack::Result<StringValue, Error> {
+        StringValue::try_new(self.index, self.data_type)
+    }
+
+    pub fn array_ref(&self) -> ArrayRefValue {
+        ArrayRefValue::new(self.index)
+    }
+
+    pub fn struct_(&self) -> error_stack::Result<StructValue, Error> {
```
jordanrfrazier, line 78: `struct_`?

crates/sparrow-expressions/src/evaluators/coalesce.rs
```diff
@@ -0,0 +1,89 @@
+use arrow_array::{ArrayRef, BooleanArray};
+use error_stack::{IntoReport, ResultExt};
+
+use crate::evaluator::Evaluator;
+use crate::evaluators::StaticInfo;
+use crate::values::ArrayRefValue;
+use crate::Error;
+
+inventory::submit!(crate::evaluators::EvaluatorFactory {
```
jordanrfrazier, line 9: as a note, might be easy to add a readme for future contributors on how to create new evaluators with this pattern

jordanrfrazier (approved):

.vscode/settings.json
```diff
@@ -1,3 +1,4 @@
 {
     "rust-analyzer.cargo.features": "all",
+    "rust-analyzer.showUnlinkedFileNotification": false,
```
jordanrfrazier, line 3: want this?

bjchambers: Not sure. Removed.

crates/sparrow-expressions/src/evaluators.rs
bjchambers: Heh. Yeah. `struct` is a reserved word. There is a weird syntax for using that as a name, but that seems to be frowned upon. You *can* do `r#struct` (which uses a raw identifier and bypasses the reserved word), but based on various things it seems like it is easier / preferred to just use an `_` to make it not a reserved word.

https://www.reddit.com/r/rust/comments/bb3ysz/comment/ekga8in/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

We *could* call this `record`, but that would be better if we called it record everywhere -- we generally use `record` when we talk about Fenl and `struct` when we talk about Arrow (to match `StructArray`) so this felt like the easiest option.

crates/sparrow-expressions/src/evaluators/coalesce.rs
bjchambers: Added the following to `lib.rs` in this crate:

```
//! Execution of expressions.
//!
//! Creating an [ExpressionExecutor] takes a sequence of
//! [expressions][sparrow_physical::Expr] and applies them to an input batch to
//! produce a sequence of computed columns. Expressions may reference columns
//! from the input to the expressions (typically the input to the step executing
//! the expressions), create a column from a literal value, or apply an
//! expression to columns computed by earlier expressions.
//!
//! # Adding an expression
//!
//! Adding an evaluator for a new kind of expression requires the following steps
//! (more details follow):
//!
//! 1. In a file for the evaluator, create a struct implementing the evaluator.
//!    This should have fields for the information necessary to execute the
//!    expression.
//! 2. Implement [evaluator::Evaluator] for the evaluator struct.
//! 3. Write a method `create` which takes [evaluators::StaticInfo] and returns
//!    an `error_stack::Result<Box<dyn Evaluator>, Error>`.
//! 4. Register the creation method with inventory of supported expressions.
//!
//! ## Fields in the Evaluator
//!
//! For each input column the evaluator operates on, it should have a
//! [work_area::WorkAreaValue]. This represents a type-safe handle for accessing
//! the input column from the work area during runtime. These values are
//! obtained from the `StaticInfo` by calling methods such as
//! [evaluators::StaticArg::primitive].
//!
//! Depending on the evaluator, it may also be necessary to have a primitive type
//! parameter, as can be seen with the various math operations.
//!
//! ## Implementing Create
//!
//! The create method should take the static info, verify that the types are
//! correct and the expression can be executed, and then create the evaluator
//! struct. Doing so also requires populating the values that will be referenced
//! at runtime. Usually, this population does the necessary type checking.
//!
//! ## Register the creation method
//!
//! Registering the evaluator is done using the following block of code.
//! This makes use of the `inventory` crate so evaluators may be registered
//! in other modules and discovered automatically.
//!
//! ```
//! inventory::submit!(crate::evaluators::EvaluatorFactory {
//!     name: "not",
//!     create: &create
//! });
//! ```
//!
//! In cases where the `create` method has a generic parameter, the
//! `create_primitive_evaluator!` macro can be used to create an implementation
//! of `create` for the EvaluatorFactory that matches on the given data type. In
//! the following registration, the first argument (position `0`) is used to
//! determine the type of evaluator to create. The *actual* `create` method
//! is called with the necessary data type.
//!
//! The final argument indicates what types should be supported by the match.
//! It can be one or more pairs of the form `(ty_case, ty_name), ...`, but is
//! more typically one of the following defined types (corresponding to the
//! equivalent Fenl type classes):
//!
//! - `number`: Supports `i32`, `i64`, `u32`, `u64`, `f32` and `f64`.
//! - `signed`: Supports `i32`, `i64`, `f32` and `f64`.
//! - `float`: Supports `f32` and `f64`.
//! - `ordered`: Supports `number` and the timestamp types.
//!
//! ```
//! inventory::submit!(crate::evaluators::EvaluatorFactory {
//!     name: "add",
//!     create: &crate::evaluators::macros::create_primitive_evaluator!(0, create, number)
//! });
//! ```
```

Issue #427, opened 2023-06-07
bug: canary helm chart is broken


epinzur:
**Description**
When deployed to a Kubernetes cluster, the current helm chart does not deploy a working instance of kaskada.

**Actual Behavior**
The current deploy creates a pod like the following:
```
Name:             kaskada-kaskada-canary-86f64fbb85-842vb
Namespace:        kaskada
Service Account:  kaskada-kaskada-canary
Labels:           [app.kubernetes.io/instance=kaskada](http://app.kubernetes.io/instance=kaskada)
                  [app.kubernetes.io/name=kaskada-canary](http://app.kubernetes.io/name=kaskada-canary)
                  pod-template-hash=86f64fbb85
Annotations:      <none>
Containers:
  engine:
    Image:         [ghcr.io/kaskada-ai/kaskada/engine:latest](http://ghcr.io/kaskada-ai/kaskada/engine:latest)
    Port:          50052/TCP
    Command:
      /bin/sparrow-main
      serve
    Environment:
      SPARROW_LOG_FILTER:  egg::=warn,sparrow_=info,info
      SPARROW_LOG_JSON:    true
    Mounts:
      /tmp from kaskada-canary-tmp (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-md4wg (ro)
  manager:
    Container ID:  containerd://0e0eff115d3438446886114adf6f02de64043d92e0fae43e871a30959f87bc28
    Image:         [ghcr.io/kaskada-ai/kaskada/engine:latest](http://ghcr.io/kaskada-ai/kaskada/engine:latest)
    Ports:         50051/TCP, 3365/TCP
    Command:
      /bin/wren
    Environment:
      DB_DIALECT:           sqlite
      DB_PATH:              /data/kaskada.db
      DB_IN_MEMORY:         false
      OBJECT_STORE_BUCKET:
      OBJECT_STORE_PATH:    /
      OBJECT_STORE_TYPE:    local
      DEBUG:                false
      LOG_FORMAT_JSON:      true
    Mounts:
      /data from kaskada-canary-db (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-md4wg (ro)
Volumes:
  kaskada-canary-db:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  kaskada-canary-db
    ReadOnly:   false
  kaskada-canary-tmp:
[...]

PR #464, opened 2023-06-30
draft: various helm chart improvements

epinzur:
* fix running without object storage (confirmed via test in EKS)
* add access-key/secret-id support for AWS S3 access (confirmed via test in EKS)
* progress on AWS IRSA s3 access
  * we need to make a code change to wren (and maybe sparrow) before I can confirm that IRSA access is working. See #527

fixes: #427


charts/kaskada-canary/README.md
```diff
@@ -15,50 +15,3 @@ Add the following repo to use the chart:
 ```console
 helm repo add kaskada https://kaskada-ai.github.io/kaskada/
 ```
-
-## Values
```
epinzur, line 19: too hard to keep this `values` documentation up-to-date in two places.  instead documented on the file itself.

charts/kaskada-canary/tests/__snapshot__/default_test.yaml.snap
```diff
@@ -0,0 +1,152 @@
+manifest should match snapshot:
```
epinzur, line 1: these snapshots are the "proof" that the chart is working as designed.  we test against these in CI.

kerinin (approved):

PR #470, opened 2023-07-05
ref: update fenl type namings

jordanrfrazier:


crates/sparrow-compiler/src/types/inference.rs
```diff
@@ -543,7 +543,7 @@ fn least_upper_bound_data_type(a: DataType, b: &DataType) -> Option<DataType> {
 /// If the `concrete` type already satisfies this constraint, it is
```
bjchambers, line 543: Maybe look at some of the comments here and/or search. Eg., "this type already satisfies this constraint" should be "this type is part of the type class", etc.

bjchambers (approved):

crates/sparrow-compiler/src/types/inference.rs
jordanrfrazier: Yeah, I'll probably add this to the next PR since I missed many in comments

Issue #465, opened 2023-06-30
feat: Use object store and async, byte-range reads


bjchambers:
**Summary**

Currently, compute uses the S3 client to retrieve Parquet files before reading them. We have started a transition to using https://docs.rs/object_store/latest/object_store/ which supports (a) reading from multiple object stores and (b) doing a direct byte-range read without fetching the file locally first.

We should finish up this migration to fully benefit from from `object_store`.

- [ ] Use object_store everywhere, replacing the S3 client and custom code. At the end, we should depend only on `object_store` (not S3) and should support reading from any object store supported by the crate.
- [ ] Use byte-range reads to read Parquet files directly, rather than fetching them to local disk.

bjchambers:
Some of this may be done as part of building the new partitioned execution logic (as part of #409).

epinzur:
I believe this work is complete for the getMetadata() and prepareData() methods, but still needs to be completed on the query execution and materialization code paths.

bjchambers:
In general -- started working on this to allow operating on many and/or large files without filling up the disk. First PR(s) are ready for review.

@epinzur re `getMetadat()` and `prepareData()` -- it isn't really complete for them either. Specifically, they still rely on downloading the whole file. For get metadata, we should only need to fetch the bytes corresponding to the footer, for prepare we should be able to use object store to read bytes in chunks, never fetching the whole thing. Similarly, it isn't used for *uploading* the file yet.

bjchambers:
Capturing some links / thoughts:

* Example of getting the minimum/maximum time from the parquet metadata (the file stats): https://github.com/kaskada-ai/kaskada/blob/7858a62bc26c4ffd2451336d6d4dee82bd393fab/crates/sparrow-runtime/src/metadata/prepared_metadata.rs#L45
* Fetching the schema is likely done by https://github.com/kaskada-ai/kaskada/blob/main/crates/sparrow-runtime/src/metadata/raw_metadata.rs

PR #471, opened 2023-07-05
feat: Read directly during compute

bjchambers:
This is part of #465 -- specifically, reading directly using `object_store` during compute.


crates/sparrow-runtime/src/read/parquet_file.rs
```diff
@@ -0,0 +1,197 @@
...
+        let path = self.path.clone();
+        let metadata = self.metadata.clone();
+
+        let mut batch_stream = ParquetRecordBatchStreamBuilder::new(self)
+            .await
+            .into_report()
+            .change_context(Error::ReadingParquetFile)
+            .attach_printable_lazy(|| path.clone())?;
+
+        batch_stream = batch_stream.with_batch_size(100_000);
```
jordanrfrazier, line 89: Reuse existing batch size const, or make this its own cosnt?

crates/sparrow-runtime/src/execute/operation/scan.rs
```diff
@@ -523,11 +522,11 @@ mod tests {
         let metadata_parquet_file = write_parquet_file(&metadata, None);
 
         let prepared = PreparedFile {
-            path: parquet_file.to_string_lossy().to_string(),
+            path: format!("file://{}", parquet_file.display()),
```
jordanrfrazier, line 525: Why does this need to be prefixed everywhere now? And is there a better way you can think of so we won't forget to leave it in the future? 

With a lot of extra work we could make a `LocalPath` wrapper that enforces the `file://` prefix. Or just change variable names to be prefixed with `local` to give us a hint to use `file://`. 

Not a blocker on this PR obviously

jordanrfrazier (approved):

crates/sparrow-runtime/src/execute/operation/scan.rs
bjchambers: Previously: We expected these to be local paths and just read them relative the current working directly. But it's worth noting that they were always local paths because we downloaded the files first, and were getting the path after the download. Now, we are getting the path at which we stored the file in S3, so we expect this to be a URL more often. 

In general, it seems like it is better to just expect a URL at this point in the code. The issues are what local paths be relative to? The root directory? The current working directory? etc. I could be convinced that we should handle this at this point, but it seems like it may be better for the client and/or API to handle that).

PR #475, opened 2023-07-05
feat: Prepare directly to object stores

bjchambers:
This is part of #465 and serves as the first example of using `object_store` and the async Parquet writer to write directly to object stores.

### DIFF
--- a/crates/sparrow-runtime/src/metadata/prepared_metadata.rs
+++ b/crates/sparrow-runtime/src/metadata/prepared_metadata.rs
@@ -90,8 +90,9 @@ impl PreparedMetadata {
 
         let num_rows = data.num_rows() as i64;
         anyhow::ensure!(num_rows > 0, "Data should be non-empty");
-        let min_time = *time.values().iter().min().expect("non-empty");
-        let max_time = *time.values().iter().max().expect("non-empty");
+        // Time column is sorted (since the file is already prepared).
+        let min_time = *time.values().iter().next().expect("non-empty");
+        let max_time = *time.values().iter().last().expect("non-empty");
 
         Self::try_from_prepared_schema(
             data_path,
--- a/crates/sparrow-runtime/src/metadata/prepared_metadata.rs
+++ b/crates/sparrow-runtime/src/metadata/prepared_metadata.rs
@@ -54,7 +54,7 @@ impl PreparedMetadata {
         );
         anyhow::ensure!(
             prepared_schema.field(0).data_type() == &TimestampNanosecondType::DATA_TYPE,
-            "First column of prepared files must be '_time'"
+            "First column of prepared files must be TimestmapNanosecondType"
         );
 
         // Compute the time statistics directly from the data.
--- a/crates/sparrow-runtime/src/metadata/prepared_metadata.rs
+++ b/crates/sparrow-runtime/src/metadata/prepared_metadata.rs
@@ -1,20 +1,15 @@
-use std::path::Path;
 use std::sync::Arc;
 
 use arrow::array::TimestampNanosecondArray;
-use arrow::datatypes::{Schema, SchemaRef};
+use arrow::datatypes::{ArrowPrimitiveType, Schema, SchemaRef, TimestampNanosecondType};
 use arrow::record_batch::RecordBatch;
-use parquet::arrow::arrow_reader::ParquetRecordBatchReaderBuilder;
-use parquet::file::metadata::ParquetMetaData;
-use parquet::file::statistics::ValueStatistics;
+
 use sparrow_api::kaskada::v1alpha::PreparedFile;
 use sparrow_arrow::downcast::downcast_primitive_array;
 use sparrow_core::TableSchema;
 
-use crate::metadata::file_from_path;
-
 #[non_exhaustive]
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct PreparedMetadata {
     pub path: String,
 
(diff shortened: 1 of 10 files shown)


crates/sparrow-runtime/src/prepare.rs
```diff
@@ -140,192 +137,92 @@ impl<'a> fmt::Display for SourceDataWrapper<'a> {
...
+        .path()
+        .change_context_lazy(|| Error::Write(url.url().clone()))?;
+    let (upload_id, writer) = object_store
+        .put_multipart(&path)
+        .await
+        .into_report()
+        .change_context_lazy(|| Error::Write(url.url().clone()))?;
+    tracing::info!("Multipart upload to {url} started with ID {upload_id}");
+
+    let mut writer = parquet::arrow::AsyncArrowWriter::try_new(writer, batch.schema(), 1024, None)
```
kevinjnguyen, line 207: Is a `1024` an arbitrary buffer size or something we've confirmed is a good buffer size? https://arrow.apache.org/rust/parquet/arrow/async_writer/struct.AsyncArrowWriter.html#method.try_new

crates/sparrow-runtime/src/prepare.rs
```diff
@@ -140,192 +137,92 @@ impl<'a> fmt::Display for SourceDataWrapper<'a> {
...
-                metadata_object_store_url,
-                Path::new(&prepared_metadata.metadata_path),
-            ));
-
-            let result_prepared_file: PreparedFile = prepared_metadata
-                .to_owned()
-                .with_path(prepared_url)
-                .with_metadata_path(metadata_prepared_url)
+        // Create the prepared file via PreparedMetadata.
+        // TODO: We could probably do this directly, eliminating the PreparedMetadata struct.
```
kevinjnguyen, line 171: What is the work required to eliminate the prepared metadata struct? It's odd to go from PreparedMetadata back to the PreparedFile.

crates/sparrow-runtime/src/metadata/prepared_metadata.rs
```diff
@@ -67,6 +70,39 @@ fn get_time_statistics(
...
+        // Compute the time statistics directly from the data.
+        //
+        // TODO: We could instead just get this from the parquet metadata.
+        let time = data.column(0);
+        let time: &TimestampNanosecondArray = downcast_primitive_array(time.as_ref())?;
+
+        let num_rows = data.num_rows() as i64;
+        anyhow::ensure!(num_rows > 0, "Data should be non-empty");
+        let min_time = *time.values().iter().min().expect("non-empty");
+        let max_time = *time.values().iter().max().expect("non-empty");
```
kevinjnguyen, line 94: Is it cheap to calculate the min and max from the iterators? Is it reading from metadata or is it iterating through all the time values?

kevinjnguyen (approved):

crates/sparrow-runtime/src/prepare.rs
bjchambers: Arbitrary.

Based on:

> The intermediate buffer will automatically be resized if necessary.

It seems unlikely it will be a problem. But, we'd likely benefit from larger buffer sizes (although that would limit the number of concurrent uploads we could do). 

S3 seems to suggest it needs to be at least 5 MB (although that may be the "resized if necessary". We could start there? Thoughts on a good starting place?

crates/sparrow-runtime/src/prepare.rs
bjchambers: We were always doing that -- in fact, for a while we were returning both and dropping one of them, so it was just "hidden". I think it could be a few parts:

1. We could just add some methods to the `PreparedFile` (the proto equivalent) of `PreparedMetadata`, which AFAICT isn't widely used outside of "get the schema of this ParquetFile" which we can do better using the `ParquetFile` struct.
2. We could also go even further -- technically, we use the prepared file to store the min/max time and the num rows. That is available directly in the ParquetMetadata, which in turn we can get from the "footer" of the file. Currently, we need to fetch the footer to initialize the `ParquetFile` (and determine the schema), but we could put that footer in the DB, which would allow us to skip that step entirely.

I don't think this should be done in this PR though. Thoughts?

crates/sparrow-runtime/src/metadata/prepared_metadata.rs
bjchambers: It is iterating through the data -- but that should be cheaper than writing to Parquet to get the stats. Actually, since we've sorted we should be able to just use the first and last, which should be basically free.

And as noted in the other response -- we could do this from the Parquet footer stats more directly without having our own proto, which might be even simpler.

crates/sparrow-runtime/src/prepare.rs
bjchambers: Updated to a constant and set to 5 MB. We can adjust in future PRs after measuring.

crates/sparrow-runtime/src/metadata/prepared_metadata.rs
bjchambers: Updated to use first/last.

crates/sparrow-runtime/src/prepare.rs
kevinjnguyen: Agreed, I was just curious to see the effort. Its also not worth prioritizing given the limited usage of the prepared metadata. Looks good to me!

wren/compute/prepare_manager.go
```diff
@@ -267,7 +267,7 @@ func (m *prepareManager) executePrepare(ctx context.Context, owner *ent.Owner, p
 		prepareReq := &v1alpha.PrepareDataRequest{
 			SourceData:       sourceData,
 			Config:           computeTable.Config,
-			OutputPathPrefix: prepareOutputURI,
+			OutputPathPrefix: prepareOutputURI + "/",
```
bjchambers, line 270: @epinzur This was my hacky fix -- I tried adding it to the code that generates `prepareOutputURI` but that didn't work (looks like Go's path parser treats trailing slash as irrelevant or something silly -- maybe there is a better URI parser, but this works for now).

PR #479, opened 2023-07-06
feat: Use metadata for retrieving the schema

bjchambers:
This is part of #465.

This also simplifies the `ObjectStore` API to hide the `ObjectStoreKey`.

### DIFF
--- a/crates/sparrow-runtime/src/metadata/raw_metadata.rs
+++ b/crates/sparrow-runtime/src/metadata/raw_metadata.rs
@@ -127,6 +127,7 @@ impl RawMetadata {
     ///
     /// For CSV, this currently needs to download a local copy of the file, since
     /// Arrow does not (as of 2023-July-06) support reading CSV using `AsyncWrite`.
+    // TODO(https://github.com/kaskada-ai/kaskada/issues/486): Async CSV support.
     async fn try_from_csv(
         path: &str,
         object_stores: &ObjectStoreRegistry,
--- a/crates/sparrow-runtime/src/metadata/raw_metadata.rs
+++ b/crates/sparrow-runtime/src/metadata/raw_metadata.rs
@@ -5,7 +5,6 @@ use std::sync::Arc;
 use arrow::array::ArrowPrimitiveType;
 use arrow::datatypes::{DataType, Field, FieldRef, Schema, SchemaRef, TimestampMillisecondType};
 use error_stack::{IntoReport, IntoReportCompat, ResultExt};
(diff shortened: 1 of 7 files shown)


epinzur (approved):

crates/sparrow-runtime/src/metadata/raw_metadata.rs
```diff
@@ -106,75 +105,57 @@ impl RawMetadata {
...
-        }
+
+        Self::from_raw_schema(parquet_file.schema)
     }
 
-    /// Create a `RawMetadata` from a CSV string path and object store registry
+    /// Create a `RawMetadata` from a CSV string path and object store registry.
+    ///
+    /// For CSV, this currently needs to download a local copy of the file, since
+    /// Arrow does not (as of 2023-July-06) support reading CSV using `AsyncWrite`.
```
kevinjnguyen, line 129: Is there a feature request/issue number we could follow to see if they'll be supporting this in the near future?

crates/sparrow-runtime/src/stores/object_stores.rs
```diff
@@ -23,29 +21,27 @@ use super::{object_store_url::ObjectStoreKey, ObjectStoreUrl};
 /// For now, the registry exists as a cache for the clients due to the overhead
 /// required to create the cache. The future goal for the registry is to
 /// control the number of possibile open connections.
-#[derive(Default)]
+#[derive(Default, Debug)]
 pub struct ObjectStoreRegistry {
-    object_stores: RwLock<HashMap<ObjectStoreKey, Arc<dyn ObjectStore>>>,
+    object_stores: DashMap<ObjectStoreKey, Arc<dyn ObjectStore>>,
```
kevinjnguyen, line 26: What's the reasoning to switch from RwLock -> DashMap? Do we get performance/simplification?

kevinjnguyen (approved):

crates/sparrow-runtime/src/stores/object_stores.rs
bjchambers: Generally just simplification (as shown from all the deletions). There are performance benefits but I don't think they're compelling. Basically -- RwLock forces us to manage the lock, and lock the entire hashmap. DashMap does that internally and manages the lock under the hood.

crates/sparrow-runtime/src/metadata/raw_metadata.rs
bjchambers: Broke the CSV case out into https://github.com/kaskada-ai/kaskada/issues/486

PR #482, opened 2023-07-07
feat: support type inference for map<k,v>

jordanrfrazier:
Adds type inference for map types. Does not yet implement any functions for maps.

### DIFF
--- a/crates/sparrow-api/src/kaskada/v1alpha/schema_traits.rs
+++ b/crates/sparrow-api/src/kaskada/v1alpha/schema_traits.rs
@@ -15,6 +15,32 @@ impl DataType {
         }
     }
 
+    /// Creates a new map from the given fields.
+    ///
+    /// `fields` should have two elements, the first being the key type
+    /// and the second being the value type.
+    pub fn new_map(fields: Vec<schema::Field>) -> Self {
+        debug_assert!(fields.len() == 2);
+        Self {
+            kind: Some(data_type::Kind::Map(Box::new(data_type::Map {
+                key: Some(Box::new(
+                    fields[0]
+                        .data_type
+                        .as_ref()
+                        .expect("data type to exist")
+                        .clone(),
+                )),
+                value: Some(Box::new(
+                    fields[1]
+                        .data_type
+                        .as_ref()
+                        .expect("data type to exist")
+                        .clone(),
+                )),
+            }))),
+        }
+    }
+
     pub fn new_primitive(primitive: data_type::PrimitiveType) -> Self {
         Self {
             kind: Some(data_type::Kind::Primitive(primitive as i32)),
@@ -155,6 +181,37 @@ impl TryFrom<&arrow::datatypes::DataType> for DataType {
                     .try_collect()?;
                 Ok(DataType::new_struct(fields))
             }
+            // Note: the `ordered` field may let us specialize the implementation
+            // to use binary search in the future.
+            arrow::datatypes::DataType::Map(s, _) => {
+                // [DataType::Map] is represented as a list of structs with two fields: `key` and `value`
+                let arrow::datatypes::DataType::Struct(fields) = s.data_type() else {
+                    // unexpected - maps should always contain a struct
+                    return Err(ConversionError::new_unsupported(s.data_type().clone()))
+                };
+
+                debug_assert!(fields.len() == 2, "expect two fields in map");
+                let key = &fields[0];
+                let value = &fields[1];
+                let key = schema::Field {
+                    name: "key".to_owned(),
+                    data_type: Some(key.data_type().try_into().map_err(
+                        |err: ConversionError<arrow::datatypes::DataType>| {
+                            err.with_prepend_field("key".to_owned())
+                        },
+                    )?),
+                };
+                let value = schema::Field {
+                    name: "value".to_owned(),
+                    data_type: Some(value.data_type().try_into().map_err(
+                        |err: ConversionError<arrow::datatypes::DataType>| {
+                            err.with_prepend_field("value".to_owned())
+                        },
+                    )?),
+                };
+
+                Ok(DataType::new_map(vec![key, value]))
+            }
             unsupported => Err(ConversionError::new_unsupported(unsupported.clone())),
         }
     }
@@ -181,6 +238,7 @@ impl TryFrom<&FenlType> for DataType {
 
     fn try_from(value: &FenlType) -> Result<Self, Self::Error> {
         match value {
+            FenlType::Collection(_, _) => Err(ConversionError::new_unsupported(value.clone())),
             FenlType::Concrete(data_type) => {
                 data_type
                     .try_into()
@@ -275,6 +333,26 @@ impl TryFrom<&DataType> for arrow::datatypes::DataType {
                 let item_type = arrow::datatypes::Field::new("item", item_type, true);
                 Ok(arrow::datatypes::DataType::List(Arc::new(item_type)))
             }
+            Some(data_type::Kind::Map(map)) => match (map.key.as_ref(), map.value.as_ref()) {
+                (Some(key), Some(value)) => {
+                    let key = arrow::datatypes::DataType::try_from(key.as_ref())
+                        .map_err(|e| e.with_prepend_field("map key".to_owned()))?;
+                    let value = arrow::datatypes::DataType::try_from(value.as_ref())
+                        .map_err(|e| e.with_prepend_field("map value".to_owned()))?;
+
+                    let fields = arrow::datatypes::Fields::from(vec![
+                        arrow::datatypes::Field::new("key", key, false),
+                        arrow::datatypes::Field::new("value", value, false),
+                    ]);
+                    let s = arrow::datatypes::Field::new(
+                        "entries",
+                        arrow::datatypes::DataType::Struct(fields),
+                        false,
+                    );
+                    Ok(arrow::datatypes::DataType::Map(Arc::new(s), false))
+                }
+                _ => Err(ConversionError::new_unsupported(value.clone())),
+            },
             None | Some(data_type::Kind::Window(_)) => {
                 Err(ConversionError::new_unsupported(value.clone()))
             }
--- a/crates/sparrow-api/src/kaskada/v1alpha/schema_traits.rs
+++ b/crates/sparrow-api/src/kaskada/v1alpha/schema_traits.rs
@@ -188,7 +188,7 @@ impl TryFrom<&FenlType> for DataType {
                         e.map_data_type(FenlType::Concrete)
                     })
             }
-            FenlType::Generic(_) => Err(ConversionError::new_unsupported(value.clone())),
+            FenlType::TypeRef(_) => Err(ConversionError::new_unsupported(value.clone())),
             FenlType::Window => Ok(Self {
                 kind: Some(data_type::Kind::Window(())),
             }),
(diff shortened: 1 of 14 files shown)


crates/sparrow-compiler/src/types/inference.rs
```diff
@@ -64,16 +67,71 @@ pub(crate) fn instantiate(
...
         match (parameter_type.inner(), argument_type.inner()) {
-            (FenlType::Generic(type_var), _) => {
+            (FenlType::TypeRef(type_var), _) => {
                 types_for_variable
                     .entry(type_var.clone())
                     .or_default()
-                    .push(argument_type);
+                    .push(argument_type.clone());
             }
+            (FenlType::Collection(_, type_vars), arg_type) => match arg_type {
```
jordanrfrazier, line 83: Here's where I am getting the `type` for each type variable within the `FenlType::Collection` (in this case, just the `map` is supported). 

The `arg_type` is parsed as a `DataType::Struct(..)`, which makes sense, since the parquet schema is a `record`:

```
{
    "name" : "map",
    "type" : [ "null", {
      "type" : "record",
      "name" : "map",
      "fields" : [ {
        "name" : "f1",
        "type" : [ "null", "string" ],
        "default" : null
      }, {
        "name" : "f2",
        "type" : [ "null", "string" ],
        "default" : null
      } ]
    } ],
    "default" : null
  },
```

jordanrfrazier: But then it's a bit weird, in that I don't have the type for the `key` which theoretically should be able to be any non-complex type. It's just assumed it's a `string` since it's in a `struct` (#120)

crates/sparrow-compiler/src/types/inference.rs
```diff
@@ -296,18 +369,54 @@ fn solve_constraint(
...
 /// Instantiate the (possibly generic) `FenlType` using the computed
 /// solutions.
 fn instantiate_type(fenl_type: &FenlType, solutions: &HashMap<TypeVariable, FenlType>) -> FenlType {
     match fenl_type {
-        FenlType::Generic(type_var) => solutions
+        FenlType::TypeRef(type_var) => solutions
             .get(type_var)
             .cloned()
             .unwrap_or(FenlType::Concrete(DataType::Null)),
+        FenlType::Collection(c, type_vars) => match c {
```
jordanrfrazier, line 384: Then down here, actually instantiating the types for the collection - for each type variable in the collection, I get the concrete fenl type. But what's weird is that I am manually assigning the field name to be `key` and `value`, as per: 

```
    /// A Map is a logical nested type that is represented as
    ///
    /// `List<entries: Struct<key: K, value: V>>`
    ///
    /// The keys and values are each respectively contiguous.
    /// The key and value types are not constrained, but keys should be
    /// hashable and unique.
    /// Whether the keys are sorted can be set in the `bool` after the `Field`.
    ///
    /// In a field with Map type, the field has a child Struct field, which then
    /// has two children: key type and the second the value type. The names of the
    /// child fields may be respectively "entries", "key", and "value", but this is
    /// not enforced.
    Map(FieldRef, bool),
```

But my actual type (which was done above) is a `DataType::Struct(fields: {f1: .., f2: ..})`. (Where `f1` and `f2` are the map keys, from the input:
```
    m                          map name                 time
0  f1     {'f1': 'dog', 'f2': 'cat'}    A  1996-12-19T16:39:57

```

jordanrfrazier: Which gets me to the error I'm at currently:
```
│
╰─▶ No field named 'value' found in struct Concrete(Struct([Field { name: "f1", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "f2", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }]))
    ├╴at crates/sparrow-compiler/src/compile.rs:97:10
    ├╴backtrace (1)
    ╰╴span trace with 1 frames (1)
```

crates/sparrow-compiler/src/types/inference.rs
bjchambers: I think this part probably needs to be moved to a helper function and/or struct. Basically, we need to be able to recursively handle both sides -- the signature ("parameters") and the actual types being used in the inference ("arguments").

The following assumes a function -- we could just as easily have a `struct` that wraps the `types_for_variable`, which may make it *easier* since we need to do things like the "add this fenl type to the vector".

I'll also say that there are likely *much* easier ways of handling location here, but it would be a pretty painful refactoring. If you hit too much pain with locations, let's discuss.

```
fn unify_types(
    parameter: &FenlType,
    argument: &FenlType,
    types_for_variable: &mut HashMap<TypeVariable, Vec<Located<FenlType>>>,
) -> error_stack::Result<(), Error> {
  match (parameter, argument) {
    (FenlType::TypeRef(variable), lhs) => {
      // Add `lhs` to the types_for_variable[variable]`
    }
    (_, FenlType::TypeRef(_)) => {
      // error -- type variables shouldn't appear in the argumetns
    }
    (FenlType::Collection(p_name, p_types), FenlType::Collection(a_name, a_types)) => {
      error_stack::ensure!(a_name == p_name, /* error -- can't use incompatible collections*);
      error_stack::ensure!(p_types.len() == a_types.len(), Error::Internal);
      // Recursively apply to each of the arguments.
      for (p, a) in p_types.zip(a_types) {
        unify_types(p, a, types_for_variable)?;
      }
     ...
     // other cases:
     // Equal concrete types vs. unequal concrete types.
     //
     // Parameter is a collection, right-hand-side is concrete.
     // Likely a bit painful, but I *think* we could actually have a "as collection" type thing.
     // Basically, we could treat `FenlType::Concrete(DataType::List(field))` as
     // `FenlType::Collection("list", vec![field.data_type()])` which would let us
     // handle it using the same code as above. Maybe we need a helper or something.
     // May be a reason to introduce the struct.
  }
}
```

bjchambers: Let's look tomorrow -- I actually think it may have inferred the wrong type for the parquet columns as well.

crates/sparrow-main/tests/e2e/collection_tests.rs
```diff
@@ -0,0 +1,124 @@
...
+    let batches: Vec<_> = reader.try_collect()?;
+    let read_schema = batches.get(0).context("no batches read")?.schema();
+    let batch = arrow::compute::concat_batches(&read_schema, &batches)
+        .context("concatenate read batches")?;
+
+    Ok(batch)
+}
+
+async fn create_temp_parquet(json_input: &str) -> &str {
+    let batch = batch_from_json(json_input).unwrap();
```
bjchambers, line 83: How does this work? I would expect this call to need to pass data types.

crates/sparrow-compiler/src/types/inference.rs
jordanrfrazier: Does what exists now sufficiently handle the non-nested case though? 

The main question I have here is if the `arguments` type is correct and is inferred to be a `DataType::Struct`, then I don't have the `key` type, so unclear how we would support non-string types. I'll keep on trying to go from `batch -> parquet` and see if that column is interpreted differently.

crates/sparrow-compiler/src/functions/collection.rs
```diff
@@ -0,0 +1,9 @@
+use sparrow_plan::InstOp;
+
+use crate::functions::{Implementation, Registry};
+
+pub(super) fn register(registry: &mut Registry) {
+    registry
+        .register("get<K: key, V: any>(map: map<K, V>, key: K) -> V")
+        .with_implementation(Implementation::Instruction(InstOp::Get));
```
jordanrfrazier, line 8: I expect I'll rename `InstOp::Get` to `InstOp::GetFromMap` internally, to then support `GetFromList` as well, but can leave for now.

jordanrfrazier: Also - prefer `Input.key | get(Input.map)` or `Input.map | get(Input.key)`? Currently is former, but latter looks more correct with pipe notation

crates/sparrow-main/tests/e2e/collection_tests.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    // Create a Parquet writer
+    let mut writer = ArrowWriter::try_new(file, record_batch.schema(), None).unwrap();
+    writer.write(&record_batch).unwrap();
+
+    // Close the writer to finish writing the file
+    writer.close().unwrap();
+}
+
+async fn arrow_collection_data_fixture() -> DataFixture {
```
jordanrfrazier, line 106: I could move this to the `fixtures` class. It is currently creating a file each time, so I could technically just remove it, but it's been nice to be able to tweak the input data as I uncover more use cases I want to tests. So I'd prefer to leave it here for this PR, then remove in the next (or just move to `fixtures`) once I work on the implementation.

crates/sparrow-compiler/src/types/inference.rs
jordanrfrazier: Current status -- current impl handles non-nested case, which I want to say is sufficient for now. I'd like to get started on the implementation for the primitive types, then can come back to support nested types and complex-type implementation afterwards.

crates/sparrow-syntax/src/syntax/fenl_type.rs
```diff
@@ -19,7 +19,11 @@ pub enum FenlType {
     /// A specific Arrow DataType.
     Concrete(DataType),
     /// A generic type with the given type variable.
-    Generic(TypeVariable),
+    TypeRef(TypeVariable),
+    /// A collection type with the given type variable(s).
+    ///
+    /// e.g. (Collection::Map, [TypeVariable("K"), TypeVariable("V")])
+    Collection(Collection, Vec<TypeVariable>),
```
bjchambers, line 26: I think the arguments to the collection should be `FenlType`. Otherwise, we can't handle `Map<string, string>` (map from string to string), `Map<string, Map<i32>>` (nested maps), etc.

crates/sparrow-main/tests/e2e/collection_tests.rs
jordanrfrazier: https://github.com/kaskada-ai/kaskada/issues/493

crates/sparrow-syntax/src/syntax/fenl_type.rs
jordanrfrazier: https://github.com/kaskada-ai/kaskada/issues/494

crates/sparrow-api/src/kaskada/v1alpha/schema_traits.rs
```diff
@@ -181,8 +181,11 @@ impl TryFrom<&arrow::datatypes::DataType> for DataType {
                     .try_collect()?;
                 Ok(DataType::new_struct(fields))
             }
-            arrow::datatypes::DataType::Map(s, _) => match s.data_type() {
-                arrow::datatypes::DataType::Struct(fields) => {
+            // Note: the `ordered` field may let us specialize the implementation
+            // to use binary search in the future.
+            arrow::datatypes::DataType::Map(s, _) => {
+                // [DataType::Map] is represented as a list of structs with two fields: `key` and `value`
+                if let arrow::datatypes::DataType::Struct(fields) = s.data_type() {
```
bjchambers, line 188: Suggest alternating the order so the "error "is more clearly the "unhappy" path... but not a big deal.

```
let arrow::datatypes::DataType::Struct(fields) = s.data_type() else {
  return Err(ConversionError::new_unsupported(s.data_type().clone())
};

...
```

bjchambers (approved):

Issue #486, opened 2023-07-07
feat: Async Read/Write via object_store for CSV


bjchambers:
**Summary**

It looks like Arrow supports this -- [read](https://github.com/apache/arrow-rs/issues/78)/[write](https://github.com/apache/arrow-rs/issues/3740). For schema inference we will likely need something like "download all or the first 10 MB of the file and do schema inference" or something like that.

bjchambers:
(Broke CSV support out of #465)

Issue #466, opened 2023-06-30
feat: Split output into multiple files


bjchambers:
**Summary**
Rather than producing a single Parquet file containing the entire result set, we should split the results into files.

There are two reasons -- separate partitions should be able to write separate files and large results should be able to roll into multiple files, allowing the index columns to be written out and dropped, etc.

The API already supports this, but the Python client (and other places) likely don't have all the plumbing in place.

For an initial pass, it is likely OK to have the Python client download all files and combine them to a single data frame, but this can (and should) evolve over time to allow paging over the files (eg., fetch the first file and turn that into a data frame) and/or streaming support (fetch files as they are available), etc.

- [ ] Have the Parquet sink rotate files every N (~1,000,000 rows or so)
- [ ] Verify everything works when producing multiple files

bjchambers:
This is likely necessary to make maximal use of partitioned execution.

kevinjnguyen:
I wrote up a Python Client proposal design doc here: https://docs.google.com/document/d/1CHTiyLDD52FpwSI-SEhqft9HT1bYB-2WFTrCNrxqC0w/edit?usp=sharing

bjchambers:
My latest PR (#495) should write multiple files. @kevinjnguyen once that goes in, would you be able to verify everything is working with the python client support?

PR #492, opened 2023-07-10
feat: Use object_store to write files

bjchambers:
Also introduces paged file output.

This does a hacky CSV write by buffering a batch at a time.

This is related to #486.
This is part of #465.
This is part of #466.


crates/sparrow-runtime/src/execute/output/object_store.rs
```diff
@@ -1,209 +1,263 @@
...
+            .change_context(Error::Internal)?;
+        let path = url.path().change_context(Error::Internal)?;
+        let (_multipart_id, mut writer) = object_store
+            .put_multipart(&path)
+            .await
+            .into_report()
+            .change_context(Error::Upload)?;
+
+        match file_type {
+            FileType::Unspecified => todo!(),
```
jordanrfrazier, line 69: Error?

crates/sparrow-runtime/src/execute/output/object_store.rs
```diff
@@ -1,209 +1,263 @@
...
+                // Write an empty batch so headers are written.
+                let mut buffer = Vec::new();
+                let mut csv_writer = arrow::csv::WriterBuilder::new()
+                    .has_headers(true)
+                    .build(&mut buffer);
+                csv_writer
+                    .write(&RecordBatch::new_empty(schema))
+                    .into_report()
+                    .change_context(Error::Write)?;
+                std::mem::drop(csv_writer);
```
jordanrfrazier, line 91: you mentioned this in a comment recently - prefer the explicit drop, or can you place the `csv_writer` in its own `{ .. } ` block?

crates/sparrow-runtime/src/execute/output.rs
```diff
@@ -89,15 +87,17 @@ pub(super) fn write(
...
-                object_store::write(store, sink_schema, progress_updates_tx, batches)
-                    .change_context(Error::WritingToDestination {
-                        dest_name: "object_store".to_owned(),
-                    })
-                    .boxed(),
-            )
-        }
+        Destination::ObjectStore(store) => Ok(object_store::write(
+            context.object_stores.clone(),
+            store,
```
jordanrfrazier, line 92: nit: rename `store` to `destination`

crates/sparrow-runtime/src/execute/output/object_store.rs
```diff
@@ -1,209 +1,263 @@
...
+                csv_writer
+                    .write(&batch)
+                    .into_report()
+                    .change_context(Error::Write)?;
+                std::mem::drop(csv_writer);
+                writer
+                    .write_all(&buffer)
+                    .await
+                    .into_report()
+                    .change_context(Error::Write)?;
```
jordanrfrazier, line 135: What changed in this implementation -- why do we need to allocate a buffer for each batch? Due to the `writer` being a trait now?

crates/sparrow-runtime/src/execute/output/object_store.rs
```diff
@@ -1,209 +1,263 @@
...
+    );
+
+    // Currently, we pull batches and upload them asynchronously.
+    // We could increase concurrency by buffering some batches and performing
+    // multiple writes simultaneously.
+    while let Some(batch) = batches.next().await {
+        if batch.num_rows() == 0 {
+            // Don't create empty files. Wait for a non-empty batch before
+            // starting a new file.
+            continue;
```
jordanrfrazier, line 209: If we have no output, is the initial empty file with headers still created?

crates/sparrow-runtime/src/execute/output/object_store.rs
```diff
@@ -1,209 +1,263 @@
...
+
+impl WriterState {
+    async fn open(
+        object_store: &dyn ObjectStore,
+        output_prefix: &ObjectStoreUrl,
+        file_type: FileType,
+        schema: SchemaRef,
+    ) -> error_stack::Result<Self, Error> {
+        let url = output_prefix
+            .join(&format!("{}", Uuid::new_v4().as_hyphenated()))
```
jordanrfrazier, line 59: There's currently no "order" to the output files like this, is that acceptable for now we think?

jordanrfrazier (approved):

crates/sparrow-runtime/src/execute/output/object_store.rs
bjchambers: I think that should be OK. We could add more book-keeping to try to generate them in a sequence, but that would also change the format of the URL. Although you raise a good point that these file names should at least have a suffix (`.parquet` or `.csv`). Thoughts on naming patterns?

@epinzur

crates/sparrow-runtime/src/execute/output/object_store.rs
bjchambers: The issue here was that the `csv_writer` takes `&mut buffer`, so until it is dropped we can't get `&buffer` to write to the actual writer. I think I'm fine with it's own block hree to explicitly scope the writer. Thoughts?

crates/sparrow-runtime/src/execute/output/object_store.rs
epinzur: for query & materialize file output, we specify the path where all files should go.  see: https://buf.build/kaskada/kaskada/docs/main:kaskada.kaskada.v1alpha#kaskada.kaskada.v1alpha.ObjectStoreDestination

I think that inside that path, order would be super helpful.  

Alternatively, for query, the order could be somewhat determined from the streamed responses. 

For materialize, devs could check file create date, but some sequence would be better.

Files should have extensions.

I don't think URL format matters much.

crates/sparrow-runtime/src/execute/output/object_store.rs
bjchambers: Because we're using an `AsyncWrite` which requires we `await`. Unfortunately, the CSV writer only supports the synchronous `Write` trait. So we can either:

1. Run the future on the current thread, blocking until it completes.
2. Run synchronous code to write to a buffer, then use async code to write the buffer to the object store.

Given that CSV is likely to be used for smaller results, it seems like (2) is better, since it allows the thread to do other things while waiting (although, if we had more threads than cores, it may be OK because the thread would likely sleep while waiting). This seems to be the approach that Arrow suggests for CSV (see https://github.com/apache/arrow-rs/issues/3740).

crates/sparrow-runtime/src/execute/output/object_store.rs
bjchambers: Yes. I originally switched to having no files, but that causes problems for tests. It seems better to always ensure we have at least one output file, even if it has no rows. So, this is why (as noted in the comment above) we create the initial state as `Some(...)` instead of starting with `None`. In the `WriterState::open` we also write an empty batch so that the header is added to the CSV, etc.

bjchambers:
@epinzur Could you help out? I think this is failing in Wren:

1. It passes `... \"outputPrefixUri\":\"file:///data/output/mat_pulsarToObjStore\" ...`
2. It writes it to `file:///data/output/a722a354-467d-4fa8-b18e-398eac5edb4f-part-0.csv`

Note that the URI doesn't end with `/` which means that the URI join treats it as a file name, and `URI.join(filename)` produces `data/output/<UUID>-part-<part>.csv`.

epinzur:
> @epinzur Could you help out? I think this is failing in Wren:
> 
> 1. It passes `... \"outputPrefixUri\":\"file:///data/output/mat_pulsarToObjStore\" ...`
> 2. It writes it to `file:///data/output/a722a354-467d-4fa8-b18e-398eac5edb4f-part-0.csv`
> 
> Note that the URI doesn't end with `/` which means that the URI join treats it as a file name, and `URI.join(filename)` produces `data/output/<UUID>-part-<part>.csv`.

Yes, I can help

PR #495, opened 2023-07-11
feat: Use object_store to read prepare inputs

bjchambers:
For CSV, this uses object store to copy the file to local disk, since we need to infer the schema from the entire file, and the CSV inference doesn't support a streaming mode.

For Parquet, this uses streaming, async reads.

This is part of #465.

This also introduced `ObjectMetaExt` to make it possible to get a hash of the etag.


crates/sparrow-runtime/src/read/parquet_file.rs
```diff
@@ -5,22 +5,21 @@ use arrow::record_batch::RecordBatch;
...
 
 use crate::stores::{ObjectStoreRegistry, ObjectStoreUrl};
 
 /// A Parquet file in an [ObjectStore]. May be read asynchronously.
 #[derive(Clone)]
 pub struct ParquetFile {
     object_store: Arc<dyn ObjectStore>,
-    path: object_store::path::Path,
-    metadata: Arc<ParquetMetaData>,
+    pub object_meta: ObjectMeta,
```
bjchambers, line 21: This makes it possible to get the `ObjectMeta` from the parquet file handle. Since we already made the `head` request, it doesn't make sense to not retain it.

crates/sparrow-runtime/src/read/parquet_file.rs
```diff
@@ -145,55 +176,37 @@ impl parquet::arrow::async_reader::AsyncFileReader for ParquetFile {
...
-    parquet::file::footer::decode_metadata(metadata_bytes.as_ref())
-        .into_report()
-        .map(Arc::new)
-        .change_context(Error::InvalidParquetMetadata)
+    let fetch = |byte_range| {
+        object_store
+            .get_range(&object_meta.location, byte_range)
+            .map_err(|e| parquet::errors::ParquetError::External(Box::new(e)))
+    };
+    parquet::arrow::async_reader::fetch_parquet_metadata(
```
bjchambers, line 203: Simplify code for fetching (and implement the "prefetching") by using a new helper method from the parquet crate.

crates/sparrow-runtime/src/stores/object_meta_ext.rs
```diff
@@ -0,0 +1,38 @@
...
+    /// For object stores that don't provide an etag, this will return a hash of
+    /// the last modified time.
+    fn etag_hash(&self) -> u64;
+}
+
+impl ObjectMetaExt for ObjectMeta {
+    fn etag_hash(&self) -> u64 {
+        match &self.e_tag {
+            None => {
+                // The underling store doesn't provide an etag.
```
bjchambers, line 17: "underlying"

crates/sparrow-runtime/src/streams/pulsar/stream.rs
```diff
@@ -50,23 +50,6 @@ pub fn execution_stream(
     }
 }
 
-/// Creates a pulsar stream to be used during preparation.
-///
-/// This stream reads messages until it times out, which (generally) indicates that
-/// no more messages exist on the stream at that point in time.
-pub fn preparation_stream(
```
bjchambers, line 57: Deleted a variety of unused pulsar functionality.

crates/sparrow-main/tests/e2e/fixture/parquet_table_builder.rs
```diff
@@ -79,6 +80,11 @@ impl ParquetTableBuilder {
             parquet_writer.close().unwrap();
         }
 
+        // The file modification time is used as the prepare hash for local files.
+        // For determinism, we set this to a fixed value in tests.
+        filetime::set_file_mtime(temp_file.path(), PARQUET_FILE_TIME)
```
bjchambers, line 85: Since systems that don't provide etags rely on the "last modified time", updated this to generate parquet files with a specific modification time.

crates/sparrow-runtime/src/prepare.rs
```diff
@@ -46,30 +38,90 @@ const UPLOAD_BUFFER_SIZE_IN_BYTES: usize = 5_000_000;
...
-                reader_from_parquet(config, reader, file_size, prepare_hash, slice).await?
+                let url = ObjectStoreUrl::from_str(source)
+                    .change_context(Error::InvalidUrl(source.to_owned()))?;
+                let file = ParquetFile::try_new(object_stores, url, None)
+                    .await
+                    .change_context(Error::CreateReader)?;
+
+                let file_size = file.object_meta.size;
+                let num_rows = file.num_rows();
+                let num_files = get_num_files(file_size);
```
jordanrfrazier, line 60: this'll be the number of prepared files we produce from the source file, correct?

crates/sparrow-runtime/src/prepare.rs
```diff
@@ -46,30 +38,90 @@ const UPLOAD_BUFFER_SIZE_IN_BYTES: usize = 5_000_000;
...
 /// batches.
 pub async fn prepared_batches<'a>(
+    object_stores: &ObjectStoreRegistry,
     source_data: &SourceData,
     config: &'a TableConfig,
     slice: &'a Option<slice_plan::Slice>,
 ) -> error_stack::Result<BoxStream<'a, error_stack::Result<(RecordBatch, RecordBatch), Error>>, Error>
 {
-    let prepare_hash = get_prepare_hash(source_data)?;
+    // let prepare_hash = get_prepare_hash(source_data)?;
```
jordanrfrazier, line 47: commented

crates/sparrow-runtime/src/read/parquet_file.rs
```diff
@@ -36,7 +35,7 @@ pub enum Error {
     ReadingParquetFile,
 }
 
-const BATCH_SIZE_ROWS: usize = 100_000;
+const BATCH_SIZE_ROWS: usize = 4_096;
```
jordanrfrazier, line 38: Why this change?

crates/sparrow-runtime/src/read/parquet_file.rs
```diff
@@ -68,58 +77,78 @@ impl ParquetFile {
...
-        let mut batch_stream = ParquetRecordBatchStreamBuilder::new(self)
+        let mut batch_stream = ParquetRecordBatchStreamBuilder::new(reader)
             .await
             .into_report()
             .change_context(Error::ReadingParquetFile)
-            .attach_printable_lazy(|| path.clone())?;
+            .attach_printable_lazy(|| self.object_meta.location.clone())?;
 
-        batch_stream = batch_stream.with_batch_size(BATCH_SIZE_ROWS);
+        batch_stream = batch_stream.with_batch_size(batch_size.unwrap_or(BATCH_SIZE_ROWS));
```
jordanrfrazier, line 108: Can you explain how this works?

jordanrfrazier (approved):

crates/sparrow-main/src/serve/file_service.rs
```diff
@@ -166,14 +166,14 @@ mod tests {
...
         let file_service = FileServiceImpl::new(object_store_registry);
-        let path = path.canonicalize().unwrap().to_string_lossy().to_string();
         let result = file_service
             .get_metadata(tonic::Request::new(GetMetadataRequest {
                 source: Some(
                     sparrow_api::kaskada::v1alpha::get_metadata_request::Source::SourceData(
                         SourceData {
                             source: Some(source_data::Source::ParquetPath(format!(
-                                "file:///{path}"
+                                "file:///{}",
```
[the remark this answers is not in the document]
bjchambers: I think they should all be absolute at this point. Maybe if the path was absolute we should put `//` and let the path have the third? But generally, should always be absolute.

crates/sparrow-runtime/src/prepare.rs
bjchambers: This is the number of files we would like to prepare, yes.

crates/sparrow-runtime/src/read/parquet_file.rs
bjchambers: Read on arrow mailing list that batches of 4k or 8k rows were preferred.

crates/sparrow-runtime/src/read/parquet_file.rs
bjchambers: If the `batch_size` is `None` then the default value (`BATCH_SIZE_ROWS` is used). If it is `Some(size)` then it is `size`.

Issue #494, opened 2023-07-10
feat: support fenl types in `Collection`


jordanrfrazier:
`FenlType::Collection` currently only supports type variables, meaning we can't do `map<i32, V>` type signatures. We should refactor the `Collection` type to accept `FenlTypes` instead of just type vars, then clean up the parsing and inference code around that.

jordanrfrazier:
Map and list types are not displayed correctly in the schema
<img width="401" alt="Screen Shot 2023-07-28 at 11 11 34 AM" src="https://github.com/kaskada-ai/kaskada/assets/122494242/46135344-42c6-44de-89e2-a7472ed36289">

PR #500, opened 2023-07-12
feat: add `get` function for maps (for string -> primitive)

jordanrfrazier:
Adds the `get` function for maps for string and large_string to primitive types.

Further additions can support more types. See https://github.com/kaskada-ai/kaskada/issues/494 for task list

Testing:
* verified that the sample map data (with a largeutf8) produces results as expected.


crates/sparrow-expressions/src/evaluators/map/get.rs
jordanrfrazier, line 1: Added as a placeholder for the new evaluator impls

bjchambers: Should have one for each evaluator (and use different names). So this may be `get_string_primtive`, etc. Feel free to omit for now.

crates/sparrow-runtime/src/read/parquet_file.rs
```diff
@@ -107,7 +107,7 @@ impl ParquetFile {
 
         batch_stream = batch_stream.with_batch_size(batch_size.unwrap_or(BATCH_SIZE_ROWS));
         if let Some(projection) = projection {
-            let mask = ProjectionMask::leaves(
+            let mask = ProjectionMask::roots(
```
jordanrfrazier, line 110: `roots` pulls direct children, `leaves` unnests, so for `map` types we were getting the incorrect top-level schema.

crates/sparrow-instructions/src/evaluators/map/get.rs
```diff
@@ -0,0 +1,123 @@
...
+use owning_ref::ArcRef;
+use sparrow_arrow::downcast::{downcast_primitive_array, downcast_string_array};
+
+use sparrow_plan::ValueRef;
+
+use crate::{Evaluator, EvaluatorFactory, StaticInfo};
+
+/// Evaluator for `get` on maps.
+#[derive(Debug)]
+pub(in crate::evaluators) struct GetStringToPrimitiveEvaluator<T>
```
bjchambers, line 16: I don't think this should need to specialize on the value type, if it uses the `take` kernel to create the values array. That would allow this to specialize only on the key type. And we can use `eq_dyn` to not specialize on the key type either.

jordanrfrazier: Ah nice, will try that. Was going to add the vectorized comparison before making this non-draft, so will see if that works

jordanrfrazier: The trouble with the vectorized comparison is that `map.keys()` flattens all entries, resulting in an array of a different length than the `key_input`. As you mentioned, the new version of arrow that supports the new `datum` trait may be of help to clean this up. For now, I'd like to merge this less efficient implementation in though

crates/sparrow-compiler/src/ast_to_dfg.rs
```diff
@@ -691,6 +697,17 @@ fn cast_if_needed(
     }
 }
 
+fn map_types_are_equal(a: &FieldRef, b: &FieldRef) -> bool {
```
jordanrfrazier, line 700: This was to get around the issue where maps
```
map(name: "m1", field: ..)
map(name: "m2", field: ..)
```
were not equal because the `name` was different. However, since we infer the concrete datatype in `type_inference`, we create a concrete `DataType::Map`, meaning we have to assign default namings. This was causing the inequality in map types that should be equal. 

I wonder if there's some edge case I'm not thinking of where this matters, though. I don't think so, but have done little exploration of this.

crates/sparrow-compiler/src/types/inference.rs
```diff
@@ -295,6 +295,21 @@ fn solve_type_class(
...
+            // error[E0010]: Invalid argument type(s)
+            //   --> Query:1:10
+            //   |
+            // 1 | { value: get("test" , Input.readings )}
+            //   |          ^^^ ------   -------------- Type: large_string
+            //   |          |   |
+            //   |          |   Type: string
+            //   |          Invalid types for call to 'get'
+            //   |
+            //   = Expected 'key'
```
jordanrfrazier, line 312: This is weird -- unsure if there's a quick ux improvement for this. the current errors make it difficult since they have a formatted string inserting these types.

crates/sparrow-main/tests/e2e/collection_tests.rs
```diff
@@ -1,119 +1,49 @@
...
-            TableConfig::new_with_table_source("Input", &Uuid::new_v4(), "time", None, "key", ""),
-            &["parquet/data_with_map.parquet"],
-        )
-        .await
-        .unwrap()
+#[tokio::test]
+async fn test_supplant() {
+    insta::assert_snapshot!(QueryFixture::new("{ value: get(\"airTemperature\" as large_string, Input.readings) | when(is_valid($input)) } ").with_dump_dot("asdf").run_to_csv(&supplant_data_fixture().await).await.unwrap(), @r###"
+    _time,_subsort,_key_hash,_key,value
+    "###);
```
jordanrfrazier, line 48: TODO: Remove this

The test works, and produces results as expected. However, we likely don't support `LargeUtf8` in other evaluators, meaning we probably can't manipulate the `keys` in any way. Also, having to cast to `as large_string` is weird, but unclear why the literal string isn't being inferred as `large_string` at the moment.

crates/sparrow-main/tests/e2e/fixtures.rs
```diff
@@ -180,3 +180,39 @@ pub(crate) async fn timestamp_ns_data_fixture() -> DataFixture {
...
+    DataFixture::new()
+        .with_table_from_files(
+            TableConfig::new_with_table_source("Input", &Uuid::new_v4(), "time", None, "key", ""),
+            &[&"parquet/data_with_map.parquet"],
+        )
+        .await
+        .unwrap()
+}
+
+pub(crate) async fn supplant_data_fixture() -> DataFixture {
```
jordanrfrazier, line 203: TODO: Remove this.

crates/sparrow-runtime/src/read/parquet_stream.rs
```diff
@@ -70,7 +70,7 @@ pub(super) async fn new_parquet_stream(
         let batch = RecordBatch::try_new(projected_schema.clone(), raw_batch.columns().to_vec())
             .into_report()
             .change_context(Error::ReadingBatch)
-            .attach_printable_lazy(|| RecordBatchAttachment::new("raw_batch", &raw_batch))
+            // .attach_printable_lazy(|| RecordBatchAttachment::new("raw_batch", &raw_batch))
```
jordanrfrazier, line 73: the `json` write did not support printing out `maps`, and was failing. But, I suppose that's only an issue if the `try_new` fails anyway. Will uncomment

crates/sparrow-api/src/kaskada/v1alpha/plan_impl.rs
```diff
@@ -423,6 +425,7 @@ impl std::fmt::Display for super::Literal {
                 write!(f, "interval_months:{i}")
             }
             Some(literal::Literal::Utf8(ref str)) => write!(f, "\\\"{str}\\\""),
```
bjchambers, line 427: This `ref str` syntax should be unnecessary (it's mostly obsolute). I'd be curious to know why we need it.

crates/sparrow-arrow/src/scalar_value.rs
```diff
@@ -544,6 +554,15 @@ impl ScalarValue {
                     Ok(Self::Utf8(None))
                 }
             }
+            DataType::LargeUtf8 => {
+                if array.is_valid(row) {
+                    let array: &LargeStringArray = downcast_string_array(array)?;
```
bjchambers, line 559: In some of these where we've already done the test (we know it's LargeUtf8) we may be better off using arrows `as_string_array` and [`as_largestring_array`](https://docs.rs/arrow/latest/arrow/array/fn.as_largestring_array.html)

crates/sparrow-compiler/src/ast_to_dfg.rs
bjchambers: Add some more comments to that effect so we can reconstruct why we're doing this?

crates/sparrow-compiler/src/dfg.rs
```diff
@@ -108,6 +108,7 @@ impl Default for Dfg {
 impl Dfg {
     pub(super) fn add_literal(&mut self, literal: impl Into<ScalarValue>) -> anyhow::Result<Id> {
         let literal = literal.into();
+        // TODO: FRAZ - do I need to support large string literal here?
```
bjchambers, line 111: Maybe. We're going to potentially run into many problems with things (eg., what is the type of the literal `"hello"`?) This is the same as numbers (where we previously assumed it was `i64`).

crates/sparrow-compiler/src/types/inference.rs
bjchambers: I'm not sure I follow, and I'm not sure that's the error. I think this is actually saying that `large_string` isn't valid as a `key` (eg., the type class) since we can't hash large string.

crates/sparrow-expressions/src/evaluators.rs
```diff
@@ -88,6 +91,10 @@ impl<'a> StaticArg<'a> {
     pub fn struct_(&self) -> error_stack::Result<StructValue, Error> {
         StructValue::try_new(self.index, self.data_type)
     }
+
+    pub fn _map_(&self) -> error_stack::Result<MapValue, Error> {
```
bjchambers, line 95: I don't think we need the `_` suffix, since I don't believe that `map` is a keyword in rust. I believe we can just omit this part of the change if we want for now and add it later.

crates/sparrow-expressions/src/evaluators/map/get.rs
```diff
@@ -0,0 +1,38 @@
+use arrow_array::ArrayRef;
```
bjchambers, line 1: suggest omitting this and other changes to the expressions crate for this PR.

crates/sparrow-instructions/src/evaluators/macros.rs
```diff
@@ -297,7 +297,91 @@ macro_rules! create_typed_evaluator {
...
+/// fn create_evaluator(args: Vec<StaticArg>, result_type: &DataType) -> anyhow::Result<Self> {
+///   create_map_evaluator!(key_type, map_value_type, GetStringToPrimitiveEvaluator, info)
+/// }
+/// ```
+macro_rules! create_map_evaluator {
+    ($key_type:expr, $input_type:expr, $string_evaluator:ident, $large_string_evaluator:ident, $info:expr) => {{
+        use arrow::datatypes::*;
+        use create_primitive_evaluator;
+        match $key_type {
+            DataType::Utf8 => create_primitive_evaluator!($input_type, $string_evaluator, $info),
```
bjchambers, line 322: I'd expect all the other evaluators to be changed to include `LargeUtf8` if they support `Utf8`. Eg., we should be able to do string len, substring, etc. on it, right?

crates/sparrow-instructions/src/evaluators/map/get_large_string_to_primitive.rs
```diff
@@ -0,0 +1,124 @@
...
+where
+    T: ArrowPrimitiveType + Sync + Send,
+{
+    map: ValueRef,
+    key: ValueRef,
+    // Make the compiler happy by using the type parameter
+    _phantom: PhantomData<T>,
+}
+
+impl<T> EvaluatorFactory for GetLargeStringToPrimitiveEvaluator<T>
```
bjchambers, line 26: The only difference between these is the size of the length (`i32` vs. `i64`). Make that another type parameter and then we won't need the duplicate implementations?

crates/sparrow-main/tests/e2e/basic_error_tests.rs
```diff
@@ -1086,3 +1088,32 @@ async fn test_unsupported_output_type_with_nested_formulas_2() {
           - ""
     "###);
 }
+
+#[tokio::test]
+async fn test_swapped_args_for_get_map() {
```
bjchambers, line 1093: Suggest putting this in `map_tests`. I think we should get rid of the `error_tests` and just test errors alongside the valid cases (the fixtures more often line up, etc.).

crates/sparrow-main/tests/e2e/fixtures.rs
```diff
@@ -180,3 +180,39 @@ pub(crate) async fn timestamp_ns_data_fixture() -> DataFixture {
...
+/// Create a simple table with a collection type (map).
+///
+/// ```json
+/// {"time": 2000, "key": 1, "e0": {"f1": 0,  "f2": 22}, "e1": 1,  "e2": 2.7,  "e3": "f1" }
+/// {"time": 3000, "key": 1, "e0": {"f1": 1,  "f2": 10}, "e1": 2,  "e2": 3.8,  "e3": "f2" }
+/// {"time": 3000, "key": 1, "e0": {"f1": 5,  "f2": 3},  "e1": 42, "e2": 4.0,  "e3": "f3" }
+/// {"time": 3000, "key": 1, "e0": {"f2": 13},           "e1": 42, "e2": null, "e3": "f2" }
+/// {"time": 4000, "key": 1, "e0": {"f1": 15, "f3": 11}, "e1": 3,  "e2": 7,    "e3": "f3" }
+/// ```
+pub(crate) async fn collection_data_fixture() -> DataFixture {
```
bjchambers, line 193: Suggest having the fixture in the `collection_tests.rs` so it is easier to find / local to the tests.

crates/sparrow-runtime/src/read/parquet_stream.rs
bjchambers: Also next version should fix it: https://github.com/apache/arrow-rs/pull/4364/files

crates/sparrow-compiler/src/types/inference.rs
jordanrfrazier: we can hash large string though..I think it's something to do with the literal key input was a `string` and the `map.key` was a `large_string`.

bjchambers: Do you have logic for least upper bound? Specifically, the least upper bound of `string` and `large_string` should be `large_string`.

jordanrfrazier: Ah, that was part of the problem -- that and comparing the incorrect types for doing promotion

crates/sparrow-api/src/kaskada/v1alpha/plan_impl.rs
jordanrfrazier: Removed ref syntax

crates/sparrow-instructions/src/evaluators/macros.rs
jordanrfrazier: Agreed. Tracking in https://github.com/kaskada-ai/kaskada/issues/494

crates/sparrow-compiler/src/dfg.rs
jordanrfrazier: I'm not sure.. I think if we assume that every string literal is just a `string`, then we can rely on the `lub` of `string` and `large_string` to cast to `large_string` if necessary?

crates/sparrow-instructions/src/evaluators/map/get_large_string_to_primitive.rs
jordanrfrazier: Yes, parameterized and remove this

crates/sparrow-compiler/src/ast_to_dfg.rs
```diff
@@ -647,6 +647,16 @@ fn cast_if_needed(
         // TODO: This maybe should create an error node and return it.
         (_, FenlType::Error) => Ok(value),
         (actual, expected) if actual == expected => Ok(value),
+
+        // This handles the case where user `map` types have different names but are
+        // otherwise identical. When constructing the concrete map during inference,
+        // we make up names for the map types, so this is a common case. This ensures that
+        // the types themselves are still treated as equal.
+        (FenlType::Concrete(DataType::Map(s, _)), FenlType::Concrete(DataType::Map(s2, _)))
```
bjchambers, line 655: Why do map types have names at all? Is this they have different *field names* in the struct?

crates/sparrow-compiler/src/ast_to_dfg.rs
bjchambers: Can you put more of that comment on the method -- explain what the method does / why it exists (and then you just need a short comment on where it is used to explain why we need it there)?

bjchambers (approved):

crates/sparrow-main/tests/e2e/map_tests.rs
```diff
@@ -0,0 +1,7 @@
+//! Basic e2e tests for the map functions.
+
+use indoc::indoc;
+use sparrow_api::kaskada::v1alpha::TableConfig;
+use uuid::Uuid;
+
+use crate::{DataFixture, QueryFixture};
```
bjchambers, line 7: This file seems mostly empty / unused? Is there any reason it exists (most tests are in `collection_tests.rs`? Should they move here?

crates/sparrow-compiler/src/ast_to_dfg.rs
jordanrfrazier: Added

crates/sparrow-compiler/src/ast_to_dfg.rs
jordanrfrazier: Updated language to say _field_ names

crates/sparrow-main/tests/e2e/map_tests.rs
jordanrfrazier: oops - removed file

PR #501, opened 2023-07-12
ref: Cleanup object_store code a bit

bjchambers:
Move `upload` and `download` to the root `ObjectStoreRegistry`.

Move `object_store_key` to a separate file and make that an implementation detail.

This is part of cleanup after most of #465.

### DIFF
--- a/crates/sparrow-runtime/src/prepare/error.rs
+++ b/crates/sparrow-runtime/src/prepare/error.rs
@@ -1,15 +1,7 @@
-use std::path::PathBuf;
-
 use url::Url;
-
-use crate::stores::ObjectStoreUrl;
 
 #[derive(derive_more::Display, Debug)]
 pub enum Error {
-    #[display(fmt = "unable to open '{path:?}'")]
-    OpenFile { path: PathBuf },
-    #[display(fmt = "unable to create '{path:?}'")]
-    CreateFile { path: PathBuf },
     #[display(fmt = "internal error")]
     Internal,
     #[display(fmt = "failed to create Parquet file reader")]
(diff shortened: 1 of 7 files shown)


crates/sparrow-runtime/src/prepare/error.rs
```diff
@@ -1,15 +1,7 @@
-use std::path::PathBuf;
-
 use url::Url;
 
-use crate::stores::ObjectStoreUrl;
-
 #[derive(derive_more::Display, Debug)]
 pub enum Error {
-    #[display(fmt = "unable to open '{path:?}'")]
-    OpenFile { path: PathBuf },
```
bjchambers, line 10: Since the error enums are public, we don't get a warning / lint to remove errors that are no longer used. That kind of sucks. On the other hand, since they are potentially part of the public API, removing them is actually a potential problem.

I suspect we may want a pattern where we have top-level, public error enums with more general cases ("InvalidArguments", "Internal", etc.) and then we can have internal (`pub crate`) error enums with more details.

That way the `pub enum` is the "public" contract, and we've set ourselves up for not needing to change that as often.

crates/sparrow-runtime/src/stores/registry.rs
```diff
@@ -47,32 +47,65 @@ impl ObjectStoreRegistry {
...
+        source_path: &path::Path,
+        destination_url: ObjectStoreUrl,
     ) -> error_stack::Result<(), Error> {
-        let target_path = target_url.path()?;
-        let object_store = self.object_store(&target_url)?;
-        let mut local_file = fs::File::open(local_file_path)
+        let target_path = destination_url.path()?;
+        let object_store = self.object_store(&destination_url)?;
+
+        let upload_error = || Error::UploadingObject {
```
bjchambers, line 56: @jordanrfrazier I've found this pattern useful. When I have a lot of errors in the same method that need the same error, create a helper closure like this that creates the error, then you can do `change_context_lazy(error_helper_name)`.

therapon (approved):

PR #503, opened 2023-07-12
feat: Use object store for rocksdb and debugging

bjchambers:
This uses `object_store` for uploading/downloading snapshots and the
query plan and flight record (debugging) information.

This removes the `s3` module, the `S3Helper` and the direct dependencies
on AWS.

This is part of #465.


crates/sparrow-runtime/src/execute/checkpoints.rs
```diff
@@ -0,0 +1,258 @@
...
+            };
+            // Drop the leading `/` (will be left over from the strip prefix above)
+            let relative_path = &relative_path[1..];
+            let source_url = resume_from.join(relative_path).change_context(Error::DownloadIo)?;
+            let destination_path = storage_path.join(relative_path);
+            tracing::info!("For object {object:?}, relative path is {relative_path}, downloading {source_url} to {}", destination_path.display());
+            Ok((source_url, destination_path))
+        })
+        .map_ok(|(source_url, destination_path)| async move {
+            // DO NOT SUBMIT: Can we fix this?
```
jordanrfrazier, line 92: Is this still a do not submit?

jordanrfrazier (approved):

crates/sparrow-runtime/src/execute/checkpoints.rs
bjchambers: Good question. I don't think it needs to be -- it was more a reminder to have this conversation. Specifically, it seems a bit hacky since we need to go back and forth between the `ObjectStoreUrl` (our wrapper of a URL supporting join and such) and the `object_store::path::Path` (which is what it expects and returns).

We *could* add more verbosity / remove some of the safety provided by our wrapper and use the path everywhere. We could also consider wrapping things more deeply... but neither seem *great*.

Looking at this code does it seem like it needs to be fixed now (given we'd need to do one of the above options) or does it seem ok?

PR #508, opened 2023-07-13
beta release flow using separate files

therapon:
* beta release flow for engine and python 
* using separate files to start with. If this works I can try `if` flows in one file again. I tried it locally and it got hairy

Wanted to commit this and try the flow before adding docs on the process.


.github/workflows/release_engine_beta.yml
```diff
@@ -0,0 +1,321 @@
...
+
+  # Allow the workflow to be manually run.
+  # When doing so, artifacts will be built but not associated with a release.
+  workflow_dispatch:
+
+# In cases of concurrent workflows running (consecutive pushes to PR)
+# leave the latest workflow and cancel the other (older) workflows
+# See https://docs.github.com/en/actions/using-jobs/using-concurrency
+concurrency:
+  group: engine-release-beta
```
bjchambers, line 14: It's hard to tell what is actually different here. Shouldn't the beta release be the same as the regular release, just with a different label?

therapon: Almost. A couple of issues 

1. Release takes over an already open draft. I did not want to alter that flow. If I use the release-drafter for Beta it takes over the draft. I would have to create 2 drafts one for beta and one for release 
2. Most changes are in the publishing of the binaries. Beta marks GitHub releases as pre-release and does not mark anything as latest. 

I would like to make it part of one file and reuse as much as possible. If this workflow works (I need to commit once and then try it out) I am planning to unify the files. (see comment in the PR)

therapon:
Closing this in favour of a new PR

PR #516, opened 2023-07-15
engine beta release fixes

therapon:
* create [beta release on GitHub](https://github.com/kaskada-ai/kaskada/releases/tag/engine%40v0.9.0-beta.1)
* pushed docker images for engine and jupyter for both GitHub and DockerHub 
    * https://github.com/kaskada-ai/kaskada/pkgs/container/kaskada%2Fengine
    * https://github.com/kaskada-ai/kaskada/pkgs/container/kaskada%2Fjupyter
    * https://hub.docker.com/layers/kaskadaio/engine/v0.9.0-beta.1/images/sha256-b5496497e718f557eade6a15e24435005cc56066a22146c8767179b2719809c6?context=explore
    * https://hub.docker.com/layers/kaskadaio/jupyter/v0.9.0-beta.1/images/sha256-8ccca38221d1b0e3a6dadac7ec785c7a5db0f0aa82a00c7b0bf29d71a776a43f?context=explore

Will follow up with a small PR on documentation.


.github/workflows/release_engine.yml
```diff
@@ -72,11 +72,16 @@ jobs:
           echo Release Tag: ${{ steps.draft_release.outputs.tag_name }}
           echo Release HTML URL: ${{ steps.draft_release.outputs.html_url }}
       
-      - name: Diagnostics for Beta 
+      # For beta releases we need to craete the beta release first so that
```
bjchambers, line 75: s/craete/create/g

bjchambers (approved):

.github/workflows/release_engine.yml
```diff
@@ -72,11 +72,16 @@ jobs:
           echo Release Tag: ${{ steps.draft_release.outputs.tag_name }}
           echo Release HTML URL: ${{ steps.draft_release.outputs.html_url }}
       
-      - name: Diagnostics for Beta 
+      # For beta releases we need to craete the beta release first so that 
+      # jobs and steps that follow can upload artifacts.  
+      - name: Create Release for Beta
```
bjchambers, line 77: We should create it with the same description as the drafter generates, ideally? Also, we should make sure that creating a beta release doesn't cause the actual release to draft based only on the changes after the beta.

.github/workflows/release_engine.yml
therapon, line 75:
```suggestion
      # For beta releases we need to create the beta release first so that 
```

.github/workflows/release_engine.yml
therapon: I do have to test that. But, the draft is already created (even now for version 0.9.0) so that will at least have the PRs up to the beta release creation. 

I am not sure if adding the same description will cause more issues?

PR #517, opened 2023-07-15
docs(beta): instructions for beta releases

therapon:


bjchambers (approved):
Looks good. Once stable would be good to have someone else run through the flow to make sure it works.

Issue #520, opened 2023-07-17
bug: slicing returns no results


kevinjnguyen:
**Description**
Slicing ([documented here](https://kaskada.io/docs-site/kaskada/main/developing/slices.html)) does not produce the expected results for either `EntityPercentFilter` or `EntityFilter`. Both appear to return no results.

**To Reproduce**
1. Create a table called transactions
```python
import kaskada.table
kaskada.table.create_table('transactions', 'time', 'name')
```
2. Load the provided CSV file: [dataset1.csv](https://github.com/kaskada-ai/kaskada/files/12069729/dataset1.csv)
```python
kaskada.table.load('transactions', 'dataset1.csv')
```
3. Create a slice and set the default slice
```python
from kaskada.slice_filters import EntityFilter
import kaskada.client

entity_keys = ["m0", "m10"]
entity_filter = EntityFilter(entity_keys)
kaskada.client.set_default_slice(entity_filter)
```

4. Perform a query  
```python
%%fenl
transactions
``` 

**Actual Behavior**
The result of the query is an empty dataframe.

**Expected Behavior**
A filtered dataset with only results for `m0` and `m10` (about 100) records.

**Additional context**
* Updated the documentation ([this PR](https://github.com/kaskada-ai/kaskada/pull/519))
* Same results for numeric and string entity keys

**Relevant Logs / Links**
Relevant stack traces, results, links, or screenshots of the issue.

Manager Logs
```
2023-07-17T13:19:53.258539Z  INFO started call grpc.code=OK grpc.component=server grpc.method=CreateQuery grpc.method_type=server_stream grpc.service=kaskada.kaskada.v1alpha.QueryService grpc.start_time=2023-07-17T08:19:53-05:00 grpc.time_ms=0.075 peer.address=[::1]:62853 protocol=grpc request_id=2d8c95eb4fd7b1f738855fb3d17361f4
[...]

PR #521, opened 2023-07-17
Entity Key Filter Integration Test (#520)

kevinjnguyen:
Adds a test for the entity key filter


epinzur:
test looks good, now need to fix the bug

clients/python/src/kaskada/query.py
```diff
@@ -11,7 +11,7 @@
 import kaskada.formatters
 import kaskada.kaskada.v1alpha.destinations_pb2 as destinations_pb
 import kaskada.kaskada.v1alpha.query_service_pb2 as query_pb
-from kaskada.client import KASKADA_DEFAULT_SLICE, Client, get_client
+from kaskada.client import Client, get_client
```
kevinjnguyen, line 14: This was the big issue for the slice configuration.  This line was changed previously in a [refactor](https://github.com/kaskada-ai/kaskada/commit/18048724a03d7c633e1b132c42a0b5764ea204d4) that altered the logic. We now have a test in place to prevent this from happening.

Although the change is subtle: `from kaskada.client import KASKADA_DEFAULT_SLICE, Client, get_client` is evaluated the moment this module is imported and never again. 

`slice_filter = kaskada.client.KASKADA_DEFAULT_SLICE` means now we will always reach out the `kaskada.client` module and ask what is the latest value.

clients/python/tests/test_query.py
```diff
@@ -48,6 +49,56 @@ def test_create_query_with_defaults(mockClient):
     )
 
 
+@patch("kaskada.client.Client")
+def test_query_uses_client_global_slice_filter(mockClient):
```
kevinjnguyen, line 53: Test to prevent the regression in the future.

crates/sparrow-runtime/src/prepare.rs
```diff
@@ -409,4 +413,117 @@ mod tests {
         let _prepared_schema = prepared_batch.schema();
         let _metadata_schema = metadata.schema();
     }
+
+    #[tokio::test]
+    async fn test_preparation_single_entity_key_slicing() {
```
kevinjnguyen, line 418: Some debugging involved looking at different stacks. Missed some tests in preparation so added some new ones.

clients/python/src/kaskada/query.py
```diff
@@ -134,7 +134,7 @@ def create_query(
         query_pb.CreateQueryResponse
     """
     if slice_filter is None:
-        slice_filter = KASKADA_DEFAULT_SLICE
+        slice_filter = kaskada.client.KASKADA_DEFAULT_SLICE
```
kerinin, line 137: Given the subtlety, it's probably worth adding a comment here explaining it...

kerinin (approved):

PR #528, opened 2023-07-17
feat: Initial partitioned execution

bjchambers:
This introduces the key components of partitioned execution.

- `sparrow-scheduler` provides functionality for managing the separate pipelines within the query plan and morsel-driven parallelism. It managing a thread-pool of workers pinned to specific CPUs pulling tasks from local queues.
- `sparrow-transforms` will provide implementations of the "transforms" (project, select, etc.) and a pipeline for executing the transforms.
- `sparrow-execution` will pull everything together to provide partitioned execution.

This is part of #409.


crates/sparrow-scheduler/src/pipeline.rs
```diff
@@ -0,0 +1,109 @@
...
+///
+/// An async input pipeline, which has no inputs, and wakes the output partition
+/// whenever new data is available.
+///
+/// A JIT compiled sequence of synchronous operators, that perform multiple operations
+/// from the physical plan as a single [`Pipeline`]. Parallelized implementations
+/// are also possible.
+///
+pub trait Pipeline: Send + Sync + std::fmt::Debug {
+    type Context: error_stack::Context;
```
bjchambers, line 54: To discuss:

Should we have a single `sparrow-scheduler::Error` or `sparrow-scheduler::PipelineError` that we require *all* pipelines produce? Currently we don't, which creates some pain. Specifically, the `Pipeline` trait is not "object safe", meaning we need to wrap it into a *different* trait before we put it behind `dyn`. See some of the complexity / cruft in `task.rs`.

crates/sparrow-scheduler/src/schedule_count.rs
```diff
@@ -0,0 +1,81 @@
+#[cfg(loom)]
+pub(crate) use loom::sync::atomic::{AtomicUsize, Ordering};
+
+#[cfg(not(loom))]
+pub(crate) use std::sync::atomic::{AtomicUsize, Ordering};
+
+#[repr(transparent)]
+#[derive(Debug, Default)]
+pub(crate) struct ScheduleCount(AtomicUsize);
```
bjchambers, line 9: This struct provides a way of ensuring we only schedule a task once even if we've queued multiple inputs for it. Needs some documentation. It's intended for use in `Task`, but splitting it out like this seems to make it easier to test.

crates/sparrow-scheduler/src/queue.rs
```diff
@@ -0,0 +1,148 @@
+//! Provide the global and per-thread (local) work queues.
```
bjchambers, line 1: These are relatively simple wrappers around the existing `work_queue` crate. It provides:

1. A fixed set of interfaces / documentation
2. Some tests around ordering

The idea is that by *not* working with the crate directly we should be more easily able to use other queues if we discover the need.

crates/sparrow-scheduler/src/queue.rs
```diff
@@ -0,0 +1,148 @@
...
+    fn push(&mut self, item: T);
+
+    /// Push an item to the *end* of the local queue, skipping the LIFO slot.
+    ///
+    /// This can be used to give other tasks a chance to run. Otherwise, there’s
+    /// a risk that one task will completely take over a thread in a push-pop
+    /// cycle due to the LIFO slot.
+    ///
+    /// When called on a global queue, items will be pushed to that.
+    fn push_yield(&mut self, item: T);
```
bjchambers, line 24: This trait was introduced so that we can run the logic of a pipeline from a worker (by passing the current worker queue) or from outside (by passing the global injector). This allows us to read from a Parquet file and push directly into the first transform, scheduling work with the global injector.

crates/sparrow-scheduler/src/worker.rs
```diff
@@ -0,0 +1,74 @@
...
+
+/// An individual worker that allows adding work to the local or global queue.
+pub struct Worker {
+    queue: LocalQueue<TaskRef>,
+}
+
+impl Worker {
+    /// Run the work loop to completion.
+    pub(crate) fn work_loop(mut self) -> error_stack::Result<(), Error> {
+        while let Some(task) = self.queue.pop() {
```
bjchambers, line 44: I think this needs some added logic. Specifically, we should probably sleep if there are no tasks and wait for new work to be injected rather than exiting the work loop.

crates/sparrow-scheduler/src/pending.rs
```diff
@@ -0,0 +1,55 @@
+use std::sync::atomic::{AtomicUsize, Ordering};
```
bjchambers, line 1: This file is unused. I had it in the earlier prototype to track the set of "pending pipelines and partitions". This may actually be useful for us to keep -- for batch pipelines, we can mark a pipeline as "done" when it's input completes, but for streaming pipelines we could keep them open. This could start to lead towards a better "true" streaming:

1. If your pipeline is run "streaming" then (a) it shouldn't have any "final" ticks, because those don't make sense and (b) we never close the corresponding inputs.
2. If your pipeline is run "batch" then (a) it may have "final" ticks and (b) we close the inputs as the pipeline completes.

crates/sparrow-transforms/src/transform_pipeline.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+}
+
+#[derive(Debug, Default)]
+struct InputBuffer {
+    buffer: VecDeque<Batch>,
+    is_closed: bool,
+}
+
+impl TransformPipeline {
```
bjchambers, line 46: This shows an example of creating a pipeline. This one runs a sequence of 1 or more transforms on an input batch. Note that it does some "low-ish" level mutex protections. This is partly intentional. As described in `pipeline.rs` it allows us to implement a variety of workflows by changing how the pipeline chooses to be implemented:

1. It could do more work in `push` which would block the producer.
2. It can do more in `do_work`

In general, `push` and `close` do a little bit of work to figure out what to wake up and push the work there. They are (ideally) short-lived, and don't hold the partition lock for long.

Do work shouldn't hold the lock while doing the bulk of the work.

crates/sparrow-transforms/src/transforms/project.rs
```diff
@@ -0,0 +1,51 @@
+use arrow_array::RecordBatch;
+use arrow_schema::SchemaRef;
+use error_stack::{IntoReport, ResultExt};
+use sparrow_arrow::Batch;
+
+use crate::transforms::Transform;
+use crate::Error;
+use sparrow_expressions::ExpressionExecutor;
+
+/// Transform for projection.
```
bjchambers, line 10: Most things should look more this -- a simple transform with no actual synchronization logic. Basically, we should use "pipelines" when we need to interact with how things are scheduled (merge, repartition, transform) and implement these kind of simpler "transforms" for the bulk of "batch in -> do a thing -> batch out".

crates/sparrow-scheduler/src/pipeline.rs
```diff
@@ -0,0 +1,109 @@
+use sparrow_arrow::Batch;
+
+use crate::queue::Queue;
+use crate::{Partition, TaskRef, Worker};
+
+/// A push-based interface used by the scheduler to drive query execution
```
therapon, line 6: I am not sure I am getting the relationships between Pipeline, partition, partition index, children and batch. 

Maybe we should have a comment that show an example or at least goes into how these things hook up to each other.

crates/sparrow-scheduler/src/queue.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    /// Push an item onto the global queue.
+    fn push_global(&self, item: T);
```
therapon, line 27: What is the difference between this function and the one preceding it?

The last comment on the preceding function spells out how it would work for a global queue.

crates/sparrow-scheduler/src/pipeline.rs
bjchambers: I think if we like that, it is pretty easy to add and cleans up code significantly, so I'll consider this something to do in this PR.

crates/sparrow-scheduler/src/pipeline.rs
bjchambers: Yeah, I'll probably try to add one to the top level. But to develop it, I'll try to answer here and then maybe you, @jordanrfrazier and I can discuss?

1. The entire query plan consists of a bunch of "transforms" (lookup request, lookup response, project, select, etc.)
2. We divide those "transforms" up into "pipeline plans" that are a single input -> sequence of transforms -> single output.
3. For execution, we create a bunch of "pipelines" that manage the data for a specific part of execution. This includes "transform pipelines" that run transforms, but also "repartition", "scan", etc. (eg., everything that executes is a pipeline, and it may be a superset of the "pipeline plans").
4. Each pipeline manages the state for *all* partitions that it may execute on. This is intentional, since it allows things like the "repartition" pipeline to accept data from any producing / input partition and store it in queues for each of the downstream partitions.
5. The above means that a pipeline has "input partitions" (the partitioning of the producer) and "processing / output partitions" (the partitioning of the actual pipeline and the output it produces).
6. A pipeline has 0 or more inputs (eg., a merge operation has two inputs, many things have 1 input, etc.). Thus, inputs are associated with the "input partition" (the partition that produced the input) and the "input index" (which of the inputs to the pipeline it is being sent to.
7. Children I don't think matters (it's basically how pipelines are connected and/or which transforms are connected together within a pipeline).
8. Batches are data that flows between pipelines.

crates/sparrow-scheduler/src/queue.rs
bjchambers: When called on a local queue, `push_global` will push to the global queue, while `push_yield` will push to the end of the local queue.

crates/sparrow-transforms/src/transform_pipeline.rs
```diff
@@ -0,0 +1,174 @@
...
+                &self.transforms.iter().map(|t| t.name()).format(","),
+            )
+            .finish()
+    }
+}
+
+#[derive(Debug, Default)]
+struct InputBuffer {
+    buffer: VecDeque<Batch>,
+    is_closed: bool,
```
jordanrfrazier, line 43: what situation is this used for?

crates/sparrow-transforms/src/transform_pipeline.rs
```diff
@@ -0,0 +1,174 @@
...
+    type Context = Error;
+
+    fn push(
+        &self,
+        partition: Partition,
+        input: usize,
+        batch: Batch,
+        queue: &mut dyn Queue<TaskRef>,
+    ) -> error_stack::Result<(), Error> {
+        error_stack::ensure!(input == 0, Error::InvalidInput { input });
```
jordanrfrazier, line 92: What's this logic?

crates/sparrow-scheduler/src/task.rs
```diff
@@ -0,0 +1,140 @@
...
+            partition: self.partition,
+        }
+    }
+}
+
+/// Trait with methods on generic `Task`.
+///
+/// This is an object safe wrapper allowing execution of the pipeline within the task.
+pub trait TaskObject: std::fmt::Debug + Sync + Send {
+    /// Record a request for scheduling.
```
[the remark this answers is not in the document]
jordanrfrazier: Does it return `false` if it was already scheduled -- are there situations when it is not schedulable?

crates/sparrow-scheduler/src/worker.rs
jordanrfrazier: What is in the "work loop" Does a worker handle work for one pipeline? Or when _should_ it exit?

crates/sparrow-transforms/src/transform_pipeline.rs
```diff
@@ -0,0 +1,174 @@
...
+        Ok(())
+    }
+
+    fn close(
+        &self,
+        partition: Partition,
+        input: usize,
+        queue: &mut dyn Queue<TaskRef>,
+    ) -> error_stack::Result<(), Error> {
+        error_stack::ensure!(input == 1, Error::InvalidInput { input });
```
jordanrfrazier, line 111: Does the `input` imply something about the state of the pipeline?

crates/sparrow-transforms/src/transforms/project.rs
```diff
@@ -0,0 +1,51 @@
...
+    fn apply(&self, batch: Batch) -> error_stack::Result<Batch, Error> {
+        assert!(!batch.is_empty());
+
+        let error = || Error::ExecuteTransform("project");
+        let columns = self.evaluators.execute(&batch).change_context_lazy(error)?;
+        let columns = self
+            .outputs
+            .iter()
+            .map(|index| columns[*index].clone())
+            .collect();
```
jordanrfrazier, line 44: Does this imply that a the evaluators may produce columns that the transform doesn't need, and just drops?

crates/sparrow-scheduler/src/pipeline.rs
bjchambers: 1. Not sure yet. I would expect that there is a step in the physical plan for applying aggregations, and then the individual aggregations are expressions within that step. Aggregations are somewhat TBD, but my thinking is that we'd have a step that says "do aggregations with window W" (so the cardinality is determined by the step) and then all the aggregations with that windowing are part of that step. Transforms and steps and pipelines probably merit some discussion.
2. Roughly. Not all pipelines at execution time correspond to "pipeline plans" in the proto.
3. `Scan` is a pipeline. Transform is just a class for "common" steps that behave a certain way, so we don't need to write them as separate pipelines, and transforms allow chaining a few them into a single pipeline. But special things (input, output, repartition, etc.) are pipelines on their own.
4. A bit of both -- the `with_key` function is the logic of changing the `key` and then it applies the `reparatition` pipeline to actually change partitioning and move data around.
5.  A pipeline can have different input/output partitions if needed, but I don't know that we'll have pipelines with that property often.
6. I think `input partition` is the partitioning from the producer. An interesting question is if we have multiple inputs do we require all of them to have the same partitioning, or do we allow it to be different. I think this could also be per-pipeline. Eg., I could see a case of the `merge` pipeline *not* having that requirement, and mixing some "repartition" logic into the merge operation.

crates/sparrow-scheduler/src/task.rs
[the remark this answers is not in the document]
bjchambers: This is related to re-entrancy.

If we schedule "run partition 1 of this transform pipeline on batch 1" and it isn't currently scheduled, then we need to add it to a local queue. If we then get *another* request to "run partition 1 of this transform pipeline on batch 2", we don't want to schedule it again (since that could lead to running it on two threads). The first time this returns "true" (indicating we should schedule that pipeline/partition) the second time it returns "false" (indicating we already have it in a queue).

Then, while we're running "partition 1 of this transform pipeline on batch 1", if we call this it will return false (causing us not to immediately schedule it), but while we're finishing we can detect "this partition was scheduled while we were running it, so add it to the queue again".

crates/sparrow-scheduler/src/worker.rs
```diff
@@ -0,0 +1,74 @@
+use crate::{queue::*, Error, TaskRef};
+
+/// An injector that allows adding work to the global queue.
```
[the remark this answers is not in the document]
bjchambers: We have one global queue that all of the local queues can steal from.
If we're adding work from outside of a worker thread, the only thing we can do is add it to the global queue and let one of the workers steal it.
If we're adding work from within a worker thread, we can choose to add it to the front of the worker, to the back of the worker, or to the global queue, depending on locality benefits.
If the local queue gets full it pushes work to the global queue.

crates/sparrow-scheduler/src/worker.rs
bjchambers: Workers take tasks from their queue and run them. Remember that with morsel driven parallelism we prefer one thread focusing on a specific batch of data and running *all pipelines*. This is actually why the pipelines themselves need mutexes and things in the pipelines so that we can make sure that when a thread starts working on it it has exclusive rights to that.

crates/sparrow-transforms/src/transform_pipeline.rs
bjchambers: Transform pipelines run on a single input, so it should always be 0.

crates/sparrow-transforms/src/transform_pipeline.rs
bjchambers: Whoops. This should be 0.

crates/sparrow-transforms/src/transform_pipeline.rs
```diff
@@ -0,0 +1,174 @@
...
+            // or when we pushed a batch after execution started.
+        }
+
+        Ok(())
+    }
+
+    fn do_work(
+        &self,
+        partition: Partition,
+        queue: &mut dyn Queue<TaskRef>,
```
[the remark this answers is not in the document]
bjchambers: 1. Each pipeline has internal state reflecting the work to be done by the pipeline.
2. Producers call `push` on the pipeline partition to queue data to be processed within the pipeline.
3. The `do_work` call causes the pipeline to process a single "unit" of data within a partition.
4. `push` and `do_work` receive a `queue` that allows them to enqueue additional work.

So, when we push a batch to a partition, we queue the work to be processed. 
Then, if nothing else interrupts us, we'll immediately start processing that batch on the same core.

crates/sparrow-transforms/src/transforms/project.rs
```diff
@@ -0,0 +1,51 @@
...
+            .change_context_lazy(|| Error::CreateTransform("project"))?;
+        Ok(Self {
+            evaluators,
+            outputs,
+            schema,
+        })
+    }
+}
+
+impl Transform for Project {
```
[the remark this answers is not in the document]
bjchambers: Each step in the physical plan will implement `Transform` or `Pipeline`. `Transform` will be used if the operation can easily run in a sequence -- generally 1 input and 1 output transforms and be part of `TransformPipeline`. `Pipeline` will be used for more sophisticated operations -- things that have multiple inputs, or require repartitioning, etc.

crates/sparrow-transforms/src/transforms/project.rs
bjchambers: Yes. Consider: `output: a + b + c`. The only column we need to project is `output`, but to compute that we first need to compute `a + b` and then use that to compute `a + b + c`. So we'll end up with expressions like:

```
0: a
1: b
2: 0 + 1
3: c
4: 2 + 3
```

And the only output column will be column 4.

crates/sparrow-transforms/src/transform_pipeline.rs
```diff
@@ -0,0 +1,174 @@
...
+use sparrow_arrow::Batch;
+use sparrow_scheduler::{
+    Partition, Partitioned, Pipeline, PipelineSink, Queue, Scheduler, Sink, TaskRef,
+};
+
+use crate::transforms::Transform;
+use crate::Error;
+
+/// Runs a linear sequence of transforms as a pipeline.
+pub struct TransformPipeline {
```
bjchambers, line 16: TODO: Mention that transforms shouldn't change the partitioning of their batch (eg., they are effectively "map this sequence of transforms over the batches in this partition"

crates/sparrow-scheduler/src/pipeline.rs
therapon: > 
>     1. The entire query plan consists of a bunch of "transforms" (lookup request, lookup response, project, select, etc.)
> 
>     2. We divide those "transforms" up into "pipeline plans" that are a single input -> sequence of transforms -> single output.
> 

  Do you mean we "group" rather than "divide" here? Are transforms the leafs of an execution (indivisible computation like an `add`) or are they decomposable? My reading of this let me to understand that transforms are the actions and pipelines compose them together. 

 What really throws me off here is that we "divide" transforms into pipeline plans that consists of a sequence of transforms again? 

  Maybe my extrapolation from the English words is getting in the way here. I would expect a pipe to be an individual element. Pipelines to compose these pipes together in some way. 
I am not seeing this connection here with transformers, pipeline plans, pipelines and the like. 

> 
>     4. Each pipeline manages the state for _all_ partitions that it may execute on. This is intentional, since it allows things like the "repartition" pipeline to accept data from any producing / input partition and store it in queues for each of the downstream partitions.

 So partitions have to do with what flows into the pipe, i.e., data. What is the relationship with a Batch? Is a batch a segment of data from 1 partition only, or, can it have data from multiple partitions?  


> 
>     5. The above means that a pipeline has "input partitions" (the partitioning of the producer) and "processing / output partitions" (the partitioning of the actual pipeline and the output it produces).

  These partitions are something that is checked/verified so that outputs from one pipeline "match" inputs to the succeeding pipeline? 

> 
>     6. A pipeline has 0 or more inputs (eg., a merge operation has two inputs, many things have 1 input, etc.). Thus, inputs are associated with the "input partition" (the partition that produced the input) and the "input index" (which of the inputs to the pipeline it is being sent to.

  I think I am missing something here. We have multiple inputs so there is an index for each input, I assume. Now each input has partitions so there is another index within the input to designate a partition, e.g., 1:3 for input 1 and partition 3 of input 1? Is this the setup? Are partitions internal components of inputs? 

> 
>     7. Children I don't think matters (it's basically how pipelines are connected and/or which transforms are connected together within a pipeline).
>

  Well I would expect children to matter a little since if they specify connections we need to ensure that the connection have right input+partition for the next step in the pipeline, no? 
 
>     8. Batches are data that flows between pipelines.

  But what is the relationship here. Is batch a segment of a stream of data coming from 1 partition connected to 1 input *only*?

crates/sparrow-scheduler/src/queue.rs
therapon: If the queue is global isn't easier to make that function part of the global queue and leave it out of this trait? Or are you looking at having the same trait for both global and local queues?

crates/sparrow-scheduler/src/pipeline.rs
```diff
@@ -0,0 +1,128 @@
...
+    ///
+    /// The number of tasks provided indicate the number of partitions this pipeline
+    /// should execute with.
+    ///
+    /// This is unfortunately separate to allow cyclic initialization. Specifically,
+    /// `Arc::new_cyclic` expects a function that doesn't take errors, while creating
+    /// many pipelines *does* produce errors. To address this, we first create the
+    /// part of the pipeline that doesn't need to reference it's own tasks, and then
+    /// we initialize it as part of `Arc::new_cyclic`.
+    fn initialize(&mut self, tasks: Partitioned<TaskRef>);
```
jordanrfrazier, line 88: Do you have any future concerns about how this cyclic dependency is handled?

crates/sparrow-scheduler/src/pipeline.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// Schedules any tasks that need to be executed on the `scheduler`.
+    fn add_input(
+        &self,
+        input_partition: Partition,
+        input: usize,
+        batch: Batch,
+        scheduler: &mut dyn Scheduler,
+    ) -> error_stack::Result<(), PipelineError>;
+
+    /// Mark an input partition and input index as complet.
```
jordanrfrazier, line 106: complete

crates/sparrow-scheduler/src/worker.rs
```diff
@@ -0,0 +1,105 @@
+use crate::{queue::*, Error, TaskRef};
+
+pub trait Scheduler {
+    /// Schedule a task for immediate, local execution.
+    ///
+    /// For local queues, this will schedule it as the next task, potentially
+    /// displacing the other task(s) scheduled at the front.
```
jordanrfrazier, line 7: Is the reasoning for this (the whole attempt-to-use-the-same-core-for-cached-data) explained elsewhere?

crates/sparrow-scheduler/src/worker.rs
```diff
@@ -0,0 +1,105 @@
...
+    ///
+    /// For local queues, this can be used to give other tasks a chance to run.
+    /// Otherwise, there’s a risk that one task will completely take over a
+    /// thread in a push-pop cycle due to the LIFO slot.
+    ///
+    /// If the local queue is full, it will move half of its tasks to the global
+    /// queue.
+    fn schedule_yield(&mut self, task: TaskRef);
+
+    /// Schedule n item for eventual execution anywher.
```
jordanrfrazier, line 27: typos

crates/sparrow-scheduler/src/worker_pool.rs
```diff
@@ -0,0 +1,139 @@
...
+    ///
+    /// `partitions` determines the number of task partitions to allocate.
+    pub fn add_pipeline<T>(&mut self, partitions: usize, pipeline: T) -> Arc<dyn Pipeline>
+    where
+        T: Pipeline + 'static,
+    {
+        let index = self.pipelines.len();
+        let name = std::any::type_name::<T>();
+
+        let pipeline: Arc<T> = Arc::new_cyclic(move |weak| {
```
jordanrfrazier, line 99: Perhaps a good spot to explain the use of the `weak` reference?

crates/sparrow-transforms/src/transform.rs
```diff
@@ -0,0 +1,24 @@
+use sparrow_arrow::Batch;
+
+#[derive(derive_more::Display, Debug)]
+pub enum Error {
+    #[display(fmt = "failed to create {_0} transform")]
+    CreateTransform(&'static str),
+    #[display(fmt = "failed to create {_0} transform")]
```
jordanrfrazier, line 7: failed to execute

crates/sparrow-transforms/src/transform_pipeline.rs
```diff
@@ -0,0 +1,264 @@
...
+    ) -> error_stack::Result<(), PipelineError> {
+        let partition = &self.partitions[input_partition];
+
+        let Some(batch) = partition.pop_input() else {
+            error_stack::ensure!(
+                partition.is_input_closed(),
+                PipelineError::illegal_state("scheduled without work")
+            );
+            return self.sink.close_input(input_partition, scheduler);
+        };
```
jordanrfrazier, line 220: Hm, so the logic here is that if a worker calls `do_work`, but the partitioned has no input scheduled, we'll close the sink (consumer) partition. However, if this pipeline's input is not already closed, then we'll error, because we expected this partition to already be closed by...the producing pipeline?

jordanrfrazier (approved):

crates/sparrow-transforms/src/transform_pipeline.rs
```diff
@@ -0,0 +1,264 @@
...
+        // If the input is closed and empty, then we should close the sink.
+        if partition.is_input_closed() && partition.is_input_empty() {
+            self.sink
+                .close_input(input_partition, scheduler)
+                .change_context(PipelineError::Execution)?;
+        }
+
+        // Note: We don't re-schedule the transform if there is input.
+        // This should be handled by the fact that we scheduled the transform
+        // when we added the batch, which should trigger the "sceduled during
```
jordanrfrazier, line 259: scheduled

crates/sparrow-scheduler/src/pipeline.rs
bjchambers: Not particularly. I don't *love* the way this is handled (in terms of the two steps). If there was an easier way to construct the cycle (`Arc::new_cyclic...`) I wouldn't be opposed to it. But, it works, and it doesn't use excess trickery. It won't ever actually keep things around since the `new_cyclic` ensures that the cycles go through `Weak` references.

And the messiness only affects the `Pipeline` implementations, which we should have relatively few of (and it's far from the hardest part about writing the pipeline implementations).

crates/sparrow-scheduler/src/task.rs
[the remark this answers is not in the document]
bjchambers: Updated wording.

crates/sparrow-scheduler/src/worker.rs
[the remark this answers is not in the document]
bjchambers: Addressed?

crates/sparrow-transforms/src/transforms/project.rs
```diff
@@ -0,0 +1,51 @@
...
+use sparrow_arrow::Batch;
+
+use crate::transforms::Transform;
+use crate::Error;
+use sparrow_expressions::ExpressionExecutor;
+
+/// Transform for projection.
+pub struct Project {
+    evaluators: ExpressionExecutor,
+    outputs: Vec<usize>,
```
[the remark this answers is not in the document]
bjchambers: There are N expressions to be evaluated. Each output column is a number `0..N` indicating which expression computed the column.

crates/sparrow-transforms/src/transforms/project.rs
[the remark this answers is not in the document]
bjchambers: Still open?

crates/sparrow-transforms/src/transforms/project.rs
bjchambers: ?

crates/sparrow-transforms/src/transform_pipeline.rs
bjchambers: ?

crates/sparrow-scheduler/src/worker.rs
bjchambers: Not particularly -- that should be part of the top level docs, which I'll add once this settles a bit. Basically, it would just be a reference to "Morsel driven parallelism".

crates/sparrow-transforms/src/transform_pipeline.rs
bjchambers: Roughly. If we get here, then *this pipeline* asked to be scheduled.

1. It may have been requested in `add_input`, in which case there should be input.
2. It may have been requested in `close_input` to signal things are closed.
3. It may be the previous `do_work` realizing there is more to do.

In any of those cases, there should either be a batch to process or we should be all closed up, or there is a bug.

Ideally, (2) could detect "nothing is running, go ahead and close directly", but that is tricky. Since it just happens at the end, it doesn't seem worth guarding against.

crates/sparrow-scheduler/src/queue.rs
bjchambers: Trying to have the same trait. The reason being that it lets us call `add_input(...)` from outside of a worker thread (eg., from tokio) and pass the global queue (so we default to queuing work) and `add_input(...)` from the worker thread adds to the local queue.

On the other hand, I could be pretty easily convinced to treat sources as special, which would let us use a different trait for them, that received the global queue only.

PR #532, opened 2023-07-18
feat: support more types for get evaluator for maps

jordanrfrazier:
Uses get index kernels to simplify evaluator types. Supports `Boolean/String/Primitive` types for `get` on maps. 

Part of https://github.com/kaskada-ai/kaskada/issues/494


crates/sparrow-compiler/tests/compiler_golden_tests.rs
```diff
@@ -614,16 +614,14 @@ async fn test_incompatible_err() {
...
       |
     1 | Sent.amount == "hello"
       | ----------- ^^ ------- Type: string
       | |           |
-      | |           Invalid types for call to 'eq'
+      | |           Incompatible types for call to 'eq'
       | Type: f64
-      |
-      = Expected 'any'
     "###);
```
jordanrfrazier, line 625: @kerinin @bjchambers Thoughts on this change? 

When two types to a function are incompatible (cannot be promoted to one another), this error will take precedence over the `InvalidArgument` error. 

Reasoning: when doing the `get` function it didn't make sense to see 

```
key: i64, Map.key: string
= Expected 'key'
```

because both of these were of the `key` type, they just weren't compatible with each other.

crates/sparrow-instructions/src/evaluators/macros.rs
```diff
@@ -315,16 +315,92 @@ macro_rules! create_typed_evaluator {
...
+            DataType::UInt64 => {
+                create_primitive_map_evaluator!(
+                    UInt64Type,
+                    $value_type,
+                    $primitive_to_primitive_evaluator,
+                    $primitive_to_string_evaluator,
+                    $info
+                )
+            }
+            // TODO: Bool
```
jordanrfrazier, line 403: TODO

crates/sparrow-instructions/src/evaluators/map/get_primitive_to_string.rs
```diff
@@ -0,0 +1,122 @@
...
+                            } else {
+                                Ok(Some(m_values.value(value_pos).to_owned()))
+                            }
+                        }
+                        None => Ok(None),
+                    }
+                })
+                .try_collect()?;
+
+            GenericStringArray::<O>::from_iter(values.iter())
```
jordanrfrazier, line 117: Unsure if there's a better way to create this -- can I avoid the clone at 109 too?

crates/sparrow-main/tests/e2e/basic_error_tests.rs
```diff
@@ -380,19 +380,17 @@ async fn test_incompatible_actual_types_error() {
...
         formatted:
-          - "error[E0010]: Invalid argument type(s)"
+          - "error[E0015]: Incompatible argument types"
           - "  --> Query:1:16"
           - "  |"
           - "1 | { m: Strings.s + Strings.n }"
           - "  |      --------- ^ --------- Type: i64"
           - "  |      |         |  "
-          - "  |      |         Invalid types for call to 'add'"
+          - "  |      |         Incompatible types for call to 'add'"
```
jordanrfrazier, line 392: This change is _not_ as desirable in this context, perhaps?

testdata/parquet/data_with_long_to_string_map.parquet
jordanrfrazier, line 1: remove

testdata/parquet/supplant.parquet
jordanrfrazier, line 1: remove

testdata/parquet/data_with_invalid_map_types.parquet
jordanrfrazier, line 1: remove

testdata/test.parquet
jordanrfrazier, line 1: remove

testdata/test2.parquet
jordanrfrazier, line 1: remove

crates/sparrow-main/tests/e2e/basic_error_tests.rs
bjchambers: Seems OK? But why did we need a new "incompatible" when we had "invalid"? Does invalid still show up ever?

bjchambers (approved):

crates/sparrow-main/tests/e2e/basic_error_tests.rs
jordanrfrazier: We were conflating two separate scenarios to the `Invalid` error 
1. When two types for a type class did not have a least upper bound 
2. When the least upper bound from step 1 could not be promoted to a type that satisfies the type class. 

So I had the first instead return `Incompatible`, attempting to imply the argument types were incompatible with each other (because they do not have a least upper bound).

jordanrfrazier: Easy to fix up if it ends up being confusing, so going to resolve and merge this.

Issue #352, opened 2023-05-10
bug: fenl catalogue generation requires a manual process since we moved to antora


therapon:
**Description**

We have a `sparrow-catalgue` binary that generates the documentation for the catalogue of fenl functions. After moving to Antora the program's output needs to be translated and updated in the docs manually

- [ ] update sparrow code to generate asciidoctor documentation
- [ ] update antora site generation. See [possible solutions](https://intellij-asciidoc-plugin.ahus1.de/docs/users-guide/antora/generated-content-in-antora.html)

PR #537, opened 2023-07-19
feat: Change catalog to produce asciidoc

bjchambers:


bjchambers:
This is part of #352. This produces separate AsciiDoc pages for each catalog entry. Let's discuss steps for connecting it. I think I see:

1. Run the existing binary built in the main CI to produce a zip that we archive. We can then use https://gitlab.com/antora/antora-collector-extension to download the zip and extract it. This potentially has issues with versioning.
2. Use https://gitlab.com/antora/antora-collector-extension with a script that builds the catalog binary and uses it. This will currently be slow. We could improve it by creating a version of the catalog binary that doesn't include sparrow (eg., a "generate-only" binary). That would have far fewer dependencies and be much faster to build.
3. Some other approach, possibly having the binary push the generated stuff to a separate "kaskada-catalog" repo, keeping a main branch in sync, as well as tagged branches for each release.

One benefit of that last option is that it makes it possible to manually edit old catalogs.

therapon (approved):

PR #540, opened 2023-07-21
feat: support first and last non-windowed aggregations over DataType::Map

jordanrfrazier:
Support non-windowed aggregations `first` and `last` over map types.


crates/sparrow-compiler/src/functions/pushdown.rs
```diff
@@ -110,7 +110,8 @@ impl Pushdown {
             | DataType::Duration(_)
             | DataType::Interval(_)
             | DataType::Utf8
-            | DataType::LargeUtf8 => {
+            | DataType::LargeUtf8
+            | DataType::Map(..) => {
```
jordanrfrazier, line 114: Probably need to rename this, since it's currently `PrimitivePushdown`, but :shrug:

crates/sparrow-instructions/src/evaluators/aggregation/map/first_map_evaluator.rs
```diff
@@ -0,0 +1,169 @@
...
+                    take_new_state[entity_index as usize] =
+                        (input_index + take_new_state.len()) as u32;
+                }
+            };
+
+            take_output_builder.append_value(take_new_state[entity_index as usize])
+        }
+
+        // TODO: Clone?
+        let old_state: ArrayRef = Arc::new(token.accum().clone());
```
jordanrfrazier, line 131: Can I avoid this clone?

crates/sparrow-instructions/src/evaluators/aggregation/map/first_map_evaluator.rs
```diff
@@ -0,0 +1,169 @@
...
+    use std::sync::Arc;
+
+    use arrow::array::{Float64Array, Int64Array};
+    use arrow::datatypes::{Float64Type, Int64Type};
+
+    use super::*;
+    use crate::{FirstPrimitive, LastPrimitive, Max, Mean, Sum};
+
+    #[test]
+    fn test_sum_f64() {
```
jordanrfrazier, line 154: Add some tests

crates/sparrow-instructions/src/evaluators/aggregation/token/map_accum_token.rs
```diff
@@ -0,0 +1,61 @@
...
+use crate::{ComputeStore, StateToken, StoreKey};
+/// Token used for string accumulators.
+///
+/// String accumulators are stored as `[passId, instId, entity_index] ->
+/// Option<String>`. Values are updated entity-by-entity.
+// #[derive(serde::Serialize, serde::Deserialize)]
+pub struct MapAccumToken {
+    /// Stores the state for in-memory usage.
+    // #[serde(with = "sparrow_arrow::serde::array_ref")]
+    accum: MapArray,
```
jordanrfrazier, line 12: So the `serde` needs to change to be able to support `MapArray`? Wasn't sure of this and haven't looked into it yet.

crates/sparrow-instructions/src/evaluators/equality.rs
```diff
@@ -30,13 +31,58 @@ impl EvaluatorFactory for NeqEvaluatorFactory {
...
+            MapNeqEvaluator,
             BoolNeqEvaluator,
             StringNeqEvaluator,
             info
         )
     }
 }
 
+/// Evaluator for `eq` on maps.
+struct MapEqEvaluator {
```
jordanrfrazier, line 43: Snuck these in here since they were easy and allowed me to continue to use the `create_typed_evaluator` macro for first/last.

crates/sparrow-main/tests/e2e/collection_tests.rs
```diff
@@ -67,6 +71,57 @@ async fn test_get_dynamic_key() {
...
+    let file = std::fs::File::open(hash).unwrap();
+    let parquet_reader = ParquetRecordBatchReaderBuilder::try_new(file)
+        .unwrap()
+        .with_batch_size(1000);
+    let reader = parquet_reader.build().unwrap();
+    let schema = reader.schema();
+    let batches: Vec<_> = reader.try_collect().unwrap();
+    let concatenated = arrow::compute::concat_batches(&schema, &batches).unwrap();
+    println!("{:?}", concatenated);
+}
```
jordanrfrazier, line 98: Verified reading the parquet file that the values are as expected! Unfortunately can't print a map in csv, but can add unit tests to do `first/last`, then a `get`.

crates/sparrow-instructions/src/evaluators/equality.rs
jordanrfrazier: Although -- map equality is not yet supported by arrow. https://docs.rs/arrow-ord/44.0.0/src/arrow_ord/comparison.rs.html#1746

jordanrfrazier: We'll return an internal error if this happens with a sparrow log noting that this is unsupported.

bjchambers: Given we want to use `eq_dyn`, we should be able to use an `EqEvaluator` that works for all types.

Within that, we could detect that we're on a map, and then do equality on the key, value and offsets separately. Alternatively, I'd suggest you add a PR to Arrow to support `eq_dyn` on map (shouldn't be too hard to add) and would let us use it.

crates/sparrow-main/tests/e2e/collection_tests.rs
```diff
@@ -1,5 +1,4 @@
 //! e2e tests for the collection operators.
```
bjchambers, line 1: Suggest naming this `map_tests.rs`.

crates/sparrow-main/tests/e2e/collection_tests.rs
```diff
@@ -109,6 +108,42 @@ async fn test_bool_to_s_get_static_key() {
     "###);
 }
 
+#[tokio::test]
+async fn test_first_last_map() {
+    // The csv writer does not support map types currently, so the output has been verified
```
bjchambers, line 113: What if we allowed running to JSON? We could add that as a fast follow-up pretty easily (I have some changes that may get me deep in the testing code, and may be able to pull that in as well).

bjchambers (approved):

crates/sparrow-instructions/src/evaluators/equality.rs
jordanrfrazier: Added a task to the issue. Planning to merge this as is to unblock a demo of map aggregations then follow up with one of the two solutions.

crates/sparrow-main/tests/e2e/collection_tests.rs
jordanrfrazier: That would be nice -- as discussed, will leave this for a follow-up if it fits into your testing changes.

PR #542, opened 2023-07-21
removed redis destination

epinzur:
we haven't supported this for a long time, lets remove it from the codebase


tests/integration/shared/helpers/helpers.go
```diff
@@ -131,7 +131,48 @@ func downloadFile(url string) (localPath string, cleanup func()) {
...
 }
 
-func GetFileURI(fileName string) string {
+// Gets an output path URI
+func GetOutputPathURI(subPath string) string {
+	subPath = vfs_utils.EnsureTrailingSlash(subPath)
+	if os.Getenv("ENV") == "local-local" {
+		workDir, err := os.Getwd()
+		Expect(err).ShouldNot(HaveOccurred())
+		path := filepath.Join(workDir, "../data/output", subPath)
```
therapon, line 140: Maybe consider passing this value in rather than hardcoding?

epinzur: this is for tests only... i think it's okay to hardcode

we already do it elsewhere for other integration test helpers

kerinin (approved):

PR #543, opened 2023-07-21
feat: Add serde support for arrays

bjchambers:
Also adds tests for `ArrayRef` and `MapArray` serialization.

### DIFF
--- a/crates/sparrow-arrow/src/serde.rs
+++ b/crates/sparrow-arrow/src/serde.rs
@@ -93,6 +93,51 @@ pub mod array_ref {
     }
 }
 
+/// Provides serde for specific `Array` values.
+///
+/// Example:
+///
+/// ```rust
+/// #[derive(serde::Serialize, serde::Deserialize)]
+/// struct Foo {
+///   #[serde(with = "sparrow_arrow::serde::array")]
+///   array: MapArray
+/// }
+/// ```
+pub mod array {
+    use std::sync::Arc;
+
+    use arrow::array::ArrayRef;
+    use arrow_array::cast::downcast_array;
+    use arrow_array::Array;
+
+    pub fn serialize<T: Array + Clone + 'static, S>(
+        array: &T,
+        serializer: S,
+    ) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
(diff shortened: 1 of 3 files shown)


crates/sparrow-arrow/src/serde.rs
```diff
@@ -93,6 +93,51 @@ pub mod array_ref {
...
+        serializer: S,
+    ) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        // We want to serialize the array using logic for RecordBatch.
+        // This requires we get an `ArrayRef` for it.
+        // This requires creating an owned array.
+        // This cloning should only do a shallow copy.
+        let array: T = array.clone();
```
bjchambers, line 125: I'm not sure I love this implementation. It may be better to encourage:

```
struct Foo {
  #[serde(with = "sparrow_array::serde::array_ref")]
  array: ArrayRef,
}

impl Foo {
  fn array(&self) -> &MapArray {
     self.array.as_map()
  }
}
```

Also, the methods here may be useful in a lot of places (where we currently downcast to a shouldn't-panic type): https://docs.rs/arrow/latest/arrow/array/trait.AsArray.html#method.as_map

We can just do `array_ref.as_map()` instead of `downcast_map(array_ref.as_ref())?`.

jordanrfrazier: Hm wonder how that's different from `as_map_array(..)` https://arrow.apache.org/rust/src/arrow_array/cast.rs.html#687

jordanrfrazier (approved):

crates/sparrow-arrow/src/serde.rs
bjchambers: Nearly the same:

```
// Function that takes `&dyn Array`
fn as_map_array(array: &dyn Array) -> &MapArray { ... }

// Trait that lets you convert things in varying ways:
// https://docs.rs/arrow/latest/arrow/array/trait.AsArray.html
trait AsArray {
  fn as_map_opt(&self) -> Option<&MapArray>;
  fn as_map(&self) -> &MapArray;
  // many other methods
}

// Impl for Arc
impl AsArray for Arc<dyn Array> { ... }
```

PR #545, opened 2023-07-21
feat: Use new hashing     

### LINKED ISSUE #211: feat: Support hashing of structs
**Summary**
Support hashing of struct types. Specifically, it should be possible to use structs to create composite keys for the purposes of `with_key`, `lookup`, etc.

**Is your feature request related to a problem? Please describe.**
For some use cases, I'd like to compute "grouped" aggregations. For instance, for each user, I'd like to count how frequently each category appears. This is possible if I rekey using `{ user, category }` as a key. But such a value is currently not "hashable", so it can't be used as a key.

**Use Case**
See problem.

**Describe the solution you'd like**
Structs should be hashable.

**Describe alternatives you've considered**
Alternatives for this specific use case include special aggregations that compute "sub-grouped" values. Those are likely *still useful* from a syntactic perspective, but the generality of being able to arbitrarily re-key and lookup is higher. We may end up with both, but extending the functionality of `with_key` and `lookup` in this reasonable way seems like a better first step.

### LINKED ISSUE #462: feat: Update hash to use the new `Hasher`
**Summary**
PR #454 introduces a new hasher based on the code from datafusion. This re-uses the space allocated for computing hashes (an efficiency win) and supports `structs` (a functionality improvement). We should probably switch to using it for everything.

1. We can use a `thread_local` to store a Hasher for each thread (or alternatively, for each CPU). This would allow us to re-use the Hasher with minimum effort.
2. This does change the hash values -- specifically, `e` used to hash as `Some(e)` and `null` used to hash as `None`. But now they hash as `e` and `0` respectively. This should be OK, although we may want to bump the prepare version.

### LINKED ISSUE #525: feat: Support hashing of structs
**Summary**

Current code uses an old hashing implementation (https://github.com/kaskada-ai/kaskada/blob/1421def1cab6c563480550a65d8f23332c7b8613/crates/sparrow-arrow/src/hash.rs#L29).

We have already added a hasher that supports structs (https://github.com/kaskada-ai/kaskada/blob/1421def1cab6c563480550a65d8f23332c7b8613/crates/sparrow-arrow/src/hasher.rs#L42).

In addition, it should be faster since it can reuse the hashing buffer. One easy way would be to re-implement the hash function from the first file using a thread-local hasher (similar to https://github.com/kaskada-ai/kaskada/blob/1421def1cab6c563480550a65d8f23332c7b8613/crates/sparrow-expressions/src/evaluators/hash.rs#L26).

- [ ] Change hash function to use thread-local hasher (hash evaluator can use the same thread-local).
- [ ] Add test for hashing of structs (and `with_key` / `lookup`)
- [ ] Update prep version and other things (since the hash function changes)

bjchambers:
This uses the new Hasher based implementation, that should be faster
(avoids re-allocating the buffer) and supports structs.
    
This closes #211.
This closes #462.
This closes #525.


crates/sparrow-main/src/serve/snapshots/sparrow_main__serve__compute_service__tests__sliced_query_smoke.snap
```diff
@@ -5,14 +5,14 @@ expression: results
 - state: 2
   is_query_done: true
   progress:
-    total_input_rows: 54068
-    processed_input_rows: 54068
+    total_input_rows: 49540
```
jordanrfrazier, line 8: This is a weird change -- total input rows changed?

crates/sparrow-catalog/catalog/yearly.toml
```diff
@@ -66,8 +66,8 @@ time,key,n
 '''
 output_csv = '''
 time,key,n
-1996-12-19T16:00:00.000000000,Ben,2
 1996-12-19T16:00:00.000000000,Ryan,3
-1997-12-20T16:01:00.000000000,Ben,9
+1996-12-19T16:00:00.000000000,Ben,2
 1997-12-21T16:00:00.000000000,Ryan,8
+1997-12-20T16:01:00.000000000,Ben,9
```
jordanrfrazier, line 72: Many of these results were reordered -- I assume that is because the `_key_hash` changed (which we drop) and reordered the results according to the hash?

crates/sparrow-catalog/catalog/hash.toml
```diff
@@ -50,10 +50,10 @@ time,key,value
...
-2021-01-01T00:00:00.000000000,Ryan,8,4864632034659211723
-2021-01-02T00:00:00.000000000,Ryan,9,2122274938272070218
-2021-01-03T00:00:00.000000000,Ben,8,4864632034659211723
-2021-01-04T00:00:00.000000000,Ben,,11832085162654999889
-2021-01-04T00:00:00.000000000,Ryan,9,2122274938272070218
+2021-01-01T00:00:00.000000000,Ben,5,16461383214845928621
+2021-01-01T00:00:00.000000000,Ryan,8,6794973171266502674
+2021-01-02T00:00:00.000000000,Ryan,9,15653042715643359010
+2021-01-03T00:00:00.000000000,Ben,8,6794973171266502674
+2021-01-04T00:00:00.000000000,Ben,,0
```
jordanrfrazier, line 57: Does this imply that `None` of any type hashes to `0`? And is that a different behavior than before (i.e. did `None(Int64)` hash the same as `None(String)`?

crates/sparrow-runtime/src/prepare/prepare_parquet.rs
```diff
@@ -0,0 +1,19 @@
+use error_stack::ResultExt;
+use sparrow_api::kaskada::v1alpha::{slice_plan, PreparedFile, TableConfig};
+
+use crate::prepare::Error;
+use crate::read::ParquetFile;
+use crate::stores::{ObjectStoreRegistry, ObjectStoreUrl};
+
+pub async fn prepare_parquet<'a>(
```
jordanrfrazier, line 8: Is this used elsewhere related to this change?

crates/sparrow-runtime/src/prepare/slice_preparer.rs
```diff
@@ -222,16 +219,21 @@ mod tests {
         .unwrap();
 
         let sliced_batch = preparer.slice_batch(batch).unwrap();
-        let expected_batch_size = 3;
+        let expected_batch_size = 2;
```
jordanrfrazier, line 222: Why would the hashing have affected this?

crates/sparrow-catalog/catalog/yearly.toml
bjchambers: I believe so, yes.

crates/sparrow-catalog/catalog/hash.toml
bjchambers: Previously, they did hash differently. I think this is OK -- we're just saying that our hash function collides on null values of different types. But note we shouldn't be using hash on different types in the same place.

crates/sparrow-runtime/src/prepare/prepare_parquet.rs
bjchambers: Nope. This file snuck in from an unrelated branch.

crates/sparrow-runtime/src/prepare/slice_preparer.rs
bjchambers: See previous comment on slicing.

bjchambers:
@epinzur looks like this is failing some of the wren integration tests for slicing. My best guess is that it's similar to the Rust test changes -- since we use the hash to determine which entities are in a slice, changing the hash changes the slices. I can take a look, but you can probably knock that out quicker than I can. Any thoughts here?

epinzur:
@bjchambers I added a commit to fix the slicing test.  The issue was that with the new hashing function, a 0.3% slice returns  no results.  I changed the test to a 1% slice and it now passes.

PR #553, opened 2023-07-25
feat: add prototype python query builder

bjchambers:
This adds a new top-level crate `sparrow-py`. It is intentionally outside of `crates` since it needs to be in a separate workspace to properly compile the Python code.

This supports basic builders, implements most of the math operators directly on expressions, and supports a `pipe` function similar to Pandas.

In Python 3.11 the exceptions are printed with locations highlighted and type information of each argument is printed.


sparrow-py/pysrc/sparrow_py/expr.py
```diff
@@ -0,0 +1,209 @@
...
+from typing import Callable
+from typing import Tuple
+from typing import Union
+from typing import final
+
+from sparrow_py import ffi
+
+
+class Expr(object):
+    """A Kaskada expression."""
```
bjchambers, line 13: Should improve the pydocs at some point based on existing docs, but this is a good start.

sparrow-py/noxfile.py
```diff
@@ -0,0 +1,181 @@
+"""Nox sessions."""
```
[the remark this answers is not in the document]
bjchambers: Would probably be good to run in CI at some point, but that can wait until we stabilize this.

sparrow-py/pysrc/sparrow_py/expr.py
```diff
@@ -0,0 +1,209 @@
...
+        Raises
+        ------
+        TypeError # noqa: DAR402
+            If the argument types are invalid for the given function.
+        """
+        ffi_args = [arg.ffi for arg in args]
+        try:
+            self.ffi = ffi.Expr(name, ffi_args)
+        except TypeError as e:
+            e = TypeError(str(e))
```
bjchambers, line 37: Want to add a test for this error augmentation in Python 3.11.

sparrow-py/pysrc/sparrow_py/expr.py
```diff
@@ -0,0 +1,209 @@
...
+    def __str__(self) -> str:
+        """Return a readable string representing the expression."""
+        return str(self.ffi)
+
+    def data_type(self) -> str:
+        """Return a string describing the data type."""
+        return self.ffi.data_type_string()
+
+    @final
+    def pipe(
```
bjchambers, line 58: This is currently based on Pandas. It allows:

`a.pipe(f, b...) => f(a, b...)`

and

`a.pipe((f, k), b...) => f(b..., k=a)`

It has errors if the first arguments (that use `b...`) are the ones named `k`. We can probably fix that with some code changes -- so we can decide how we want to implement this. I like this as an alternative to adding a bunch of functions (and thus names) to the `Expr` class, and also as a way of documenting the behavior of pipe.

sparrow-py/pysrc/sparrow_py/expr.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def __and__(self, rhs: "Expr") -> "Expr":
+        """Logical and."""
+        return Expr("and", self, rhs)
+
+    def __or__(self, rhs: "Expr") -> "Expr":
+        """Logical or."""
+        return Expr("or", self, rhs)
+
+    def __getitem__(self, key: Union["Expr", int, float, str]) -> "Expr":
```
bjchambers, line 181: This allows:

`expr['foo']` on structs, and `expr['bar']` on maps, etc.

One thing to note here is that it uses `Union["Expr", int, float, str]` instead of the more typical `Expr`. This allows passing a literal in without explicitly annotating it (eg., `Literal(5)` makes an expressions, but you can omit it here).

My general feeling is that *in general* we should prefer `Expr`, since that requires the `Literal(5)` and makes it clear that we're "freezing" that value. Otherwise, we could have:

`five + date.now()` and that would be treated as `five + Literal(date.now())` -- it would freeze the `date.now()` at whatever Python returned when the query was created, rather than the re-evaluating it each time. It seems like in many places the explicit `Literal(date.now())` to change a python value into an expression is desirable.

sparrow-py/pysrc/sparrow_py/ffi.pyi
```diff
@@ -0,0 +1,14 @@
+from typing import Sequence
+
+class Session:
```
bjchambers, line 3: Type annotations ("stubs") for the Rust-exposed FFI.

sparrow-py/pytests/expr_test.py
```diff
@@ -0,0 +1,43 @@
...
+    five = Literal(5)
+    six = Literal(6)
+    five_lhs = five.pipe(math.add, six)
+    assert str(five_lhs) == "(add (literal 5u64) (literal 6u64))"
+
+    six_lhs = five.pipe((math.add, "rhs"), six)
+    assert str(six_lhs) == "(add (literal 6u64) (literal 5u64))"
+
+    six_lhs = five.pipe((math.add, "rhs"), lhs=six)
+    assert str(six_lhs) == "(add (literal 6u64) (literal 5u64))"
```
bjchambers, line 43: Add some more tests for the other expressions / overloads.

sparrow-py/.darglint
```diff
@@ -0,0 +1,3 @@
+[darglint]
+strictness = short
+docstring_style=numpy
```
therapon, line 3: I think the python client code is predominantly Google style. We should decide on one style and go with that. 

Also this is another python tool. We should also make a selection of what python tools we are to use and be consistent across python code in the repo

sparrow-py/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+`sparrow-py` provides a Python library for building and executing Kaskada queries using an embedded Sparrow library.
+It uses [Pyo3][pyo3] to generate Python wrappers for the Rust code, and then provides more Python-friendly implementations on top of that.
+
+[pyo3]: https://github.com/PyO3/pyo3
+
+## Install Python
+
+Use `pyenv` and install at least `3.7` or `3.8`.
+If multiple versions are installed, `nox` will test against each of them.
```
therapon, line 11: Currently the existing python code relies on the CI to test cross versions. I think it would be a good idea to be able to do that locally as well. 

If that is what we want we should add `nox` to the rest of the python code and be consistent.

sparrow-py/docs/conf.py
```diff
@@ -0,0 +1,11 @@
+"""Sphinx configuration."""
```
therapon, line 1: Make our Sphinx config consistent across python code in the repo.

sparrow-py/pysrc/sparrow_py/expr.py
```diff
@@ -0,0 +1,209 @@
...
+
+from sparrow_py import ffi
+
+
+class Expr(object):
+    """A Kaskada expression."""
+
+    ffi: ffi.Expr
+
+    def __init__(self, name: str, *args: "Expr") -> None:
```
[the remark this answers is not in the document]
therapon: One way to look at this is the less versions we support the less work. The other way is that we might have clients that are not going to be able to easily upgrade to use our tools. So ... decisions, decisions. :)

therapon (approved):
Most of the comment are on decisions to make things uniform across our python code. These are not blockers more TODOs really.

sparrow-py/pysrc/sparrow_py/expr.py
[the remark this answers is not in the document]
bjchambers: Agreed. I think for now, >=3.8 is nice. It's relatively minimal:

If we moved to >=3.10 we could use slightly simpler type annotations. (`Expr` instead of `"Expr"`). This doesn't seem worth it.

If we moved to >=3.11 we could be confident we produce better errors. This seems pretty weak -- the user can choose to use the version that produces better errors if they want better errors.

kevinjnguyen (approved):

jordanrfrazier (approved):

PR #558, opened 2023-07-26
ref: Expose a QueryBuilder

bjchambers:
This is pretty ugly -- it has to do create a lot of fake locations and create enums (like `ExprOp`) just to call into the existing AST -> DFG compilation. We should definitely clean this up if we continue down the path of using builders.

### DIFF
--- a/crates/sparrow-compiler/src/data_context.rs
+++ b/crates/sparrow-compiler/src/data_context.rs
@@ -11,7 +11,8 @@ use sparrow_plan::{GroupId, TableId};
 use sparrow_syntax::Location;
 use uuid::Uuid;
 
-use crate::dfg::Operation;
+use crate::dfg::{Dfg, Operation};
+use crate::AstDfgRef;
 
 /// Represents the "data context" for a compilation.
 ///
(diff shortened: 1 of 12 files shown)


jordanrfrazier (approved):

crates/sparrow-compiler/src/data_context.rs
```diff
@@ -432,6 +401,39 @@ impl TableInfo {
...
+        use sparrow_syntax::FenlType;
+
+        use crate::ast_to_dfg::AstDfg;
+        use crate::time_domain::TimeDomain;
+
+        // Add the table reference to the environment.
+        let value = dfg.add_operation(
+            Operation::Scan {
+                table_id: self.table_id,
+                slice: None,
```
jordanrfrazier, line 417: Should you add a `TODO` for supporting slices?

bjchambers: I don't think it's needed here -- specifically, this was extracted from old code for compilation, and it worked with slicing. I believe the reason is that slices aren't relevant during compilation -- we just need them when we run the query.

PR #559, opened 2023-07-26
feat: Prototype calling Python UDFs from rust

bjchambers:


sparrow-py/pysrc/sparrow_py/udf.py
```diff
@@ -0,0 +1,33 @@
...
+FuncType = Callable[..., pd.Series]
+
+class Udf(object):
+    def __init__(self, name, func: FuncType, signature: str) -> None:
+        functools.update_wrapper(self, func)
+        self.name = name
+        self.func = func
+        self.signature = signature
+
+    def run_pyarrow(self, result_type: pa.DataType, *args: pa.Array) -> pa.Array:
```
kerinin, line 16: Is there a use case for changing the result type across calls to `run_pyarrow`? Seems like the result type ought to be a constructor argument.

kerinin (approved):

sparrow-py/pysrc/sparrow_py/udf.py
bjchambers: I don't think so. Consider:

1. User defines a UDF for "add two numbers and return the result".
2. Signature of that is reasonable `add<T: Number>(x: T, y: T) -> T`.
3. They wrap that function once to create the `add` UDF.
4. They apply that to `add(x_i64, y_i64)` (two i64 values) to produce an `i64` and they also call it with `add(x_f64, y_f64)` to produce an `f64`.

Specifically, the `UDF` object represents the UDF, and may be applied within a pipeline multiple times. Each application may have a different type.

PR #561, opened 2023-07-27
feat: Connect Python builder to Rust

bjchambers:
This creates DFG nodes and checks their types.
    
Some extra features:
    
- Uses the pyarrow types of the expression to limit some overloads.
  In the case of `__getattr__` this prevents mistakes leading easily to
  infinite recursion. In the case of `expr[expr]` this lets us use the
  correct methods.
- Allows using python literals (`str`, `int` and `float`) as arguments
  to expressions.
- Renames `ffi` module to `_ffi` to indicate it is private.
- Add some tests fro error cases


sparrow-py/pysrc/sparrow_py/session.py
```diff
@@ -0,0 +1,45 @@
...
+        """Create a new session."""
+        self._session = _ffi.Session()
+        self.tables = []
+
+    def add_table(
+        self,
+        name: str,
+        time_column_name: str,
+        key_column_name: str,
+        schema: pa.Schema,
```
kevinjnguyen, line 31: This is a bit different from our previous usage of `add` data to a table. It's a stronger typed which I like, but what's the requirement to require the schema at table creation time?

bjchambers: 1. Right now, we don't yet have a way to add data, so the only way to test that expressions are properly typed, is to specify a type.
2. Longer term, if creating a table returns an expression, we need to know the type of that expression to verify things like `table1.x` is referencing a valid field.

So we need to do something like (a) require specifying a schema when we create a table or (b) not let you use the table as an expression until you've added data to it. For now, since we can't add any data, it seems like (a) is the easiest.

In general, it seems like we could probably factor that in some to the eventual API. For instance, if we had some notion of "sources" that provided the generator of batches to add to a table but could also provide the schema, then we could do something like the following:

```
source = SlackSource(...)

table = add_table(name='...', schema=source.schema())
# or even
table = add_table(name=...', source)
```

kevinjnguyen (approved):

PR #562, opened 2023-07-27
feat: `index` for DataType::List

jordanrfrazier:
Adds type inference support for lists and adds just the `index` function. 

Further additions: https://github.com/kaskada-ai/kaskada/issues/494


crates/sparrow-arrow/src/downcast.rs
```diff
@@ -56,6 +49,14 @@ pub fn downcast_struct_array(array: &dyn Array) -> anyhow::Result<&StructArray>
         .with_context(|| format!("Unable to downcast {:?} to struct array", array.data_type()))
 }
 
+/// Downcast an `ArrayRef` to a `ListArray`.
+pub fn downcast_list_array(array: &dyn Array) -> anyhow::Result<&ListArray> {
```
bjchambers, line 53: Thought: We probably don't need this method (based on our other discussions around just using the Arrow cast methods). Maybe see if we can get by without it?

jordanrfrazier:
NOTE: Happy to discuss going to `get_map` and `get_list` instead of `get/index` naming

crates/sparrow-compiler/src/types/inference.rs
```diff
@@ -359,7 +381,20 @@ fn instantiate_type(fenl_type: &FenlType, solutions: &HashMap<TypeVariable, Fenl
...
         }
-        FenlType::Collection(Collection::List, _) => todo!("unsupported"),
+        FenlType::Collection(Collection::List, type_vars) => {
+            debug_assert!(type_vars.len() == 1);
+            let concrete_type = solutions
+                .get(&type_vars[0])
+                .cloned()
+                .unwrap_or(FenlType::Concrete(DataType::Null));
+            let field = match concrete_type {
+                // TODO: Should the fields be nullable?
```
bjchambers, line 391: probably. alternatively, it needs to be something we track / propagate.

basically:

1. Non-nullable fields allow certain optimizations. Often, doesn't matter since the kernel looks at the `null_count` anyway, and can use the optimization on an array that is nullable but has no null values or on a non-nullable array.
2. Non-nullable fields cause a panic if a null value is added. I could easily see eventually creating a list from an input that is null... so we should support it.

crates/sparrow-compiler/tests/snapshots/compiler_golden_tests__lookup_same_grouping.snap
```diff
@@ -217,8 +217,11 @@ operations:
         result_type:
           kind:
             List:
-              kind:
-                Primitive: 10
+              name: item
```
bjchambers, line 220: Ugh. All these protos are a pain. This could be another thing we simplify if we just used serde for plans or something.

crates/sparrow-instructions/src/columnar_value.rs
```diff
@@ -62,6 +62,13 @@ impl ColumnarValue {
         ArcRef::new(array).try_map(|a| downcast_map_array(a))
     }
 
+    /// Specialized version of `array_ref` that downcasts the array to a
+    /// list array.
+    pub fn list_array(&self) -> anyhow::Result<ArcRef<dyn Array, ListArray>> {
+        let array = self.array_ref()?;
+        ArcRef::new(array).try_map(|a| downcast_list_array(a))
```
bjchambers, line 69: Instead of `downcast_list_array(a)` can do `a.as_list_opt()` (see https://docs.rs/arrow/latest/arrow/array/trait.AsArray.html#tymethod.as_list_opt)

crates/sparrow-instructions/src/evaluators/list/index.rs
```diff
@@ -0,0 +1,189 @@
...
+        };
+
+        let (index, list) = info.unpack_arguments()?;
+        Ok(Box::new(Self { index, list }))
+    }
+}
+
+impl Evaluator for IndexEvaluator {
+    fn evaluate(&mut self, info: &dyn crate::RuntimeInfo) -> anyhow::Result<ArrayRef> {
+        let list_input = info.value(&self.list)?.list_array()?;
```
bjchambers, line 40: Could also do `.array_ref().as_list()` here, since we know the type is a list.

crates/sparrow-instructions/src/evaluators/list/index.rs
```diff
@@ -0,0 +1,189 @@
...
+            ))
+        }
+        unsupported => {
+            anyhow::bail!("unsupported list type: {:?}", unsupported)
+        }
+    )
+}
+
+/// Generic implementation of `list_indices` for arrays implementing `ArrayAccessor`.
+fn accessible_array_list_indices<T: Send + Sync, A: ArrayAccessor<Item = T>>(
```
bjchambers, line 98: This doesn't need the values, which means it doesn't need to know the type. Specifically, if `value[computed_index]` is null, then `take(values, computed_index)` will be null. So we can *just* compute the index.

With map we needed to handle the case where the *key* was null, but that doesn't happen here. I wouldn't expect list indexing to need to care about the type (or nullness) of values in the list, since it just works with the indices.

bjchambers:
> NOTE: Happy to discuss going to `get_map` and `get_list` instead of `get/index` naming

I'm somewhat fine with either. I suspect if we lean towards the Python side, I'd prefer more specific names (`get_map` and `get_list` and then we can have the Python layer choose the specific name to use for a more user-friendly name), as shown in #561.

crates/sparrow-arrow/src/downcast.rs
jordanrfrazier: Yeah, can do so. Was just going to keep the same pattern until we update them all at once, but might as well do it now and less work later.

crates/sparrow-instructions/src/columnar_value.rs
jordanrfrazier: Removed `ArcRef` and using `as_list()` as per discussion

crates/sparrow-instructions/src/evaluators/list/index.rs
jordanrfrazier: Ahh okay, I thought so but for some reason thought `take` would get the `value` regardless of nullness

crates/sparrow-compiler/src/types/inference.rs
jordanrfrazier: Hm..I'm pretty sure we propagate this now actually. I tested this by changing the nullability and the item name to something other than what is expected in the dataset, and still see the projected schema in the `Scan` set to the dataset's schema:
The `bool_list`'s item was changed to `test` and the nullability to false, which you can see is not here. (This is the correct schema as given by pqrs).
```
2023-07-27T18:40:45.067696Z DEBUG Operation{index=0 operation_label="scan"}: sparrow_runtime::execute::operation: Full operation is ScanOperation { projected_schema: Schema { fields: [Field { name: "time", data_type: Timestamp(Nanosecond, None), nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "subsort", data_type: UInt64, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "key", data_type: UInt64, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "i64_list", data_type: List(Field { name: "item", data_type: Int64, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }), nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "bool_list", data_type: List(Field { name: "item", data_type: Boolean, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }), nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "string_list", data_type: List(Field { name: "item", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }), nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "index", data_type: Int32, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }], metadata: {} }, key_hash_index: KeyHashIndex { key_hash_to_index: "0 entries" }, .. }
```

Schema:
```
  OPTIONAL group bool_list (LIST) {
    REPEATED group list {
      OPTIONAL BOOLEAN item;
    }
  }

```

jordanrfrazier: So this is just a placeholder type that we create during inference, and we verify they match. However..I'm not sure if downstream of this that type is expected rather than the projected schema's type? I haven't seen any issues yet, but perhaps if we have kernels specialized for null and non-null maps/lists we'll see problems if they don't match?

crates/sparrow-arrow/src/downcast.rs
bjchambers: It looks like the method is no longer used -- want to delete it?

crates/sparrow-instructions/src/evaluators/list/index.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/// Gets the indices in the list where the values are at the index within each list.
+fn list_indices(list: &ListArray, indices: &Int64Array) -> anyhow::Result<Int32Array> {
+    let offsets = list.offsets();
+    Ok(accessible_array_list_indices(offsets, indices))
+}
+
+/// Generic implementation of `list_indices` for arrays implementing `ArrayAccessor`.
+fn accessible_array_list_indices(offsets: &OffsetBuffer<i32>, indices: &Int64Array) -> Int32Array {
```
bjchambers, line 60: Change the name and comment -- it no longer uses `ArrayAccesor`. In fact, this can probably just be merged with `list_indices` since all that does is call this method.

bjchambers (approved):

crates/sparrow-arrow/src/downcast.rs
jordanrfrazier: It was used in a few places but moved those to `as_list` as well. Deleted

PR #563, opened 2023-07-27
bug: fix non-generic conversions in string spread 

jordanrfrazier:
Fixes spots in string spread that weren't generic

### DIFF
--- a/crates/sparrow-runtime/src/execute/operation/spread.rs
+++ b/crates/sparrow-runtime/src/execute/operation/spread.rs
@@ -1265,7 +1265,7 @@ where
         let values: &StringArray = downcast_string_array(values.as_ref())?;
         let mut values = values.iter();
 
-        let mut builder = StringBuilder::with_capacity(grouping.len(), 1024);
+        let mut builder = GenericStringBuilder::<O>::with_capacity(grouping.len(), 1024);
         for signal in signal.iter() {
             match signal {
                 Some(true) => builder.append_option(values.next().context("missing value")?),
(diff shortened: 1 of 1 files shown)


crates/sparrow-runtime/src/execute/operation/spread.rs
```diff
@@ -1294,22 +1289,187 @@ where
     }
 }
 
+#[derive(serde::Serialize, serde::Deserialize, Debug)]
+struct UnlatchedLargeStringSpread;
```
bjchambers, line 1293: You shouldn't need all this. The problem is that in the current `*StringSpread` you have this:

https://github.com/kaskada-ai/kaskada/blob/main/crates/sparrow-runtime/src/execute/operation/spread.rs#L1265

```
let values: &StringArray = downcast_string_array(values.as_ref())?;
...
let mut builder = StringBuilder::with_capacity(grouping.len(), 1024);
```

Both of those hard-code the fact that it is operating on a string, ignoring the generic.

Note that:

```
pub type StringBuilder = GenericByteBuilder<GenericStringType<i32>>;
pub type StringArray = GenericByteArray<GenericStringType<i32>>;
```

https://docs.rs/arrow/latest/arrow/array/type.StringBuilder.html

https://docs.rs/arrow/latest/arrow/array/type.StringArray.html

So you replace that orginal block with something that uses the type parameter to choose the right string array and string builder:

```
let values: &GenericByteArray<GenericStringType<O>> = downcast_string_array(values.as_ref())?;
...
let mut builder = GenericByteBuilder::<GenericStringType<O>>::with_capacity(grouping.len(), 1024);
```

Or something like that. There may be some short hands we can use -- for instance

```
type GenericStringArray<O: OffsetSizeTrait> = GenericByteArray<GenericStringType<O>>;
```

for instance... but either way, we just need to *actually* make the implementation generic.

jordanrfrazier: Ah, I did just miss this one spot. the latched implementation is already generic. Okay cool

bjchambers (approved):

PR #565, opened 2023-07-27
feat: sparrow_py initial table management

kevinjnguyen:
# Sparrow_py Initial Table Management

## Changes
* Adds an `add` method to `Table`.
* Adds a `get_data` method to `Table`.
* Adds schema assertion requirements to load data
* Adds naive schema validation at table creation time

### DIFF
--- a/sparrow-py/pysrc/sparrow_py/session.py
+++ b/sparrow-py/pysrc/sparrow_py/session.py
@@ -1,6 +1,7 @@
-from typing import Optional
 from typing import List
+from typing import Optional
 
+import pandas as pd
 import pyarrow as pa
 from sparrow_py import Expr
 from sparrow_py import _ffi
@@ -9,10 +10,66 @@ from sparrow_py import _ffi
 class Table(Expr):
     """ "A table expression."""
 
-    def __init__(self, name: str, ffi: _ffi.Expr):
+    data: List[pa.Table] = []
+
+    @classmethod
+    def validate_table(
+        name: str,
+        time_column_name: str,
+        key_column_name: str,
+        schema: pa.Schema,
+        subsort_column_name: Optional[str] = None,
+        grouping_name: Optional[str] = None,
+    ) -> None:
+        # Checking if the column exist. If they do not, it'll throw a KeyError.
+        key_field = schema.field(key_column_name)
+        if key_field.nullable:
+            raise KeyError(f"Key Column: {key_column_name} must be non-nullable")
+        time_field = schema.field(time_column_name)
+        if time_field.nullable:
+            raise KeyError(f"Time Column: {time_column_name} must be non-nullable")
+
+        if subsort_column_name is not None:
+            subsort_field = schema.field(subsort_column_name)
+            if subsort_field.nullable:
+                raise KeyError(
+                    f"Subsort Column: {subsort_column_name} must be non-nullable"
+                )
+        if grouping_name is not None:
+            grouping_field = schema.field(grouping_name)
+            if grouping_field.nullable:
+                raise KeyError(f"Grouping Column: {grouping_name} must be non-nullable")
+
+    def __init__(
+        self,
+        name: str,
+        time_column_name: str,
+        key_column_name: str,
+        schema: pa.Schema,
+        ffi: _ffi.Expr,
+        subsort_column_name: Optional[str] = None,
+        grouping_name: Optional[str] = None,
+    ):
         """Create a new table expression."""
+        Table.validate_table(
+            time_column_name,
+            key_column_name,
+            schema,
+            subsort_column_name=subsort_column_name,
+            grouping_name=grouping_name,
+        )
         self.name = name
+        self.time_column_name = time_column_name
+        self.key_column_name = key_column_name
+        self.schema = schema
+        self.subsort_column_name = subsort_column_name
+        self.grouping_name = grouping_name
         self._ffi = ffi
+
+    def add(self, df: pd.DataFrame) -> None:
+        """Add a dataframe to a table"""
+        pa_table = pa.Table.from_pandas(df)
+        self.data.append(pa_table)
 
 
 class Session(object):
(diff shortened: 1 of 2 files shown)


sparrow-py/pysrc/sparrow_py/session.py
```diff
@@ -2,17 +2,52 @@
 from sparrow_py import Expr
 from typing import Optional
 import pyarrow as pa
+import pandas as pd
+
+class InvalidSchemaException(Exception):
```
bjchambers, line 7: We should decide if we want to define new exceptions for things, or use existing exceptions. I kind of like having more specific ones, but so far I've been trying to use existing things (eg., `ValueError` or `TypeError`, etc.)

sparrow-py/pysrc/sparrow_py/__init__.py
```diff
@@ -1,7 +1,6 @@
 """Kaskada query builder and local executon engine."""
 from .expr import Expr
-from .session import Table
-from .session import Session
+from .session import Session, Table, InvalidSchemaException
```
bjchambers, line 3: The linter is going to expand these imports again.

sparrow-py/pysrc/sparrow_py/session.py
```diff
@@ -2,17 +2,52 @@
...
         """Create a new table expression."""
         self.name = name
+        self.time_column_name = time_column_name
+        self.key_column_name = key_column_name
+        self.schema = schema.remove_metadata()
+        self.subsort_column_name = subsort_column_name
+        self.grouping_column_name = grouping_column_name
         self._ffi = ffi
+        
+    def add(self, df: pd.DataFrame) -> None:
```
bjchambers, line 35: I think at least some of this is going to need to be implemented on the rust side. Specifically, we'll probably want Rust to expose `add_parquet` and `add_pyarrow` or something like that. Then the Python side can expose a variety of methods taking different types and converting to calls to one or the other of those methods.

sparrow-py/pysrc/sparrow_py/session.py
```diff
@@ -23,23 +58,51 @@ def __init__(self) -> None:
...
         self,
         name: str,
         time_column_name: str,
         key_column_name: str,
         schema: pa.Schema,
         subsort_column_name: Optional[str] = None,
-        grouping_name: Optional[str] = None,
+        grouping_column_name: Optional[str] = None,
     ) -> Table:
+        Session.validate_table(name, time_column_name, key_column_name, schema, subsort_column_name=subsort_column_name, grouping_column_name=grouping_column_name)
```
bjchambers, line 96: I don't think this should be on `Session`. It should be a class method on `Table` or a top-level method. We can discuss.

sparrow-py/pysrc/sparrow_py/session.py
```diff
@@ -23,23 +58,51 @@ def __init__(self) -> None:
...
+            raise KeyError(f"Time Column: {time_column_name} must be non-nullable")
+        
+        if subsort_column_name is not None:
+            subsort_field = schema.field(subsort_column_name)
+            if subsort_field.nullable:
+                raise KeyError(f"Subsort Column: {subsort_column_name} must be non-nullable")
+        if grouping_column_name is not None:
+            grouping_field = schema.field(grouping_column_name)
+            if grouping_field.nullable:
+                raise KeyError(f"Grouping Column: {grouping_column_name} must be non-nullable")
```
bjchambers, line 85: Interesting. I had called field refs for fields that don't exist `AttributeError`, but I can see for tables ti also makes sense to treat it as a `KeyError`. Maybe it makes sense for them to be different in these two contexts ("column in the input" vs. "field in the struct") but maybe worth discussing.

sparrow-py/pytests/session_test.py
```diff
@@ -1,13 +1,100 @@
...
+    for member_id in member_ids:
+        for i in range(0, num_records_per_member):
+            records.append({
+                'time': random.randint(1000000000000000, 9000000000000000), # number of seconds from epoch
+                'key': member_id,
+            })
+    df = pandas.DataFrame(records)
+    return df
+
+def generate_dataset_invalid(num_members, num_records_per_member):
```
bjchambers, line 30: I wonder if these could be a fixture or a class that generates the dataset.

https://docs.pytest.org/en/6.2.x/fixture.html

sparrow-py/pysrc/sparrow_py/session.py
bjchambers: Based on this: https://www.perplexity.ai/search/9d6868a2-8c13-4bae-93ed-a21c5af25af5?s=u

I think we should probably plan to define our own. This looks reasonable for now, although I think we should probably call this an `IncompatibleSchemaError` or something like that (eg., the schema is valid, but isn't compatible with our expectations) and it should end in `Error` like most of the Python exceptions.

In the future, this exception may be defined in Rust and re-thrown, but I'm not sure.

sparrow-py/pysrc/sparrow_py/__init__.py
bjchambers: Run `nox` (no arguments) and make sure all sessions pass. We don't have anything in CI yet.

sparrow-py/pysrc/sparrow_py/session.py
bjchambers: Discussed. We'll do this in a follow-up.

sparrow-py/pysrc/sparrow_py/session.py
bjchambers: I think this may be OK. Alternatively, we could define a separate `UndefinedColumnError` or `UndefinedFieldError` exception for this case.

Specifically, `AttributeError` seems right for `a.b` when `b` doesn't exist, and `KeyError` seems right for `a['b']` when `a` doesn't contain `'b'`.

sparrow-py/pysrc/sparrow_py/session.py
```diff
@@ -1,53 +1,83 @@
...
 
 
 class Table(Expr):
     """ "A table expression."""
-    
+
     data: list[pa.Table] = []
 
-    def __init__(self, 
+    def validate_table(
```
bjchambers, line 19: I think this should be `@classmethod` then (since it's not a valid method on the `Table`).

bjchambers (approved):

PR #566, opened 2023-07-28
feat: Add a Table class on the Rust side

bjchambers:
This moves data management into the Rust codebase. It only counts the number of record batches that have been added.

Next steps after this PR:

1. Preparing (and possibly concatenating) the record batches in-memory.
2. Executing queries over the in-memory batches.


kevinjnguyen (approved):

sparrow-py/src/table.rs
```diff
@@ -0,0 +1,60 @@
...
+
+        let table = Table { name, num_data: 0 };
+        let expr = Expr { node, session };
+        Ok((table, expr))
+    }
+
+    /// Add PyArrow data to the given table.
+    fn add_pyarrow(&mut self, _data: PyArrowType<RecordBatch>) -> PyResult<()> {
+        self.num_data += 1;
+        Ok(())
```
kevinjnguyen, line 58: This is cool to get a record batch here and then we can process it directly without worrying about formats and types. :)

bjchambers: Yep. My next step is working on this to actually connect it to preparation / compilation / execution!

PR #568, opened 2023-07-28
feat: Prepare data in rust

bjchambers:
This breaks some of the Rust code into a new `sparrow-session` crate providing a place for implementing things that aren't specific to the Python FFI.

This also introdoces some pieces for error conversions, with the hope that we can grow to a point where we log errors on the rust side but surface relevant details to the Python side.


crates/sparrow-session/src/table.rs
```diff
@@ -0,0 +1,73 @@
...
+            data: RecordBatch::new_empty(prepared_schema),
+        }
+    }
+
+    pub fn schema(&self) -> SchemaRef {
+        self.preparer.schema()
+    }
+
+    pub fn add_data(&mut self, batch: RecordBatch) -> error_stack::Result<(), Error> {
+        let prepared = self
```
kevinjnguyen, line 48: The `prepare_hash` is hardcoded to 0. In the future, would the prepare_hash be an optional parameter or something the `add_data` would calculate from the source?

sparrow-py/pytests/table_test.py
```diff
@@ -89,11 +89,14 @@ def test_add_dataframe(session, dataset) -> None:
...
     )
     table = Table(session, "table1", "time", "key", schema)
-    assert table._ffi_table.num_data == 0
+    assert table._ffi_table.prepared_data().num_rows == 0
     table.add_data(dataset)
-    assert table._ffi_table.num_data == 1
+    assert table._ffi_table.prepared_data().num_rows == len(dataset)
+
+    prepared = table._ffi_table.prepared_data().to_pandas()
+    assert prepared['_time'].is_monotonic_increasing
```
kevinjnguyen, line 102: We may also want to check the subsort column is also monotonic increasing

kevinjnguyen (approved):

crates/sparrow-runtime/src/prepare/preparer.rs
```diff
@@ -0,0 +1,142 @@
...
+        let sort = |array: &ArrayRef| {
+            arrow::compute::take(array.as_ref(), &indices, None)
+                .into_report()
+                .change_context(Error::SortingBatch)
+        };
+        columns.push(sort(&time)?);
+        columns.push(sort(&subsort)?);
+        columns.push(sort(&key_hash)?);
+
+        // TODO: Slicing?
```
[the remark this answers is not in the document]
bjchambers: I have a prototype of that for the preparing Parquet case. Just need to wrap it up and get it out for review.

sparrow-py/pytests/table_test.py
bjchambers: This is just a place holder until we actually compute. I don't think we want to test the implementation details.

PR #569, opened 2023-07-29
feat: collect to list (non-windowed) (primitive/strings/booleans)

jordanrfrazier:
Adds the `collect` function for non-windowed, primitive/string/boolean types. 

This doesn't support `LargeUtf8` because the other string aggregations that use the `create_typed_evaluator` macro don't support generic offset size, so that will come in a follow up.


crates/sparrow-instructions/src/evaluators/list/collect_primitive.rs
```diff
@@ -0,0 +1,131 @@
...
+        let mut list_builder = ListBuilder::new(builder);
+
+        izip!(entity_indices.values(), input).for_each(|(entity_index, input)| {
+            let entity_index = *entity_index as usize;
+
+            self.token.add_value(self.max, entity_index, input);
+            let cur_list = self.token.state(entity_index);
+
+            list_builder.append_value(cur_list.clone());
+        });
```
jordanrfrazier, line 127: I expect there's a cleverer way to do this by creating the list directly using the offsets and `take`

crates/sparrow-compiler/src/ast_to_dfg.rs
```diff
@@ -527,31 +527,28 @@ pub(super) fn add_to_dfg(
...
-                            )?,
-                            window.location().clone(),
-                        );
-
-                        (null_arg.clone(), null_arg)
-                    }
-                    unexpected => anyhow::bail!("expected window, found {:?}", unexpected),
-                };
+                let (condition, duration) =
+                    flatten_window_args_if_needed(window, dfg, data_context, diagnostics)?;
```
bjchambers, line 531: Thought for the future: One nice thing about the Python route -- I think we can make this a *lot* cleaner. What I'm imagining is that Sparrow exposes more specific functions for kinds of windowed aggregation, and then this flattening / expansion can happen within Python. But -- we can use *Python types* to manage windows and such.

Consider something like the following. This is nice because we don't need to encode the "special" (builder-only) types within Sparrow. 

```
def sum(input: Expr, window: Window = None) -> Expr:
  if window is None:
    return Expr.call("sum", input)
  elif isinstance(window, SinceWindow):
    return Expr.call("sum_windowed", input, window.predicate)
  else:
    pass
```

crates/sparrow-compiler/src/ast_to_dfg.rs
```diff
@@ -527,31 +527,28 @@ pub(super) fn add_to_dfg(
...
 
                 dfg.exit_env();
                 // [agg_input, condition, duration]
                 vec![args[0].clone(), condition, duration]
+            } else if function.name() == "collect" {
+                // The collect function contains a window, but does not follow the same signature
+                // pattern as aggregations, so it requires a different flattening strategy.
+                //
+                // TODO: Flattening the window arguments is hacky and confusing. We should instead
```
bjchambers, line 540: +1 to eliminating the tick. I was also thinking that it would help us avoid generating all the ticks. Consider:

1. We know that we tick every hour
2. We know that the window was *previously* empty
3. Therefore, the new tick doesn't have a different value
4. Therefore, we don't need to do anything for this entity.

Vs.

1. We need to produce a tick for every hour no matter what.

crates/sparrow-compiler/src/ast_to_dfg.rs
```diff
@@ -610,6 +607,37 @@ pub(super) fn add_to_dfg(
     }
 }
 
+#[allow(clippy::type_complexity)]
+fn flatten_window_args_if_needed(
+    window: &Located<Box<ResolvedExpr>>,
+    dfg: &mut Dfg,
+    data_context: &mut DataContext,
+    diagnostics: &mut DiagnosticCollector<'_>,
+) -> anyhow::Result<(Located<Arc<AstDfg>>, Located<Arc<AstDfg>>)> {
```
bjchambers, line 616: Location tracking is *so* fun ;)

crates/sparrow-instructions/src/evaluators/list/collect_primitive.rs
```diff
@@ -0,0 +1,131 @@
...
+        let builder = PrimitiveBuilder::<T>::new();
+        let mut list_builder = ListBuilder::new(builder);
+
+        izip!(entity_indices.values(), input).for_each(|(entity_index, input)| {
+            let entity_index = *entity_index as usize;
+
+            self.token.add_value(self.max, entity_index, input);
+            let cur_list = self.token.state(entity_index);
+
+            list_builder.append_value(cur_list.clone());
```
bjchambers, line 126: Cloning this list is likely inefficient. I think it would be better to do:

```
let (a, b) = self.token.state(entity_index).slices();
list_builder.values().append_values(a);
list_builder.values().append_values(b);

// Q: If there are no values, is the result empty or null? Or does it vary?
list_builder.append(true);
```

crates/sparrow-instructions/src/evaluators/list/collect_string.rs
```diff
@@ -0,0 +1,109 @@
...
+use arrow::array::{ArrayRef, AsArray, ListBuilder, StringBuilder};
+use arrow::datatypes::DataType;
+use itertools::izip;
+use sparrow_arrow::scalar_value::ScalarValue;
+use sparrow_plan::ValueRef;
+use std::sync::Arc;
+
+/// Evaluator for the `collect` instruction.
+///
+/// Collect collects a stream of values into a List<O>. A list is produced
```
bjchambers, line 11: For the comment, I wouldn't say `Collect collects` -- if you want to refer to the thing put it in backticks. Otherwise, Just say what it does -- "Collects a stream of values ..."

crates/sparrow-compiler/src/ast_to_dfg.rs
jordanrfrazier: Yeah, that looks like it helps with managing window type sin the python side. The other side of things is how the window functions like `daily()` need to know the entity domain from the `input`. If we make it such that functions have windows built-in such that they're inherently part of the same domain then that would get rid of the `env` binding we have to do in `ast_to_dfg`.

crates/sparrow-compiler/src/ast_to_dfg.rs
jordanrfrazier: Though, we still want to produce a row for that entity at each hour even if the value is empty.

crates/sparrow-instructions/src/evaluators/list/collect_primitive.rs
jordanrfrazier: > // Q: If there are no values, is the result empty or null? Or does it vary?

I had this same question written out here, but decided on empty. My reasoning was that it more clearly indicates that we've attempted to collect values into this `list`, but there were no values, hence the size of the list is 0. As a follow-up, having the `size_of(list) or `len(list)` function will be nice.

jordanrfrazier: Side discussion on "should collect collect `null` (it currently does)

Maybe it should not though. `count(E) == len(collect(E))` -- If this should be true, then `collect` should "skip" adding nulls to the list, but still produce an empty list if the current buffer is empty. `count` is special in that it produces `0` by default if no value exists, which would keep the `len(collect([null, null])) == 0` invariant in line.

crates/sparrow-instructions/src/evaluators/list/collect_string.rs
```diff
@@ -101,7 +101,7 @@ impl CollectStringEvaluator {
                 .add_value(self.max, entity_index, input.map(|s| s.to_owned()));
             let cur_list = self.token.state(entity_index);
 
-            list_builder.append_value(cur_list.clone());
+            list_builder.append_value(cur_list.iter().copied());
```
bjchambers, line 104: You can't copy the strings, so you need `cloned` for that case.

bjchambers (approved):

crates/sparrow-instructions/src/evaluators/list/collect_string.rs
jordanrfrazier: Oops, clearly didn't test this one.

PR #572, opened 2023-07-31
feat: Allow writing to a channel

bjchambers:
This introduces a `Destination` enum and converts the existing protobuf to that before execution. This allows providing an output channel for direct output (eg., to Python via PyArrow).

This also removes the deprecated Redis output.


crates/sparrow-runtime/src/execute/output.rs
```diff
@@ -94,38 +124,39 @@ pub(super) fn write(
...
+    channel: tokio::sync::mpsc::Sender<RecordBatch>,
+    _progress_updates_tx: tokio::sync::mpsc::Sender<ProgressUpdate>,
+) -> error_stack::Result<(), Error> {
+    while let Some(next) = batches.next().await {
+        channel
+            .send(next)
+            .await
+            .map_err(|_e| error_stack::report!(Error::WritingToDestination("channel")))?;
+
+        // progress_updates_tx.send(ProgressUpdate::Output { num_rows })
```
jordanrfrazier, line 154: I believe this should happen here?

crates/sparrow-runtime/src/execute/output.rs
```diff
@@ -40,13 +37,49 @@ pub enum Error {
...
+/// The output destination.
+///
+/// TODO: Replace the protobuf destinations with pure Rust structs.
+#[derive(Debug)]
+pub enum Destination {
+    ObjectStore(ObjectStoreDestination),
+    #[cfg(feature = "pulsar")]
+    Pulsar(PulsarDestination),
+    Channel(tokio::sync::mpsc::Sender<RecordBatch>),
+}
```
jordanrfrazier, line 49: This is nice. Can pretend this is the start of our proto -> rust type layer :))

jordanrfrazier (approved):

crates/sparrow-runtime/src/execute/output.rs
bjchambers: It should -- I'm leaving it commented out because I'm not 100% sure it will work (eg., it could fill the progress channel if nobody listens). Will revisit once we have more hooked up.

crates/sparrow-runtime/src/execute/output.rs
bjchambers: Yeah. Although, I think it would be better if this was extensible. Eg.

```
trait Destination {
  fn write(stream: BoxStream<'static, RecordBatch>) -> ...
  // or
  fn write(batch: RecordBatch) -> ...
}

struct Pulsar { ... }
impl Destination for Pulsar { ... }
```

Ideally, that would let us (a) add destinations without extending the enum, allowing user-defined destinations and (b) let us put the plusar deps in a later crate in the compilation tree, tec.

PR #574, opened 2023-07-31
feat: support windows in collect aggregations

jordanrfrazier:
TODO:
* Add sliding window aggregations

Also fixes bug where since windows were...never used. Because we used a duration of `1`, it technically was using the two stacks implementation with a single active window. Likely, this is less efficient.


jordanrfrazier:
I've forgotten the state of this. Might just close for now and reopen if we need to.

PR #575, opened 2023-07-31
feat: Execute Python queries

bjchambers:
This removes the ability to inspect the prepared data from the Python FFI. Instead, the Python FFI now has an `execute` method allowing you to inspect the results of an expression, including the bare table.

This changes the execute methods a bit to have a simpler API to interface with.

This relies on pushing "in-memory" data into the `DataContext`, which is very brittle and hacky. For example, it currently only supports adding data once, and requires that the schemas line up.


crates/sparrow-compiler/src/data_context.rs
```diff
@@ -60,6 +61,10 @@ impl DataContext {
         self.table_info.get(&id)
     }
 
+    pub fn table_info_mut(&mut self, id: TableId) -> Option<&mut TableInfo> {
```
bjchambers, line 64: Previously, DataContext was only used for compiling queries. Now, it's the easiest place to collect the data that we're going to run on (we don't have any way of plumbing "here is some data") today -- we just plumb "here are the files". This should change if we continue down this route.

crates/sparrow-session/src/session.rs
```diff
@@ -181,6 +188,87 @@ impl Session {
             Ok(result)
         }
     }
+
+    pub fn execute(&self, expr: &Expr) -> error_stack::Result<RecordBatch, Error> {
```
bjchambers, line 192: This is the main entry point -- a pure-rust execution mechanism on the `sparrow-session` crate. The Rust/Python layer just wraps this and exposes it to Python (and manages type conversions).

sparrow-py/pytests/math_test.py
```diff
@@ -0,0 +1,69 @@
...
+    ]), dtype = {
+        'm': 'Int64',
+        'n': 'Int64',
+    })
+    # parse_dates = ['time'],)
+
+    # TODO: Options for outputting to different destinations (eg., to CSV).
+    # TODO: Allow running a single expression (eg., without field names)
+
+    #result = (table_int64.m + tableint64.n).execute()
```
bjchambers, line 51: I'd like to run this, but currently can't since I can't make a record nor can I execute a single column. Will fix in a later PR. For now, demonstrating that we can run a query for just the table.

sparrow-py/pytests/math_test.py
```diff
@@ -0,0 +1,69 @@
...
+            pa.field("time", pa.string(), nullable=False),
+            pa.field("key", pa.string(), nullable=False),
+            pa.field("m", pa.int64()),
+            pa.field("n", pa.int64()),
+        ]
+    )
+    return Table(session, "table1", "time", "key", schema)
+
+def test_read_table(session, table_int64) -> None:
+    table_int64.add_csv('\n'.join([
```
bjchambers, line 34: In a follow-up, I need to figure out how to cleanly write these tests -- I may look at Pandas and Polars and see what they do. Over time, we'll likely have the equivalent of our e2e tests written directly in Python.

sparrow-py/pytests/table_test.py
```diff
@@ -94,9 +94,9 @@ def test_add_dataframe(session, dataset) -> None:
         ]
     )
     table = Table(session, "table1", "time", "key", schema)
-    assert table._ffi_table.prepared_data().num_rows == 0
+    # assert table._ffi_table.prepared_data().num_rows == 0
```
bjchambers, line 97: Can remove these commented out lines.

jordanrfrazier:
Approving to unblock since much will likely change anyway, will take a closer look in a couple hours

jordanrfrazier (approved):

PR #580, opened 2023-08-01
Update example notebook to align to recent discussions

kerinin:


examples/slackbot/Notebook.ipynb
```diff
@@ -27,89 +27,93 @@
...
-   "metadata": {},
-   "outputs": [],
+   "metadata": {
+    "collapsed": true,
+    "jupyter": {
+     "outputs_hidden": true
+    },
+    "tags": []
+   },
+   "outputs": [
```
bjchambers, line 39: Strip output, since it's not particularly meaningful?

examples/slackbot/Notebook.ipynb
```diff
@@ -27,89 +27,93 @@
...
-    "#   \"ts\": \"1355517523.000005\" \n",
-    "# }\n",
+    "def prompt(messages):\n",
+    "    last_message = messages.lag(1)\n",
+    "    since_last_message = messages.time_of().seconds_between(last_message.time_of())\n",
+    "    conversation_start = since_last_message > k.minutes(5)\n",
     "\n",
-    "# Create a new local session\n",
-    "s = k.session()\n",
+    "    k.make_record( \n",
```
bjchambers, line 63: I believe this suggests that `make_record` is a top level function. Why not just `record`?

```
def record(**kwargs: Expr) -> Expr:
  pass
```

Should work. It will require that the arg names be valid keywords, which doesn't seem unreasonable.

examples/slackbot/Notebook.ipynb
```diff
@@ -27,89 +27,93 @@
...
-    "# { \t\n",
-    "#   \"type\": \"message\", \t\n",
-    "#   \"channel\": \"C123ABC456\", \t\n",
-    "#   \"user\": \"U123ABC456\", \t\n",
-    "#   \"text\": \"Hello world\", \t\n",
-    "#   \"ts\": \"1355517523.000005\" \n",
-    "# }\n",
+    "def prompt(messages):\n",
+    "    last_message = messages.lag(1)\n",
+    "    since_last_message = messages.time_of().seconds_between(last_message.time_of())\n",
```
bjchambers, line 60: Feels weird to call this seconds between with a left-hand side. Not sure though.

bjchambers, line 60: `time()` seems more reasonable than `time_of()` now.

examples/slackbot/Notebook.ipynb
```diff
@@ -27,89 +27,93 @@
...
+    "    conversation_start = since_last_message > k.minutes(5)\n",
     "\n",
-    "# Create a new local session\n",
-    "s = k.session()\n",
+    "    k.make_record( \n",
+    "        # A list of all messages over the past 10 minutes (up to 100)\n",
+    "        recent_messages=messages\n",
+    "            .select(\"user\", \"type\", \"text\")\n",
```
bjchambers, line 66: This could get confusing since `select` often means "filter rows". Maybe `select fields`? Although, I think `polars` call it `select`. Maybe that's OK.

examples/slackbot/Notebook.ipynb
```diff
@@ -27,89 +27,93 @@
...
-    "# -> https://api.slack.com/events/message \n",
-    "# Example Event:\n",
-    "# { \t\n",
-    "#   \"type\": \"message\", \t\n",
-    "#   \"channel\": \"C123ABC456\", \t\n",
-    "#   \"user\": \"U123ABC456\", \t\n",
-    "#   \"text\": \"Hello world\", \t\n",
-    "#   \"ts\": \"1355517523.000005\" \n",
-    "# }\n",
+    "def prompt(messages):\n",
```
bjchambers, line 58: What is the type of messages here?

examples/slackbot/Notebook.ipynb
```diff
@@ -27,89 +27,93 @@
...
+    "            .when(messages.col(\"reactions\").is_not_null())\n",
+    "            .count(window=since(conversation_start)),\n",
+    "    )\n",
     "\n",
-    "# Load messages into Kaskada\n",
-    "# For training: message = k.read_parquet(..., entity=\"user_id\")\n",
-    "message = s.read_pulsar(..., entity=\"user_id\")\n",
+    "def examples(messages):\n",
+    "    k.make_record(\n",
+    "        prompt=prompt(messages).lag(1).lookup(messages.get(\"channel\")),\n",
```
bjchambers, line 77: Is this suggesting we need lookup?

examples/slackbot/Notebook.ipynb
```diff
@@ -27,89 +27,93 @@
...
     "\n",
-    "# Each example will predict who will engage next, given\n",
-    "# the last 10 minutes of conversation history\n",
-    "examples = k.make_record(\n",
-    "  prompt=prompt.lag(1).lookup(message.get(\"channel\")),\n",
-    "  completion=completion,\n",
-    ")\n",
-    "examples = examples.when(examples.get(\"prompt\").is_not_null())\n",
+    "    tl = examples(messages = k.ParquetSource(files=[\"./messages.parquet\"]))\n",
+    "    tl = tl.when(tl.col(\"prompt\").is_valid())\n",
```
bjchambers, line 88: Suggest `filter` instead of `when`, given how it normally works.

bjchambers, line 88: `tl.col("prompt")` -- I thought we settled on `tl["prompt]"` as the one way to do this?

examples/slackbot/Notebook.ipynb
```diff
@@ -27,89 +27,93 @@
...
+    "    k.init_session()\n",
     "\n",
-    "# Each example will predict who will engage next, given\n",
-    "# the last 10 minutes of conversation history\n",
-    "examples = k.make_record(\n",
-    "  prompt=prompt.lag(1).lookup(message.get(\"channel\")),\n",
-    "  completion=completion,\n",
-    ")\n",
-    "examples = examples.when(examples.get(\"prompt\").is_not_null())\n",
+    "    tl = examples(messages = k.ParquetSource(files=[\"./messages.parquet\"]))\n",
```
bjchambers, line 87: Considering putting the sources in a separate module. This would likely be:

```
from sparrow_py.sources import ParquetSource

...
```

(not sure if there is a way to import multiple modules as the same name)

examples/slackbot/Notebook.ipynb
```diff
@@ -27,89 +27,93 @@
...
-    "    messages=[{\"role\": \"user\", \"content\": p}],\n",
-    "    n=5, logprobs=5,\n",
-    "  )\n",
-    "  for completion in completions:\n",
-    "    if sum(completion.logprobs) > 0.05:\n",
-    "      user = completion.completion\n",
-    "      # Use the Slack API to PM the user?\n",
-    "      print(f\"Notify {user} of conversation: \\n{p}\") "
+    "    # Handle messages in realtime\n",
+    "    for p in prompt(live_messages).run(starting=datetime.now()).to_generator():\n",
```
bjchambers, line 106: Still TBD how feasible some of these in a short time frame.

examples/slackbot/Notebook.ipynb
bjchambers: Alternative:

```
def record(fields: dict<str, Expr>) -> Epr:
  pass
```

Would allow:

```
record({
  "x": ...,
  "y": ...,
})
```

Potential benefit to the field names being strings, especially if you want to compute them.

examples/slackbot/Notebook.ipynb
kerinin: Not sure - `Expr`? For clarity in the demo I'm not sure if it's worth using type annotations, unless it helps communicate what's happening.

examples/slackbot/Notebook.ipynb
kerinin: True - let's say `seconds_since`, which is also nice in that it's more clear how the sign will work.

examples/slackbot/Notebook.ipynb
kerinin: Using the dict form as discussed yesterday

examples/slackbot/Notebook.ipynb
kerinin: old code, removing

examples/slackbot/Notebook.ipynb
kerinin: Did some brief googling about Python idioms - it seems there's some preference for function calls over class constructors in API's (this aligns with Pandas, Polars, NumPy, etc):

* https://stackoverflow.com/questions/64487903/why-are-classes-mostly-instantiated-through-functions
* https://devblogs.microsoft.com/python/idiomatic-python-functions-versus-classes/

I agree it makes sense to put this in a module, I also think it will be preferable to put the constructor behind a factory function.

examples/slackbot/Notebook.ipynb
kerinin: Yup

kerinin: Agreed

examples/slackbot/Notebook.ipynb
kerinin: It's going to be hard to demo this if we can't do something in real time. Let's keep an eye on this - shortcuts / simplifications are fine.

examples/slackbot/Notebook.ipynb
bjchambers: Interesting. I wonder why they don't use classmethods. I like using methods so you don't need to conflate everything in the signature for the constructor, but if you make it a class method then it reads like: `ParquetSource.from_files(...)`, so it let's you group the factory methods for a class within the name of the class.

bjchambers:
How does this relate to #591?

epinzur:
should we close this PR?

PR #581, opened 2023-08-01
ref: Cleanup sources in Python

bjchambers:
Separate the different kinds of sources, so we have an `ArrowSource` for PyArrow and Pandas based data. Default to taking an initial batch in the constructor, so the schema can be inferred.

### DIFF
--- a/sparrow-py/pysrc/sparrow_py/sources/arrow.py
+++ b/sparrow-py/pysrc/sparrow_py/sources/arrow.py
@@ -1,0 +1,59 @@
+"""Provide sources based on PyArrow, including Pandas and CSV."""
+from io import StringIO
+from typing import Optional
+from typing import Union
+
+import pandas as pd
+import pyarrow as pa
+
+from .source import Source
+
+
+class ArrowSource(Source):
+    """A source reading data batches from Arrow or Pandas."""
+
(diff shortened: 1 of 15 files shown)


sparrow-py/pysrc/sparrow_py/sources/arrow.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class ArrowSource(Source):
+    """A source reading data batches from Arrow or Pandas."""
+
+    def __init__(
+        self,
+        time_column_name: str,
+        key_column_name: str,
+        data: Union[pd.DataFrame, pa.RecordBatch],
```
kerinin, line 19: Probably a good idea to subclass this eventually, for people who aren't familiar with Arrow and are looking for how to get their Pandas data in.

kerinin (approved):

sparrow-py/pysrc/sparrow_py/sources/arrow.py
bjchambers: Makes sense. I was debating that while writing this PR.

Looking around, I actually wonder if this should be `_arrow.py` since people should import from `sparrow_py.sources`.

PR #586, opened 2023-08-02
feat: Support aggregation functions in Python

bjchambers:
Introduces Python classes for representing windows and flattens them into arguments for the corresponding aggregation functions.

This also changes the AST to DFG logic so that it only tries to flatten if the original AST is available. This does mean that ticks won't be recreated, meaning they won't have the proper domain.

Both of these likely need to be extended to cover other aggregations.

### DIFF
--- a/crates/sparrow-compiler/src/ast_to_dfg.rs
+++ b/crates/sparrow-compiler/src/ast_to_dfg.rs
@@ -99,7 +99,7 @@ pub fn add_to_dfg(
     diagnostics: &mut DiagnosticCollector<'_>,
     op: &ExprOp,
     mut arguments: Resolved<Located<AstDfgRef>>,
-    expr: Option<&ResolvedExpr>,
+    original_ast: Option<&ResolvedExpr>,
 ) -> anyhow::Result<AstDfgRef> {
     let argument_types = arguments.transform(|i| i.with_value(i.value_type().clone()));
 
@@ -517,15 +517,28 @@ pub fn add_to_dfg(
                 .try_collect()?;
 
             let args: Vec<_> = if function.is_aggregation() {
-                // If the function is an aggregation, we may need to flatten the window.
-                dfg.enter_env();
-                dfg.bind("$condition_input", args[0].inner().clone());
+                let window_arg = original_ast.map(|e| &e.args()[1]);
+                let (condition, duration) = match window_arg {
+                    Some(window) => {
+                        // If the function is an aggregation, we may need to flatten the window.
+                        dfg.enter_env();
+                        dfg.bind("$condition_input", args[0].inner().clone());
 
-                let window = &expr.unwrap().args()[1];
-                let (condition, duration) =
-                    flatten_window_args_if_needed(window, dfg, data_context, diagnostics)?;
+                        let result =
+                            flatten_window_args_if_needed(window, dfg, data_context, diagnostics)?;
+                        dfg.exit_env();
+                        result
+                    }
+                    None => {
+                        // If `expr` is None, we're running the Python builder code,
+                        // which already flattened things.
+                        //
+                        // Note that this won't define the `condition_input` for the
+                        // purposes of ticks.
+                        (args[1].clone(), args[2].clone())
+                    }
+                };
 
-                dfg.exit_env();
                 // [agg_input, condition, duration]
                 vec![args[0].clone(), condition, duration]
             } else if function.name() == "collect" {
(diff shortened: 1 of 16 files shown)


crates/sparrow-compiler/src/functions/aggregation.rs
```diff
@@ -39,7 +39,7 @@ pub(super) fn register(registry: &mut Registry) {
     registry
         .register("sum<N: number>(input: N, window: window = null) -> N ")
         .with_dfg_signature(
-            "sum<N: number>(input: N, window: window = null, duration: i64 = null) -> N",
+            "sum<N: number>(input: N, window: bool = null, duration: i64 = null) -> N",
```
jordanrfrazier, line 42: Looking into this. This will cause sparrow tests to fail.

jordanrfrazier: This type here is a`window` (and not `boolean`) for type-checking the dfg. The dfg checks that the window argument to the signature is in fact a window, and not just a boolean. After if verifies that and instantiates the types, it flattens the window argument. Additionally, the dfg signature in the registry needs to have the duration argument, so it can include the duration within the pattern.
```
        .with_dfg_signature("sum<N: number>(input: N, window: window = null, duration: i64 = null) -> N")
        .with_implementation(Implementation::new_pattern(&format!(
            "(sum ({}) ({}) ({}))",
            "transform (if ?input_is_new ?input_value) (merge_join ?input_op ?window_op)",
            "?window_value",
            "?duration_value"
```

crates/sparrow-compiler/src/ast_to_dfg.rs
```diff
@@ -517,15 +517,28 @@ pub fn add_to_dfg(
...
+                        dfg.exit_env();
+                        result
+                    }
+                    None => {
+                        // If `expr` is None, we're running the Python builder code,
+                        // which already flattened things.
+                        //
+                        // Note that this won't define the `condition_input` for the
+                        // purposes of ticks.
+                        (args[1].clone(), args[2].clone())
```
jordanrfrazier, line 538: The difference here is that `args[1]` is already a boolean, not a window. That's why the change to the `dfg_signature` works with the python test, and not sparrow. 

Which is somewhat interesting -- with some refactoring to the `ast_to_dfg`, we could probably support `window=daily()` instead of `window=since(daily())`.

jordanrfrazier: This should work after fixing my signature bugs in https://github.com/kaskada-ai/kaskada/pull/587

crates/sparrow-compiler/src/functions/aggregation.rs
jordanrfrazier: This should work after fixing my signature bugs in https://github.com/kaskada-ai/kaskada/pull/587

sparrow-py/pysrc/sparrow_py/expr.py
```diff
@@ -285,3 +299,12 @@ def run_to_csv_string(self) -> str:
         """Execute the expression and return the CSV as a string."""
         # Convert PyArrow -> Pandas -> CSV, omitting the index (row).
         return self.run().to_csv(index=False)
+
+def _aggregation(op: str, input: Expr, window: Optional[Window]) -> Expr:
+    """Helper for creating an aggregation."""
+    if window is None:
+       return Expr.call(op, input, None, None)
+    elif isinstance(window, SinceWindow):
+       return Expr.call(op, input, window._predicate, 1)
```
jordanrfrazier, line 308: Can you change the `1` to `None`? This will ensure we actually use the `since` window implementations, which should be more performant than the TwoStacks. (Also, collect only supports `since`)

Ideally we pass the type of the window rather than relying on "number of non-null args" to determine window type..but, future change.

sparrow-py/src/expr.rs
```diff
@@ -31,9 +36,16 @@ impl Expr {
...
-                arg.into_ast_dfg_ref(&mut rust_session)
-                    // DO NOT SUBMIT: Better error handling.
-                    .map_err(|_| PyRuntimeError::new_err("unable to create literal"))
+                match arg {
+                    None => rust_session
+                        .add_literal(Literal::Null)
+                        .map_err(|_| PyRuntimeError::new_err("unable to create null literal")),
+                    Some(arg) => {
+                        arg.into_ast_dfg_ref(&mut rust_session)
+                            // DO NOT SUBMIT: Better error handling.
```
jordanrfrazier, line 45: blocker?

jordanrfrazier (approved):

sparrow-py/src/expr.rs
bjchambers: not really -- definitely need to really improve the error story, but TBD

PR #589, opened 2023-08-02
test: Introduce a golden test helper for Python

bjchambers:


sparrow-py/pytests/conftest.py
```diff
@@ -2,8 +2,66 @@
...
     """Automatically initialize the session for this PyTest session."""
     init_session()
+
+
+def pytest_addoption(parser: pytest.Parser):
+    parser.addoption("--save-golden", action="store_true", help="update golden files")
+
+@pytest.fixture
+def golden(request: pytest.FixtureRequest, pytestconfig: pytest.Config):
+    output = 1
```
bjchambers, line 27: I use the index so we can have multiple golden checks in a test.

I could omit the `_1` if it is the first iteration, which would simplify some of the output files.

sparrow-py/pytests/conftest.py
```diff
@@ -2,8 +2,66 @@
...
+                df.to_csv(filename, index=False)
+            elif format == "parquet":
+                df.to_parquet(filename)
+            else:
+                raise ValueError(f"Unknown format {format}")
+
+        if format == "csv":
+            dtypes = {}
+            parse_dates = []
+            for name, dtype in df.dtypes.to_dict().items():
```
bjchambers, line 56: Some trickier to configure the "read CSV" so that datatypes round trip (parse out date times, etc.)

jordanrfrazier: Yeah, seems fair to use this.

jordanrfrazier (approved):

PR #595, opened 2023-08-03
feat: sparrow py collect function

jordanrfrazier:
support `collect` function in sparrow-py


sparrow-py/pysrc/sparrow_py/expr.py
```diff
@@ -287,6 +287,9 @@ def is_not_null(self) -> "Expr":
         """Return a boolean expression indicating if the expression is not null."""
         return Expr.call("is_valid", self)
 
+    def collect(self, max: Optional[int] = None, window: Optional[Window] = None) -> "Expr":
```
bjchambers, line 290: Do we want to default to "unbounded" collection, or require you specify? I think I'd prefer `max: Optional[int]` which forces the user to specify `max = None` if they want unbounded collection.

sparrow-py/pysrc/sparrow_py/expr.py
```diff
@@ -287,6 +287,9 @@ def is_not_null(self) -> "Expr":
...
+        return _collect(self, max, window)
```
bjchambers, line 291: I think we can generalize the helper, so this could be `_aggregation("collect", self, window, max)`.

sparrow-py/pysrc/sparrow_py/expr.py
```diff
@@ -314,8 +317,19 @@ def _aggregation(op: str, input: Expr, window: Optional[Window]) -> Expr:
...
     elif isinstance(window, SinceWindow):
-        return Expr.call(op, input, window._predicate, 1)
+        return Expr.call(op, input, window._predicate, None)
     elif isinstance(window, SlidingWindow):
         return Expr.call(op, input, window._predicate, window._duration)
     else:
         raise ValueError(f"Unknown window type {window!r}")
+
+def _collect(input: Expr, max: Optional[int], window: Optional[Window]) -> Expr:
+    """Helper for creating collect expr"""
```
bjchambers, line 327: Run `nox` to make sure this passes in all lints, etc. I suspect you're going to get an error saying the first sentence needs to be imperative (describe what the function does). Eg. "Create a collect expression" vs. "Helper for ..."

sparrow-py/pysrc/sparrow_py/expr.py
```diff
@@ -314,8 +317,19 @@ def _aggregation(op: str, input: Expr, window: Optional[Window]) -> Expr:
...
         return Expr.call(op, input, None, None)
     elif isinstance(window, SinceWindow):
-        return Expr.call(op, input, window._predicate, 1)
+        return Expr.call(op, input, window._predicate, None)
     elif isinstance(window, SlidingWindow):
         return Expr.call(op, input, window._predicate, window._duration)
     else:
         raise ValueError(f"Unknown window type {window!r}")
+
+def _collect(input: Expr, max: Optional[int], window: Optional[Window]) -> Expr:
```
bjchambers, line 326: Can generalize the helper. It would be easiest if the `max` argument came *after* the window, but we can make it work:

```
def _aggregation(op: str, input: Expr, window: Optional[Window], *args: Optional[Arg]) -> Expr:
  """
  Create an aggregation.

  Parameters
  ----------
  op : str
      The operation to create.
  input : Expr
      The input to the expression.
  window : Optional[Window]
    The window to use for the aggregation.
  *args : Optional[Arg]
      Additional arguments to provide before `input` and the flattened window.

  Returns
  -------
  The resulting expression.
  """
  if window is None:
    # Note: things would be easier if we had a more normal order, which
    # we could do as part of "aligning" Sparrow signatures to the new direction.
    return Expr.call(op, *args, input, None, None)
  ...

File: sparrow-py/pytests/collect_test.py [hunk 5]
```diff
@@ -0,0 +1,66 @@
...
+def test_collect_since_window(source, golden) -> None:
+    """Test we can collect values to a list in a since window"""
+    m = source["m"]
+    golden(record(
+        {
+            "m": m,
+            "since_m": m.sum(window=SinceWindow(m > 10))
+        }
+    ))
+
```
bjchambers, line 59: Trailing white space makes me think this needs to have `nox -s fix_lint` run on it (or manually run `black`).

sparrow-py/pytests/conftest.py
```diff
@@ -27,9 +27,21 @@ def golden(request: pytest.FixtureRequest, pytestconfig: pytest.Config):
...
     def handler(
         query: sparrow_py.Expr,
-        format: Union[Literal["csv"], Literal["parquet"]] = "csv",
+        format: Union[Literal["csv"], Literal["parquet"], Literal["json"]] = "json",
     ):
-        """Check query results against a golden file."""
+        """
+        Check query results against a golden file.
+
+
```
bjchambers, line 35: Too much white space. I think this should be just one line before `Parameters`.

sparrow-py/pysrc/sparrow_py/expr.py
jordanrfrazier: Agreed -- that's how it should've been. Fixed

sparrow-py/pysrc/sparrow_py/expr.py
jordanrfrazier: I intentionally didn't to not associate collect with windows, but I can maybe rename to a helper?

jordanrfrazier: Actually -- with the `max` parameter, wouldn't that add an additional argument to the aggregation calls?

bjchambers (approved):

PR #596, opened 2023-08-03
feat: Expose execution control from Rust to Python

bjchambers:
This is a building block for allowing Python code to expose batches as generators, async generators, etc.


sparrow-py/src/execution.rs
```diff
@@ -0,0 +1,31 @@
...
+
+impl Execution {
+    pub(crate) fn new(execution: RustExecution) -> Self {
+        Self(Some(execution))
+    }
+}
+
+#[pymethods]
+impl Execution {
+    fn collect_pyarrow(&mut self, py: Python<'_>) -> Result<Vec<PyObject>, Error> {
```
kerinin, line 19: Could the be confused with `Expr.collect()`?

crates/sparrow-session/src/execution.rs
```diff
@@ -0,0 +1,104 @@
...
+            }
+        }
+    }
+
+    pub async fn next(&mut self) -> error_stack::Result<Option<RecordBatch>, Error> {
+        self.is_done()?;
+        Ok(self.output.next().await)
+    }
+
+    pub async fn next_blocking(&mut self) -> error_stack::Result<Option<RecordBatch>, Error> {
```
kerinin, line 78: `collect_all_blocking` isn't `async` but this is - intentional?

kerinin (approved):
So I guess the eventual idea is to wrap the `async` rust functions with async Python functions?

sparrow-py/src/execution.rs
bjchambers: I don't think so? This is part of `sparrow_py._ffi.Execution`, while that is a method on `sparrow-py.expr.Expr`, and has a different name (`collect` vs. `collect_pyarrow`).

crates/sparrow-session/src/execution.rs
bjchambers: Oh. This should be non-async (it uses the tokio `runtime` directly, and waits for that).

bjchambers:
> So I guess the eventual idea is to wrap the `async` rust functions with async Python functions?

Yeah. I'm not sure how that will work -- I think we need to have the rust runtime (tokio) run the async function and then call a method to resolve the Python future on it's pool. But not 100% sure (exploring next).

sparrow-py/src/execution.rs
kerinin: It's not a big deal, but in one case we're describing an operation and in another, we're controlling execution. I could imagine the execution class moving around or becoming more "public". Not a big deal - I was just wondering if there's a similar word we could use.

bjchambers: Ah. I think it's pretty common for the "collect the results of this thing"... I think we're probably OK -- one being called `collect` and the other `collect_pyarrow` or `collect_dataframes` etc. In terms of the user, it's about whether they've already run or not:

```
collect().run()
vs.
run().collect_dataframes()
```

PR #598, opened 2023-08-03
feat: More work on Python API

bjchambers:
- Introduce options for limits, and the method `show()` for displaying the results of a query. Will use `ipython.display()` if available, or print the Pandas dataframe to stdout otherwise.
- Added pydoc for collect
- Fixed some issues in the pytest for adding data (specifically around how the `_time` column round-tripped through the JSON in the golden test).
- Updated some of the modules to allow `import sparrow_py as kt` and using `kt.sources.CsvSource`. The trick was to make sure the root module (`sparrow_py` currently) re-exports the submodule (`sources`).


sparrow-py/pysrc/sparrow_py/_expr.py
```diff
@@ -11,14 +11,16 @@
 
 import pandas as pd
 import pyarrow as pa
-import sparrow_py
+import sparrow_py as kt
```
jordanrfrazier, line 14: why `kt`?

sparrow-py/pysrc/sparrow_py/_expr.py
```diff
@@ -288,8 +290,9 @@ def is_not_null(self) -> "Expr":
         return Expr.call("is_valid", self)
 
     def collect(
-        self, max: Optional[int] = None, window: Optional[Window] = None
+        self, max: Optional[int], window: Optional[Window] = None
```
jordanrfrazier, line 293: could've sworn I had this. Maybe I stashed a change I meant to add

sparrow-py/pysrc/sparrow_py/_expr.py
```diff
@@ -345,7 +367,7 @@ def _aggregation(
     # us to add the *args like so.
     if window is None:
         return Expr.call(op, *args, input, None, None)
-    elif isinstance(window, SinceWindow):
+    elif isinstance(window, kt.SinceWindow):
```
jordanrfrazier, line 370: do you have to do `kt.` since you've already imported it
```
from ._windows import SinceWindow
```
if so, or not, should we do the same for `SlidingWindow` below?

sparrow-py/pysrc/sparrow_py/_ffi.pyi
```diff
@@ -8,6 +8,9 @@ from sparrow_py.udf import Udf
 class Session:
     def __init__(self) -> None: ...
 
+class ExecutionOptions(object):
+    def __init__(self, row_limit: Optional[int] = None) -> None: ...
+    row_limit: Optional[int]
```
jordanrfrazier, line 13: I don't know python, but chatgpt said this looks a lot like a data class and can be done (in 3.7 >)
```
from dataclasses import dataclass

@dataclass
class ExecutionOptions:
    row_limit: Optional[int] = None
```

sparrow-py/pytests/conftest.py
```diff
@@ -83,10 +83,14 @@ def handler(
...
             correct = pd.read_json(
-                filename, orient="records", lines=True, dtype=df.dtypes.to_dict()
+                filename, orient="records", lines=True, dtype=df.dtypes.to_dict(),
+                date_unit = 'ns',
             )
         else:
             raise ValueError(f"Unknown format {format}")
+        print(df.iloc[:, 0])
+        print(df.dtypes)
+        print(correct.iloc[:, 0])
```
jordanrfrazier, line 93: Intended to stay?

jordanrfrazier (approved):

sparrow-py/pysrc/sparrow_py/_expr.py
bjchambers: Leaning towards renaming this to `timelines` and thought `t` was too short, so went with `kt` for "kaskada timelines". We can adjust, but `kt` is also pretty easy to search for compared to `t` ;)

sparrow-py/pysrc/sparrow_py/_expr.py
bjchambers: 🤷 it's fine

sparrow-py/pysrc/sparrow_py/_expr.py
bjchambers: Dropped the import and switched to `kt`. There are some painful things about how imports within the same package are resolved (seems to lead to lots of circular imports). We may end up wanting to import specific modules (files) from within `sparrow_py` and only do the `import sparrow_py as kt` thing in tests / outside the package.

For instance, if you want `class Window(kt.Expr)` then that doesn't work if the window file is imported before the expr file in the `__init__.py` in `kt`, since when the window is imported, it hasn't yet been defined in the `__init__` module.

I think things are working now... but if it continues to be a problem I may suggest that within `pysrc` we always import from files (using relative paths).

sparrow-py/pysrc/sparrow_py/_ffi.pyi
bjchambers: Yeah. But this is currently the type signature (`pyi` not `py`) for a class that is defined in Rust code. We could maybe go the other way, but then we'd need to do a lot more casting / conversion that pyo3 defines for us this way.

sparrow-py/pytests/conftest.py
bjchambers: nope removed.

sparrow-py/pysrc/sparrow_py/_ffi.pyi
bjchambers: I guess we could do that though -- we could do:

1. Define the dataclass only in python
2. Pass that to rust.
3. Use https://docs.rs/pyo3/latest/pyo3/types/struct.PyAny.html#method.getattr to convert:

```
let row_limit: Option<usize> = options.getattr("row_limit").extract()?;
...
```

As we have more options that would decrease duplication. Leave open and I'll add this to the PR.

jordanrfrazier: ohh got it okay

PR #600, opened 2023-08-04
ref: Use a dataclass for execution options

bjchambers:
This reduces some duplication, in that we don't need to have the FFI define a Python-compatible copy of the Rust execution options. Instead we have it read straight from the documented dataclass.


jordanrfrazier (approved):

jordanrfrazier:
This looks better -- is there any functional/practical difference?

PR #601, opened 2023-08-04
feat: Introduce the Result class

bjchambers:
This allows multiple ways of interacting with the results of a query.

Also change the `show` method (which printed to stdout) to `preview` which returns a DataFrame "preview" of the results. This should play more nicely with testing as well as printing / visualizing as the user wishes.


sparrow-py/pysrc/sparrow_py/_result.py
```diff
@@ -1,2 +1,59 @@
...
+import pyarrow as pa
+
 class Result(object):
     """Result of running a timestream query."""
+
+    def __init__(self, ffi_execution: _ffi.Execution) -> None:
+        """Create the result."""
+        self._ffi_execution = ffi_execution
+
+    def to_pandas(self) -> pd.DataFrame:
```
bjchambers, line 13: One annoying thing is that you can't do:

```
results = expr.run()
results.to_pandas()
results.to_pandas()
```

Because the first call "consumes" the results. We may want to make this actually capture the expression and execute it on demand, which may change some of the naming. Let's discuss tomorrow?

@kerinin

kerinin (approved):

PR #602, opened 2023-08-04
feat: support `DataType::Struct` in collect function

jordanrfrazier:
Supports `DataType::Struct` in the collect function.


crates/sparrow-instructions/src/evaluators/aggregation/token/collect_struct_token.rs
```diff
@@ -0,0 +1,47 @@
...
+use crate::{ComputeStore, StateToken, StoreKey};
+
+/// Token used for collecting structs
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub struct CollectStructToken {
+    /// Stores the state for in-memory usage.
+    ///
+    /// A [ListArray] comprised of lists of structs for each entity.
+    #[serde(with = "sparrow_arrow::serde::array_ref")]
+    pub state: ArrayRef,
```
bjchambers, line 12: Eventulaly, we may be able to just have a single `ArrayRefStateToken` for any of these that use an Arrow array as state.

crates/sparrow-instructions/src/evaluators/list/collect_struct.rs
```diff
@@ -0,0 +1,797 @@
...
+
+        assert_eq!(expected.as_ref(), result);
+
+        // Batch 2
+        let mut builder = StructBuilder::new(fields(), field_builders());
+        builder
+            .field_builder::<Int64Builder>(0)
+            .unwrap()
+            .append_value(3);
+        builder
```
bjchambers, line 430: I think this can be cleaned up a bit. What I've done in the past is use a helper that has methods for `field1` and `field2` with the right types. Other patterns exist in the code elsewhere for testing things with structs. Not sure it's necessary. One easy thing would be to allow this test to return a `Result` and then replace `.unwrap()` with `?`.

bjchambers (approved):

crates/sparrow-instructions/src/evaluators/aggregation/token/collect_struct_token.rs
jordanrfrazier: Yes, agreed. Would love to spend some time to simplify the aggregations, tokens, and agg functions.

crates/sparrow-instructions/src/evaluators/list/collect_struct.rs
jordanrfrazier: Ah yeah -- I can clean this up big time by constructing each field's column separately and not using the builder.

PR #605, opened 2023-08-05
feat: First pass at doc site for Python

bjchambers:
This uses Sphinx and the
[sphinx-book-theme](https://sphinx-book-theme.readthedocs.io/en/stable/) to create organized documentation of the available methods for creating and executing timestreams.

Sphinx makes it easy to include syntax-highlighted examples from the documentation, as well as embedding the **rendered** notebook output of cells (embedding them in documentation we write) and even including notebooks as part of the documentation.

With a few tweaks, we can make every notebook cell executable as well. See for instance
https://jupyterbook.org/en/stable/interactive/thebe.html.


jordanrfrazier:
Haven't done any research into sphinx yet but approving to unblock and we can look at how it works together tomorrow

jordanrfrazier (approved):

PR #606, opened 2023-08-05
feat: Async iteration through Rust

bjchambers:


sparrow-py/pysrc/sparrow_py/_result.py
```diff
@@ -65,3 +66,32 @@ def iter_rows(self) -> Iterator[dict]:
...
+    async def iter_pandas_async(self) -> AsyncIterator[pd.DataFrame]:
+        """
+        Asynchronously iterate over the results as Pandas DataFrames.
+
+        Yields
+        ------
+        pd.DataFrame
+            The next Pandas DataFrame.
+        """
+        next_batch = await self._ffi_execution.next_pyarrow_async()
```
bjchambers, line 79: Curious on thoughts here.

This (and the other methods) currently rely on a simple method (returning an Option) and then use Python generators to turn that into an iterator.

An alternative would be having the rust code return the iterator itself (by implementing `__next__`/`__anext__`). We could do that, but those get more complicated (eg., https://pyo3.rs/v0.15.0/class/protocols.html#awaitable-objects and needing to return https://pyo3.rs/main/doc/pyo3/pyclass/enum.iteranextoutput). It's totally feasible, but I wonder if it's beneficial?

Even if we wanted the Python result to be iterable directly (eg, `async for expr.run()` would automatically iterate over batches, we can do that by *just* implementing `__aiter__` in Python).

And at the batch level, we shouldn't be crossing Python/Rust so much that the extra layer is too expensive, especially since it keeps the code much more straightforward.

Any thoughts here?

sparrow-py/pysrc/sparrow_py/_result.py
```diff
@@ -65,3 +66,32 @@ def iter_rows(self) -> Iterator[dict]:
             for row in next_batch.to_pylist():
                 yield row
             next_batch = self._ffi_execution.next_pyarrow()
+
+    async def iter_pandas_async(self) -> AsyncIterator[pd.DataFrame]:
```
bjchambers, line 70: I think it may be possible to combine `iter_pandas` and `iter_pandas_async` by returning a class that implements both `__next__` and `__anext__`. Then if you try to use it asynchronously, it will call `__anext__` and you'll need to await it. If not, it will call `__next__` and it will block.

Pro: One method to document -- the thing that iterates over pandas.
Con: Seems like it could lead to really easy errors where you do `for ...` instead of `async for ...` and things end up blocking unexpectedly.

Looking around, I don't see cases for doing that, so I suspect we probably shouldn't unless we get some confirmation it makes sense / is a way people do things.

jordanrfrazier: Agreed -- would like to make it as hard as possible to make errors like that that are hard to diagnose, especially while we're still developing.

sparrow-py/src/execution.rs
```diff
@@ -1,44 +1,97 @@
...
             Some(batch) => Some(batch.to_pyarrow(py)?),
             None => None,
         };
         Ok(result)
     }
+
+    fn next_pyarrow_async<'py>(&self, py: Python<'py>) -> PyResult<&'py PyAny> {
+        let execution = self.clone();
+        pyo3_asyncio::tokio::future_into_py(py, async move {
+            // We can't use `?` here because it attempts ot acquire the GIL unexpectedly.
```
jordanrfrazier, line 79: We can just wait a few years and they'll have removed the GIL! https://discuss.python.org/t/a-steering-council-notice-about-pep-703-making-the-global-interpreter-lock-optional-in-cpython/30474

sparrow-py/pysrc/sparrow_py/_result.py
jordanrfrazier: This looks good to me. It's simple to comprehend, I can print out batches as they are produced by the generator here, we wouldn't have to introduce more types I don't know about. 

I haven't done much reading on Py03 or the async options we have here. Are there any benefits to returning the iterator that you've since found?

jordanrfrazier (approved):

sparrow-py/src/execution.rs
bjchambers: Maybe? not sure if that will always be removed -- and it will change a good deal of things (eg., right now we need to get the lock to allocate, but then we know that we are the only one interacting with Python stuff).

sparrow-py/pysrc/sparrow_py/_result.py
bjchambers: I think it would just be letting Rust implement hte Async generator directly, rather than having an Async generator in Python that calls a method in Rust that returns `Option<RecordBatch>` (and is basically the `anext` method). But, it would require more classes (a separate class for each kind of iteration). I may poke at it later, but it unless Python is horribly slow about running the generator, it doesn't seem like it buys us much, and adds complexity to the layer between Python/Rust.

kerinin (approved):

PR #607, opened 2023-08-07
feat: add min parameter to collect

jordanrfrazier:
Adds the `min` parameter to collect, which allows a user to specify a minimum list length before a non-null value is produced from `collect()`. also reorders parameters to make the python builder arg pattern a little cleaner. 

Unfortunately, we still aren't able to use `collect` as a replacement for `lag`, given the current behavior. `Lag` does not count `null` as a valid value, while `collect` does. See the ignored unit test for an example of differences.

### DIFF
--- a/crates/sparrow-compiler/src/functions/collection.rs
+++ b/crates/sparrow-compiler/src/functions/collection.rs
@@ -14,9 +14,9 @@ pub(super) fn register(registry: &mut Registry) {
         .set_internal();
 
     registry
-        .register("collect<T: any>(const max: i64, input: T, window: window = null) -> list<T>")
+        .register("collect<T: any>(input: T, const max: i64, const min: i64 = 0, window: window = null) -> list<T>")
         .with_dfg_signature(
-            "collect<T: any>(const max: i64, input: T, window: bool = null, duration: i64 = null) -> list<T>",
+            "collect<T: any>(input: T, const max: i64, const min: i64 = 0, window: bool = null, duration: i64 = null) -> list<T>",
         )
         .with_implementation(Implementation::Instruction(InstOp::Collect))
         .set_internal();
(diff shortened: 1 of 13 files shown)


crates/sparrow-compiler/src/functions/collection.rs
```diff
@@ -14,9 +14,9 @@ pub(super) fn register(registry: &mut Registry) {
         .set_internal();
 
     registry
-        .register("collect<T: any>(const max: i64, input: T, window: window = null) -> list<T>")
+        .register("collect<T: any>(input: T, const max: i64, const min: i64 = 0, window: window = null) -> list<T>")
```
bjchambers, line 17: What's the benefit of making this part of `collect`? Are there ways it could be implemented more efficiently as a result? Are there ways it is more expressive to the user? If the former, then it could be part of the instruction, but not necessarily the function. If the latter, then it could be part of the Python builder, but not necessarily required.

The alternative would be to filter after `collect` using existing functionality:
```python
collected = collect(input, max)
collected.if_(collected.len_() >= min)
```

One benefit of that is that it is trivial to share the collecting across multiple cases with different min (we could even do different max that way). That said, if we wish to do this fully generally, we'd likely need to have an optimization that does something like that, so it doesn't hurt to have this be part of the function.

Thoughts? Worth discussing tomorrow?

bjchambers (approved):

crates/sparrow-compiler/src/functions/collection.rs
jordanrfrazier: This is a good point. Let's discuss tomorrow

jordanrfrazier: As discussed, one major benefit is that we avoid cloning values from `0..min`. For larger lists and min values, this can be a big performance gain. 

Especially if `collect` is used to implement lag, we'll consistently have the signature 
```
collect(min=n, max=n) | index(0)
```
where we only need to emit once the list is at size `n`.

PR #608, opened 2023-08-07
ref: Introduce a class for golden tests

bjchambers:
This makes things a lot clearer than using the argument to determine how the golden file should be produced.

Change `json` to `jsonl` to reflect that it is line-delimited.


sparrow-py/pytests/conftest.py
```diff
@@ -17,89 +17,102 @@ def pytest_addoption(parser: pytest.Parser):
...
+            if pd.api.types.is_datetime64_dtype(dtype):
+                parse_dates.append(name)
+            else:
+                dtypes[name] = dtype
+        golden = pd.read_csv(filename, dtype=dtypes, parse_dates=parse_dates)
+
+        pd.testing.assert_frame_equal(df, golden)
+
+    def jsonl(self, data: Union[kt.Timestream, pd.DataFrame]) -> None:
+        """Golden test against JSON file."""
```
jordanrfrazier, line 54: nit: can you add the comment that `jsonl` is the "json newlines format"

jordanrfrazier (approved):

PR #610, opened 2023-08-07
docs: Some doc tweaks

bjchambers:
Get the type alias to render, along with links. Rename the type alias to `Literal` and use it just for the literals.

### DIFF
--- a/sparrow-py/pysrc/sparrow_py/_timestream.py
+++ b/sparrow-py/pysrc/sparrow_py/_timestream.py
@@ -11,6 +11,7 @@ from typing import Sequence
 from typing import Tuple
 from typing import Union
 from typing import final
+from typing_extensions import TypeAlias
 
 import pandas as pd
 import pyarrow as pa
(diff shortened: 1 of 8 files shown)


sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -21,13 +22,9 @@
 from ._result import Result
 
 
-Arg = Union["Timestream", int, str, float, bool, None]
-"""
-The type of arguments to most Timestream functions.
-May be a Timestream, a literal (int, str, float, bool), or null (`None`).
-"""
```
jordanrfrazier, line 28: Hm, I kind of liked this conciseness. Maybe could have renamed `Arg` to something a bit more unique, but any reason you want the explicit `Union`?

jordanrfrazier (approved):

sparrow-py/pysrc/sparrow_py/_timestream.py
bjchambers: 1. The docstring never shows up in Sphinx etc., so I just removed it.
2. I kind of like being able to say "the argument is a timestream or a literal" which is why I changed from `Arg` to `Literal` as the common definition. That way we don't need to think about "what is `Arg`"

PR #611, opened 2023-08-07
feat: Introduce InMemoryBatches for managing data

bjchambers:
This supports the existing "add a batch to a table", but also allows subscribing to a stream of additions.

As part of this moved the existing (old) merge code to a separate crate, so it could be used by the compiler (which is currently responsible for the data contextt).


crates/sparrow-compiler/src/table_content.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    pub fn stream(&self) -> BoxStream<RecordBatch> {
+        async_stream::stream! {
+            let mut version = self.version;
+            let mut recv = self.updates.subscribe();
+
+            tracing::info!("Starting subscriber with version {version}");
+            yield self.merged;
+            while let Ok((recv_version, batch)) = recevier.recv().await {
```
jordanrfrazier, line 53: Would think this doesn't compile. Should be `recv`?

crates/sparrow-compiler/src/table_content.rs
```diff
@@ -0,0 +1,63 @@
...
+                tracing::
+                if version < recv_version {
+                    tracing::info!("Recevied version {recv_version}");
```
jordanrfrazier, line 56: Received

crates/sparrow-compiler/src/table_content.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub fn stream(&self) -> BoxStream<RecordBatch> {
+        async_stream::stream! {
+            let mut version = self.version;
+            let mut recv = self.updates.subscribe();
+
+            tracing::info!("Starting subscriber with version {version}");
+            yield self.merged;
+            while let Ok((recv_version, batch)) = recevier.recv().await {
+                tracing::
+                if version < recv_version {
```
jordanrfrazier, line 55: Logic here being to avoid race conditions where multiple callers `add_batch` and send updates out of order? Does there need to be a lock around the `add_batch` method? i.e. is it possible we have 
```
Self { batch: A} 
add_batch(B)
add_batch(C)
merge(A + B)
merge(A + C)
self.updates.send(A+B)
self.updates.send(A+C)
self.merge = A+B
self.merge=A+C
```

crates/sparrow-merge/src/in_memory_batches.rs
```diff
@@ -0,0 +1,116 @@
...
+
+        async_stream::try_stream! {
+            tracing::info!("Starting subscriber with version {version}");
+            yield merged;
+
+            loop {
+                match recv.recv().await {
+                    Ok((recv_version, batch)) => {
+                        if version < recv_version {
+                            tracing::info!("Recevied version {recv_version}");
```
jordanrfrazier, line 93: Received

crates/sparrow-merge/src/in_memory_batches.rs
```diff
@@ -0,0 +1,116 @@
...
+                            yield batch;
+                            version = recv_version;
+                        } else {
+                            tracing::warn!("Ignoring old version {recv_version}");
```
jordanrfrazier, line 97: perhaps include old version here as well? `Ignoring old version {recv_version}, current version {version}`. -- Might help us debug how far/if things are out of order for some reason

jordanrfrazier (approved):

crates/sparrow-compiler/src/table_content.rs
bjchambers: In this PR -- this is fine, because `add_batch` has `&mut self` which means it is the only reference. In the next PR, I had to fix that by introducing the lock.

crates/sparrow-merge/src/in_memory_batches.rs
bjchambers: K.

PR #614, opened 2023-08-07
Update notebook per Eric's prep steps

kerinin:


examples/slackbot/Notebook.ipynb
```diff
@@ -74,120 +67,89 @@
...
-    "SYSTEM_CONTEXT = \"\"\"\n",
-    "You are a helpful assistant designed to suggest the Slack usernames of \n",
-    "people who need to know about a Slack conversation.\n",
-    "\n",
-    "Only respond as a JSON list containing the Slack usernames of people to notify \n",
-    "of the conversation, or return an empty list if no should be notified.\n",
+    "def build_prompts(messages):\n",
+    "    message_time = messages.col(\"ts\")\n",
+    "    last_message_time = message_time.lag(1) # !!!\n",
+    "    is_new_conversation = message_time - last_message_time > 10 * 60\n",
```
bjchambers, line 77: We could get `seconds_between`. Getting a cleaner "single timedelta" approach will require more work -- Arrow has different kinds of durations for things like "Calendar Days", so a single "timedelta" may actually be multiple things (when you say days do you mean with or without leap seconds, etc.)

examples/slackbot/Notebook.ipynb
```diff
@@ -42,9 +34,10 @@
    "source": [
     "from datetime import datetime, timedelta\n",
     "from sklearn.model_selection import train_test_split\n",
-    "from slack_sdk.socket_mode import SocketModeClient\n",
+    "from slack_sdk.socket_mode import SocketModeClient, SocketModeResponse\n",
     "import openai\n",
     "import sparrow_pi as k\n",
+    "import sparrow_pi.windows as w\n",
```
bjchambers, line 40: We should be careful about the import aliases -- it can be very confusing, especially when it's not one of the "accepted" abbreviations. For searchability, I'd suggest:

1. We use two characters (more like `pd`, `pa` which is more likely to be unique compared to a single letter). Perhaps `kt` for kaskada timestreams?
2. Instead of improting windows separately, we should do `k.windows.Since(...)` (similar to `pd.testing.assert_frame_equal(...)`).

examples/slackbot/Notebook.ipynb
```diff
@@ -74,120 +67,89 @@
...
-  {
-   "cell_type": "markdown",
-   "id": "0035f558-23bd-4b4d-95a0-ed5e8fece673",
-   "metadata": {},
-   "source": [
-    "## Fine-tune the model"
+    "import pandas\n",
+    "\n",
+    "input_df = pandas.read_parquet(\"./messages.parquet\")\n",
+    "messages = sources.ArrowSource(\"ts\", \"channel\", input_df)\n",
```
bjchambers, line 123: We should probably hack a parquet source in Python, even if we don't have it plumbed up to read from Parquet in Sparrow yet. Eg., we could pretty easily add a `sources.ParquetSource` or `sources.Parquet(...)` class.

examples/slackbot/Notebook.ipynb
```diff
@@ -74,120 +67,89 @@
...
+    "messages = messages.with_key(k.record({  # !!!\n",
+    "        \"channel\": messages.col(\"channel\"),\n",
```
bjchambers, line 125: We don't have a `col` method. Did you mean `messages["thread_id"]`?

bjchambers: Although, maybe it would be better to use `col` *exclusively* for field access, and have `base[index]` be just for list/map indexing. One, that let's those be more clearly distinct. Two, it would allow `col` to automatically apply to `list<record>` and `map<_, record>` by applying `col` across the records.

examples/slackbot/Notebook.ipynb
kerinin: OK - going to call this `seconds_since` so the base operand is more clear.

examples/slackbot/Notebook.ipynb
kerinin: Yeah, I think the issue is that `[]` is ambiguous if you have a map of records.

epinzur (approved):

PR #615, opened 2023-08-07
feat: Add basic time arithmetic

bjchambers:


sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -22,7 +23,7 @@
 from ._result import Result
 
 
-Literal = Union[int, str, float, bool, None]
+Literal = Union[int, str, float, bool, None, timedelta]
```
bjchambers, line 26: Torn on whether to try to use more specific types to limit where `timedelta` can be used. I think probably not -- even if we protect against a literal in an unexpected place, a timedelta column could show up there.

sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -175,8 +181,28 @@ def __add__(self, rhs: Union[Timestream, Literal]) -> Timestream:
...
             The Timestream resulting from `self + rhs`.
-        """
-        return Timestream._call("add", self, rhs)
+
+        Raises
+        ------
+        ValueError
+            If attempting to perform an addition that is not possible.
+        """
+        if isinstance(rhs, timedelta):
```
bjchambers, line 190: This is a bit ugly. Basically, `timedelta` is represented as `days, seconds, microseconds`. We need to decide whether add days means "add 24*60*60 seconds" or "add 1 calendar day". Given it is kept in a granular form, and we said `timedelta(days=5)` I interpreted this as "add 5 calendar days".

Then the other problem is that that doesn't match the Arrow (or Sparrow) representation, which has separate "duration" and "interval" columns. Under the hood, a timedelta is almost a triple of `{ day: IntervalDaysColumn, seconds: DurationSeconds, microseconds: DurationMicroseconds }`. We could consider that as a generic representation, possibly with some optimizations when we know the column is zero. Short of that, we'll need to spend some time thinking about how to map the Python ("just timedelta") to Arrow (and Sparrow), and probably do some more work under the hood.

bjchambers: Actually, it's unclear how that would even work if you wanted to add "1 month". How much is that? Is it Februray? Is it a leap year?

This is why we also have an `IntervalMonths` column type.

bjchambers: It seems like we'd want to actually use `dateutil.relativedelta` for that (https://dateutil.readthedocs.io/en/stable/relativedelta.html).

So maybe the answer is:

1. `timedelta` is always an absolute (fixed) number of seconds.
2. `relativedelta` is used to capture a number of months / days / seconds to add.

But that is a bigger can of worms.

bjchambers: I think given ^^^ that, this should be https://docs.python.org/3/library/datetime.html#datetime.timedelta.total_seconds

And then we can use relativedelta if we need other things.

jordanrfrazier: Ah okay, that makes sense. So if we do have a literal `timedelta` we can just do the following?
```
result = Timestream._call("add_time", rhs.total_seconds(), result, session=session)
```

sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -175,8 +181,28 @@ def __add__(self, rhs: Union[Timestream, Literal]) -> Timestream:
...
+                days = Timestream._call("days", rhs.days, session = session)
+                result = Timestream._call("add_time", days, result)
+            if rhs.seconds != 0:
+                seconds = Timestream._call("seconds", rhs.seconds, session = session)
+                result = Timestream._call("add_time", seconds, result)
+            if rhs.microseconds != 0:
+                raise ValueError("Cannot add microseconds to a Timestream")
+            return result
+        else:
+            return Timestream._call("add", self, rhs)
```
jordanrfrazier, line 205: Doesn't this need to use `add_time` if we have a `rhs` that is a time?

sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -700,7 +726,7 @@ def lookup(self, key: Union[Timestream, Literal]) -> Timestream:
         Parameters
         ----------
         key : Union[Timestream, Literal]
-            The foreign key to lookup. 
+            The foreign key to lookup.
```
jordanrfrazier, line 729: Remind me the lint command?

jordanrfrazier (approved):

sparrow-py/pysrc/sparrow_py/_timestream.py
jordanrfrazier: I'm not sure what you mean by the last sentence. Is the timedelta not a literal here?

bjchambers: It is, and can be passed *anywhere* we accept a `Literal`. Alternative would be for `add` to say:

```
def __add__(self, rhs: Union[Timestream, Literal, timedelta]) -> Timestream:
```

Basically, choose where to allow it. But then we can still have a timedelta arise as a itmestream (eg., `seconds_between`, so it doesn't buy us much safety.

sparrow-py/pysrc/sparrow_py/_timestream.py
bjchambers: At this point we're in `else` an don't have a time.

sparrow-py/pysrc/sparrow_py/_timestream.py
bjchambers: This was just vscode. I think `nox -s fix_lint` should reformat though. Or you can run `black` directly.

sparrow-py/pysrc/sparrow_py/_timestream.py
bjchambers: 1. Need to convert to `seconds` first (and convert to int):

```
seconds = Timestream._call("seconds", int(rhs.total_seconds(), session=session)
return Timestream._call("add_time", seconds, self)
```

PR #617, opened 2023-08-08
feat: Asynchronous materializations

bjchambers:


kerinin (approved):

crates/sparrow-session/src/execution.rs
```diff
@@ -70,6 +78,16 @@ impl Execution {
         }
     }
 
+    /// Send the stop signal.
+    ///
+    /// This method does *not* wait for all batches to be processed.
```
jordanrfrazier, line 83: What do you mean by this? We will still process all in-flight batches, until each downstream component of the `Scan` receives the `None` message, right?

sparrow-py/pytests/collect_test.py
```diff
@@ -64,7 +64,7 @@ def test_collect_with_min(source, golden) -> None:
 def test_collect_with_min_and_max(source, golden) -> None:
     m = source["m"]
     n = source["n"]
-    golden(
+    golden.jsonl(
```
jordanrfrazier, line 67: May need to save the golden files again, looks like they were all deleted? Unless there are multiples

jordanrfrazier (approved):

crates/sparrow-session/src/execution.rs
bjchambers: Exactly -- it dosen't wait for those batches to be finished. Additionally, in-flight batches may still make their way out of the system after the call to stop returns.

PR #618, opened 2023-08-08
docs: some polishing / landing page work

bjchambers:


kerinin (approved):
We'll probably want to revisit once we've revised messaging.

PR #620, opened 2023-08-08
feat: shift sparrow functions

jordanrfrazier:
Adds shift-to and shift-until and shift-by sparrow functions

Lints by 
`nox -s fix-lint`


sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -730,6 +740,53 @@ def lookup(self, key: Union[Timestream, Literal]) -> Timestream:
...
+            This must be a datetime or a Timestream of timestamp_ns.
+
+        Returns
+        -------
+        Timestream
+            Timestream containing the shifted points.
+        """
+        if isinstance(time, datetime):
+            time_ns = time.timestamp() * 1e9
+            return Timestream._call("shift_to", time_ns, self)
```
jordanrfrazier, line 763: This doesn't work because it requires a `timestamp_ns` type. Aside from changing the fenl signature, what else could we quickly do here?

bjchambers: I think we'll need to implement cast, then we can do:

```
time_ns = time.timestamp() * 1e9
time_ns = Timestream._cast(time_ns, pa.timestamp('ns'))
return Timestream._call("shift_to", time_ns, self)
```

sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -776,6 +792,83 @@ def lookup(self, key: Union[Timestream, Literal]) -> Timestream:
         """
         return Timestream._call("lookup", key, self)
 
+    def shift_to(self, time: Union[Timestream, datetime]) -> Timestream:
+        """
+        Create a Timestream shifting the time of each point to `time`.
```
bjchambers, line 797: Maybe "shifting each point forward to `time`" (conceputally, shifts the point, not the time of the point)

sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -776,6 +792,83 @@ def lookup(self, key: Union[Timestream, Literal]) -> Timestream:
...
+        """
+        Create a Timestream shifting the time of each point by the delta `time`.
+
+        If multiple values are shifted to the same time, they will be emitted in
+        the order in which they originally occurred.
+
+        Parameters
+        ----------
+        time : Union[Timestream, datetime]
+            The time to shift the point forward by.
```
bjchambers, line 833: The distance? Interval? delta?

Also, type is wrong -- should be `timedelta`

bjchambers (approved):

PR #625, opened 2023-08-08
feat: implement lag with collect

jordanrfrazier:
This also changes the behavior of lag to `IGNORE NULLS` (likely similar to the options in this lag: https://learn.microsoft.com/en-us/sql/t-sql/functions/lag-transact-sql?view=sql-server-ver16).


crates/sparrow-compiler/src/dfg/const_eval.rs
```diff
@@ -34,7 +34,6 @@ pub(super) fn evaluate_constant(
             // now to fix various panics caused by not having *some* behavior defined.
             InstOp::CountIf => return Ok(ScalarValue::UInt32(Some(0))),
             InstOp::First => return Ok(inputs[0].null()),
-            InstOp::Lag => return Ok(inputs[0].null()),
```
bjchambers, line 37: Makes sense for now. I added an item to the list of "deferred" things -- we probably want to either (a) implement `Lag` more directly or (b) do the thing we talked about where we use a "collect buffer" and support outputting just specific elements. Cloning everything to output the nested list just to take one element is definitely going to be inefficient.

bjchambers (approved):

bjchambers:
Actually, now I see it. Thought -- if we continue down the Python route, more of these can be defined *exclusively* in Python.

jordanrfrazier:
> Actually, now I see it. Thought -- if we continue down the Python route, more of these can be defined _exclusively_ in Python.

Agreed. Can move the `fenl_rewrites` to python, and optimize there.

PR #628, opened 2023-08-08
feat: Support arithmetic chains

bjchambers:


sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -180,7 +180,7 @@ def pipe(
         else:
             return func(self, *args, **kwargs)
 
-    def __add__(self, rhs: Union[Timestream, Literal]) -> Timestream:
+    def add(self, rhs: Union[Timestream, Literal]) -> Timestream:
```
bjchambers, line 183: Interestingly, `numpy` suggests also having a top level `add` method, and then referencing it.

https://numpydoc.readthedocs.io/en/latest/format.html#method-docstrings

We likely don't want that for everything, but it's interesting that they like supporting `add(a, b)` and `a.add(b)`.

jordanrfrazier (approved):

PR #632, opened 2023-08-09
Update notebook to reflect what's possible

kerinin:


epinzur:
@kerinin should we close this?

epinzur:
we can move over the example from beepGPT when it is ready

PR #633, opened 2023-08-09
feat: hack collect lists by wrapping in struct

jordanrfrazier:
Hacks `collect(list<t>) -> list<list<t>>` support by wrapping it in a struct then field reffing. Only works from pyton builder.


sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -675,7 +675,13 @@ def collect(
         Timestream
             Timestream containing the collected list at each point.
         """
-        return _aggregation("collect", self, window, max, min)
+        if pa.types.is_list(self.data_type):
+            print("collecting list")
```
bjchambers: print?

sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -675,7 +675,13 @@ def collect(
...
+            input = record({"value": self})
```
bjchambers, line 678: Could also write as:

```
return record({"value": self}).collect(window=window, max=max, min=min).col("value")
```

(Eg., using collect rather than `_aggregation`). Not sure which is better.

bjchambers (approved):

PR #639, opened 2023-08-09
feat: Update key hash inverses as new data arrives

bjchambers:
Previously, we froze the key hash inverse based on the data that was available when a query started (by computing it from the merged data). Now, we compute it as data is added and maintain it.

Note: This does mean that if a `with_key` runs with the same grouping as an existing table, any keys it adds will be added to the in-memory key hash inverse.


crates/sparrow-runtime/src/execute.rs
```diff
@@ -200,23 +188,58 @@ pub async fn execute_new(
...
+/// Parameters
+/// ----------
+/// - key_hash_inverse: If set, specifies the key hash inverses to use. If None, the
+///   key hashes will be created.
+pub async fn execute_new(
+    plan: ComputePlan,
+    destination: Destination,
+    mut data_context: DataContext,
+    options: ExecutionOptions,
+    key_hash_inverses: Option<Arc<ThreadSafeKeyHashInverse>>,
```
jordanrfrazier, line 222: make this singular?

crates/sparrow-runtime/src/key_hash_inverse.rs
```diff
@@ -143,30 +155,53 @@ impl KeyHashInverse {
...
+        error_stack::ensure!(key_hashes.null_count() == 0, Error::KeyHashContainedNull);
+        error_stack::ensure!(
+            keys.data_type() == self.key.data_type(),
+            Error::IncompatibleKeyTypes {
+                expected: self.key.data_type().clone(),
+                actual: keys.data_type().clone(),
+            }
+        );
+        let mut len = self.key_hash_to_indices.len() as u64;
+        // Determine the indices that we nede to add.
```
jordanrfrazier, line 174: typo

crates/sparrow-runtime/src/key_hash_inverse.rs
```diff
@@ -255,8 +296,18 @@ impl ThreadSafeKeyHashInverse {
     /// This method is thread safe. It acquires the read lock to check if
     /// any of the keys need to be added to the inverse map, and only acquires
     /// the write lock if needed.
-    pub async fn add(&self, keys: ArrayRef, key_hashes: &UInt64Array) -> anyhow::Result<()> {
-        anyhow::ensure!(keys.len() == key_hashes.len());
+    pub async fn add(
```
jordanrfrazier, line 299: Is this method still safe? Why did we require a `blocking_add`?

crates/sparrow-session/src/session.rs
```diff
@@ -310,13 +329,24 @@ impl Session {
...
+            });
+
         // Hacky. Use the existing execution logic. This weird things with downloading checkpoints, etc.
         let progress = rt
             .block_on(sparrow_runtime::execute::execute_new(
                 plan,
                 destination,
                 data_context,
                 options,
+                Some(key_hash_inverses),
```
jordanrfrazier, line 349: Why are we creating an empty key hash inverse here? Don't we want it to be `None` so we trigger the `load_key_hash_inverse`?

bjchambers: Nope -- the loading of the key hash inverse is for old code. Basically, for Python we already have the key has for all the data that has been added to the in-memory session, so we just need to use that one. Any data adding during a materialization is updated (behind the thread-safety of the key hash inverse).

crates/sparrow-runtime/src/key_hash_inverse.rs
bjchambers: This one was async and most of the python code doesn't call from an async executor.

crates/sparrow-runtime/src/execute.rs
bjchambers: try to be consistent with other place (we use them interchangeably)

jordanrfrazier (approved):

PR #641, opened 2023-08-09
feat: trailing windows in collect for `DataType::Struct`

jordanrfrazier:
Only adds trailing windows for `DataType::Struct` in `collect()` function.


crates/sparrow-instructions/src/evaluators/aggregation/token/collect_struct_token.rs
```diff
@@ -18,7 +29,9 @@ impl StateToken for CollectStructToken {
             let state: CollectStructToken = state;
             self.state = state.state;
         };
-        Ok(())
+
+        // TODO: restore times
+        panic!("time restoration not implemented")
```
jordanrfrazier, line 34: I'd have to make a new store token and add some logic to the expression executor, so left this out for now.

crates/sparrow-instructions/src/evaluators/aggregation/token/collect_struct_token.rs
```diff
@@ -28,19 +41,45 @@ impl StateToken for CollectStructToken {
...
+            "item",
+            DataType::Timestamp(TimeUnit::Nanosecond, None),
+            true,
+        ));
+        let times_type = DataType::List(fr);
+        Self {
+            state,
+            times: new_empty_array(&times_type),
+        }
+    }
```
jordanrfrazier, line 54: Probably should split this up? But didn't want to have multiple `CollectStructEvaluators`. Unsure what the best separation is at the moment, so just using this as a combined token, and the non-windowed and since-windowed evaluators ignore the `times` array.

jordanrfrazier: Minor performance/memory hit in that we have to resize the `times` array, but it's all `nulls`.

crates/sparrow-instructions/src/evaluators/aggregation/token/collect_struct_token.rs
```diff
@@ -10,6 +13,14 @@ pub struct CollectStructToken {
     /// A [ListArray] comprised of lists of structs for each entity.
     #[serde(with = "sparrow_arrow::serde::array_ref")]
     pub state: ArrayRef,
+    /// Stores the times of the state values.
+    ///
+    /// A [ListArray] comprised of lists of timestamps for each entity.
+    ///
+    /// This array is only used when we have a `trailing` window.
+    /// Likely this should be separated into a different implementation.
```
bjchambers, line 21: It's probably OK if we just leave it an empty array -- wouldn't take too much space.

Thoughts on whether it should be a `Vec<Vec<Timestamp>>` or something? I guess it could even be just a `Vec<Timestamp>` with flattening, but not sure which would be easiest to work with.

crates/sparrow-instructions/src/evaluators/aggregation/token/collect_struct_token.rs
```diff
@@ -28,19 +41,45 @@ impl StateToken for CollectStructToken {
 
 impl CollectStructToken {
     pub fn new(state: ArrayRef) -> Self {
-        Self { state }
+        let fr = Arc::new(Field::new(
```
bjchambers, line 44: `fr`?

crates/sparrow-instructions/src/evaluators/list/collect_struct.rs
```diff
@@ -395,16 +419,142 @@ impl CollectStructEvaluator {
...
+                                combined_times.value(take_index as usize) - duration;
+                            while oldest_time < min_window_start {
+                                v.pop_front();
+                                // Safety: we know there's elements in the vec, and we can't
+                                // have popped the last element because we just added one at a time
+                                // greater than the `min_window_start`.
+                                let front_i = v.front().unwrap();
+                                oldest_time = combined_times.value(*front_i as usize);
+                            }
+                        }
```
jordanrfrazier, line 503: There might be a better solution where I map over the `take_output_builder` and `take_output_indices`, and do the slicing to time there, but this works for now (I think!)

sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -1047,11 +1047,13 @@ def _aggregation(
...
-        trailing_seconds = int(window.duration.total_seconds())
-        # HACK: Use null predicate and number of seconds to encode trailing windows.
-        return Timestream._call(op, input, *args, None, trailing_seconds)
+        if op != "collect":
+            raise NotImplementedError(
+                f"Aggregation '{op} does not support trailing windows"
+            )
+        trailing_ns = int(window.duration.total_seconds() * 1e9)
+        # HACK: Use null predicate and number of nanoseconds to encode trailing windows.
+        return Timestream._call(op, input, *args, None, trailing_ns)
```
jordanrfrazier, line 1056: time columns and duration are expected as nanoseconds

crates/sparrow-instructions/src/evaluators/list/collect_struct.rs
```diff
@@ -395,16 +419,142 @@ impl CollectStructEvaluator {
...
 
         Ok(Arc::new(result))
     }
+
+    /// Evaluates the collect function for structs with a `trailing` window.
+    ///
+    /// State is handled in order of "update -> emit -> reset".
+    ///
+    /// Follows the same implementation as above, but includes values in [current time - duration].
+    fn evaluate_trailing_windowed(
```
bjchambers, line 433: I wonder if there is a way to just use the existing logic. This is probably fine for now, but we're getting pretty duplicated. Really want to see if we can have better aggregation code in the partitioned path.

bjchambers (approved):

crates/sparrow-instructions/src/evaluators/aggregation/token/collect_struct_token.rs
jordanrfrazier: Hm, definitely a possibility. Though, since the machinery exists for serde, it's easy to operate on and use `take` with, so will leave as is for this PR.

PR #642, opened 2023-08-10
feat: Add union operation

bjchambers:


kerinin (approved):

crates/sparrow-instructions/src/evaluators/list/union.rs
```diff
@@ -0,0 +1,120 @@
...
+
+            if a_len == 0 && b_len == 0 {
+                // Nothing to do
+            } else if a_len == 0 {
+                // Only need to take from b.
+                offset += b_len as u32;
+                indices.extend((b_start..b_end).map(|n| (1, n)));
+            } else if b_len == 0 {
+                // Only need to take from b.
+                offset += b_len as u32;
```
jordanrfrazier, line 83: Only need to take from `a`
offset += a_len

bjchambers: 🤦 Just spent 30m debugging that and the fact that `b_list = a.as_list()` causes issues. But we got a rust unit test out of it, so I guess that's good?

crates/sparrow-instructions/src/evaluators/list/union.rs
```diff
@@ -0,0 +1,120 @@
...
+        let mut offsets = BufferBuilder::new(a_list.len() + 1);
+
+        let mut indices = Vec::with_capacity(a_list.values().len() + b_list.values().len());
+
+        let mut offset = 0u32;
+        offsets.append(offset);
+
+        let mut included = HashSet::new();
+        let a_offsets = a_list
+            .value_offsets()
```
jordanrfrazier: should these use `offsets` rather than `value_offsets`? https://docs.rs/arrow/latest/arrow/array/struct.GenericListArray.html#method.offsets

sparrow-py/pytests/conftest.py
```diff
@@ -103,6 +98,13 @@ def _data_to_dataframe(data: Union[kt.Timestream, pd.DataFrame]) -> pd.DataFrame
     else:
         raise ValueError(f"data must be a Timestream or a DataFrame, was {type(data)}")
 
+def _data_to_pyarrow(data: Union[kt.Timestream, pa.RecordBatch, pa.Table]) -> Union[pa.RecordBatch, pa.Table]:
+    if isinstance(data, kt.Timsetream):
```
jordanrfrazier, line 102: timestream

sparrow-py/pytests/golden/length_test/test_length.jsonl
jordanrfrazier, line 1: a

jordanrfrazier (approved):

crates/sparrow-instructions/src/evaluators/list/union.rs
bjchambers: They are equivalent

PR #645, opened 2023-08-10
feat: Run simplifications in sparrow-py

bjchambers:
I'm not sure why, but this seems to fix the issue applying `last` to `with_key`.


crates/sparrow-compiler/src/dfg/expr.rs
```diff
@@ -67,6 +68,25 @@ impl DfgExpr {
         &self.expr
     }
 
+    pub fn simplify(self, options: &CompilerOptions) -> anyhow::Result<Self> {
```
bjchambers, line 71: Operating on the expression is actually a lot cleaner than how the compiler currently does it (by mutating the DFG). Definitely a potential cleanup here... but not pushing it through the codebase yet.

jordanrfrazier (approved):

PR #651, opened 2023-08-10
feat: add a `seconds since previous` function

jordanrfrazier:
Not quite the `seconds_since(other)` function we wanted, but this should encapsulate the following behavior at least
```
nanoseconds_since_previous: (time_of(Foo) as i64) - (((time_of(Foo) | lag(n)) as i64)
```

### DIFF
--- a/sparrow-py/pysrc/sparrow_py/_timestream.py
+++ b/sparrow-py/pysrc/sparrow_py/_timestream.py
@@ -877,6 +877,22 @@ class Timestream(object):
         """
         return Timestream._call("shift_until", predicate, self)
 
+    def seconds_since_previous(self, n: int = 1) -> Timestream:
+        """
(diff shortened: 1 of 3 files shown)


sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -877,6 +877,22 @@ def shift_until(self, predicate: Timestream) -> Timestream:
         """
         return Timestream._call("shift_until", predicate, self)
 
+    def seconds_since_previous(self, n: int = 1) -> Timestream:
```
bjchambers, line 880: We could have `e.seconds_since()` (from #652) automatically be seconds_since_previous.

jordanrfrazier:
Closing -- going to combine this with `seconds_since`

PR #652, opened 2023-08-10
feat: add seconds_since timestamp_ns

jordanrfrazier:


sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -952,6 +952,36 @@ def cast(self, data_type: pa.DataType) -> Timestream:
         """
         return Timestream(self._ffi_expr.cast(data_type))
 
+    def seconds_since(self, time: Union[Timestream, datetime]) -> Timestream:
```
bjchambers, line 955: We could use this to implement `seconds_since_previous().` We could even have `e.seconds_since()` be the seconds since the previous one, and `e.seconds_since(f)` be the second since `f`.

sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -952,6 +952,36 @@ def cast(self, data_type: pa.DataType) -> Timestream:
...
+            nanos = Timestream.cast(nanos, pa.timestamp('ns', None))
+            self_nanos = Timestream._call("time_of", self)
+            return Timestream._call(
+                "seconds_between", nanos, self_nanos
+            ).cast(pa.int64())
+        else:
+            return Timestream._call(
+                "seconds_between",
+                time,
+                self
```
bjchambers, line 982: Wouldn't this need to be self_nanos to apply `time_of` to self?

sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -952,6 +952,36 @@ def cast(self, data_type: pa.DataType) -> Timestream:
...
+        Returns
+        -------
+        Timestream
+            Timestream containing the number of seconds since `time`.
+        """
+        if isinstance(time, datetime):
+            session = self._ffi_expr.session()
+            nanos = Timestream._literal(time.timestamp() * 1e9, session=session)
+            nanos = Timestream.cast(nanos, pa.timestamp('ns', None))
+            self_nanos = Timestream._call("time_of", self)
```
bjchambers, line 974: Based on the conversation, it sounds like this shouldn't call `time_of` -- it should expect `self` to be a timestream of timestamps.

jordanrfrazier: Perhaps in the demo, though I was intending for this function to be called on any streaming event. 

Essentially just syntactic sugar that skips you having to get the time column out and call lag yourself.

jordanrfrazier: Ope, I'm confusing myself with the other PR.

sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -993,6 +993,44 @@ def cast(self, data_type: pa.DataType) -> Timestream:
...
+        """
+        Create a Timestream containing the number of seconds since `time`.
+
+        Parameters
+        ----------
+        n : Union[Timestream, Literal]
+            The time to compute the seconds since.
+
+            This can be either a stream of timestamps, a datetime literal, or 
+            an integer literal representing the number of points to lag by.
```
kerinin, line 1006: Can we pull this behavior out into a separate function? This is a very subtle difference in usage that produces a substantially different behavior - for example it would be completely reasonable to assume `e.seconds_since(1691761917)` means "the time since a unix timestamp to each value in `e`".

jordanrfrazier: Yeah -- I originally had another PR out with a function called `seconds_since_previous` that separated that behavior. I agree with you here. And could add the timestamp functionality to this one.

python/pysrc/kaskada/_timestream.py
```diff
@@ -1009,6 +1009,51 @@ def cast(self, data_type: pa.DataType) -> Timestream:
         """
         return Timestream(self._ffi_expr.cast(data_type))
 
+    def seconds_since(self, time: Union[Timestream, Literal]) -> Timestream:
+        """
+        Create a Timestream containing the number of seconds since `time`.
```
bjchambers, line 1014: Clarify docs -- what is "now"? Is this "seconds between `time` and the value of `self`? "seconds between `time` and `self.time_of()`, etc.?

python/pysrc/kaskada/_timestream.py
```diff
@@ -1009,6 +1009,51 @@ def cast(self, data_type: pa.DataType) -> Timestream:
...
+        Timestream
+            Timestream containing the number of seconds since `time`.
+        """
+        if isinstance(time, datetime):
+            session = self._ffi_expr.session()
+            nanos = Timestream._literal(time.timestamp() * 1e9, session=session)
+            nanos = Timestream.cast(nanos, pa.timestamp("ns", None))
+            return Timestream._call("seconds_between", nanos, self).cast(pa.int64())
+        else:
+            return Timestream._call("seconds_between", time, self).cast(pa.int64())
```
bjchambers, line 1034: Why does this cast back to integer vs. leaving it as `seconds` (the duration unit)?

python/pysrc/kaskada/_timestream.py
```diff
@@ -1009,6 +1009,51 @@ def cast(self, data_type: pa.DataType) -> Timestream:
...
+
+    def seconds_since_previous(self, n: int = 1) -> Timestream:
+        """
+        Create a Timestream containing the number of seconds since the time
+        `n` points ago.
+
+        Parameters
+        ----------
+        n : int
```
bjchambers, line 1043: Mention the default?

python/pysrc/kaskada/_timestream.py
```diff
@@ -1009,6 +1009,51 @@ def cast(self, data_type: pa.DataType) -> Timestream:
...
+            The number of points to lag by.
```
bjchambers, line 1044: "lag" seems like it could use an explanation?

bjchambers (approved):

python/pysrc/kaskada/_timestream.py
jordanrfrazier: Changed back to using `duration('s')` and modified `_previous` as well to do the same. Casted explicitly in the tests.

PR #655, opened 2023-08-10
docs: Update existing docs

bjchambers:
Tweak wording on a few pages.
Make sure all the methods are documented.

Use poetry groups for dependencies in noxfile.

### DIFF
--- a/sparrow-py/pysrc/sparrow_py/_timestream.py
+++ b/sparrow-py/pysrc/sparrow_py/_timestream.py
@@ -469,6 +469,10 @@ class Timestream(object):
         -------
         Timestream
             The Timestream indicating whether the `self` and `other` are equal.
+
+        Note
+        ----
+        Equality is *not* available as `a == b`.
         """
         return Timestream._call("eq", self, other)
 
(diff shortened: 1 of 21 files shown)


sparrow-py/pysrc/sparrow_py/_timestream.py
```diff
@@ -485,10 +489,14 @@ def ne(self, other: Union[Timestream, Literal]) -> Timestream:
...
             The Timestream indicating whether `self` and `other` are not equal.
+
+        Note
+        ----
+        Inequality is *not* available as `a != b`.
         """
         return Timestream._call("neq", self, other)
 
-    def __getitem__(self, key: Union[Timestream, Literal]) -> Timestream:
+    def index(self, key: Union[Timestream, Literal]) -> Timestream:
```
jordanrfrazier, line 499: So when we have maps we'll do `map.index("field_name")`?

jordanrfrazier (approved):

sparrow-py/pysrc/sparrow_py/_timestream.py
jordanrfrazier: Or, `map["field_name"]` I suppose is available too?

bjchambers: We can do either -- `a[b]` calls `a.index[b]`. It seemed cleaner to say "there is a function called index, and it does X, Y, Z, and you can write it as `a[b]` for short" than to document `__getitem__` directly.

PR #658, opened 2023-08-11
feat: hack collect trailing with shift by

jordanrfrazier:
Attempts to fix the behavior of trailing windows in collect by forcing a merge with a shifted message to the next `duration`, which clears out the window of the previous event.


sparrow-py/pytests/golden/collect_test/test_collect_bool_trailing_window_1s.jsonl
```diff
@@ -1,6 +1,12 @@
...
+{"_time":"1996-12-19T16:39:59.000","_subsort":1,"_key_hash":12960666915911099378,"_key":"A","b":null,"collect_b":[]}
+{"_time":"1996-12-19T16:40:00.000","_subsort":1,"_key_hash":12960666915911099378,"_key":"A","b":null,"collect_b":[]}
+{"_time":"1996-12-19T16:40:00.000","_subsort":2,"_key_hash":12960666915911099378,"_key":"A","b":false,"collect_b":[false]}
+{"_time":"1996-12-19T16:40:01.000","_subsort":2,"_key_hash":12960666915911099378,"_key":"A","b":null,"collect_b":[]}
+{"_time":"1996-12-19T16:40:01.000","_subsort":3,"_key_hash":12960666915911099378,"_key":"A","b":false,"collect_b":[false]}
+{"_time":"1996-12-19T16:40:02.000","_subsort":3,"_key_hash":12960666915911099378,"_key":"A","b":null,"collect_b":[]}
+{"_time":"1996-12-19T16:40:02.000","_subsort":4,"_key_hash":12960666915911099378,"_key":"A","b":true,"collect_b":[true]}
+{"_time":"1996-12-19T16:40:03.000","_subsort":4,"_key_hash":12960666915911099378,"_key":"A","b":null,"collect_b":[]}
+{"_time":"1996-12-19T16:40:04.000","_subsort":5,"_key_hash":12960666915911099378,"_key":"A","b":null,"collect_b":[]}
+{"_time":"1996-12-19T16:40:05.000","_subsort":0,"_key_hash":12960666915911099378,"_key":"A","b":null,"collect_b":[]}
```
jordanrfrazier, line 12: These are the results I expect. However, if you uncomment the 
```
       # return collect.filter(is_input)
```
and rerun this test, the results change unexpectedly. I presume it is some continuity behavior happening, but unsure. 

```
{"_time":"1996-12-19T16:39:57.000","_subsort":0,"_key_hash":12960666915911099378,"_key":"A","b":true,"collect_b":[true]}
{"_time":"1996-12-19T16:39:59.000","_subsort":1,"_key_hash":12960666915911099378,"_key":"A","b":null,"collect_b":[true]}
{"_time":"1996-12-19T16:40:00.000","_subsort":2,"_key_hash":12960666915911099378,"_key":"A","b":false,"collect_b":[false]}
{"_time":"1996-12-19T16:40:01.000","_subsort":3,"_key_hash":12960666915911099378,"_key":"A","b":false,"collect_b":[false]}
{"_time":"1996-12-19T16:40:02.000","_subsort":4,"_key_hash":12960666915911099378,"_key":"A","b":true,"collect_b":[true]}
{"_time":"1996-12-19T16:40:04.000","_subsort":5,"_key_hash":12960666915911099378,"_key":"A","b":null,"collect_b":[true]}

```

Note that the second row also contains a `[true]`, as well as the last row. It may be storing the incorrect values in the latched spread, as we can tell that without the filter, we are producing an empty list (and not null). However, that leads to the conclusion that there may be a bug in the latched spread, which I'm hesitant to blame without more evidence.

crates/sparrow-instructions/src/evaluators/list/collect_boolean.rs
```diff
@@ -116,13 +116,13 @@ impl CollectBooleanEvaluator {
...
             // Do not collect null values
-            if input.is_some() {
+            if let Some(input) = input {
                 self.token.add_value(self.max, entity_index, input);
             }
 
             let cur_list = self.token.state(entity_index);
             if cur_list.len() >= self.min {
-                list_builder.append_value(cur_list.iter().copied());
+                list_builder.append_value(cur_list.iter().map(|i| Some(*i)));
```
jordanrfrazier, line 125: Should I just leave `cur_list` as a `Vec<Option<T>>` so I don't have to re-wrap them in `Option` each iteration?

python/pytests/golden/collect_test/test_collect_basic.jsonl
```diff
@@ -4,3 +4,4 @@
 {"_time":"1996-12-19T16:40:00.000","_subsort":3,"_key_hash":12960666915911099378,"_key":"A","m":null,"collect_m":[5,17],"n":9.0,"collect_n":[10,6,9]}
 {"_time":"1996-12-19T16:40:01.000","_subsort":4,"_key_hash":12960666915911099378,"_key":"A","m":12.0,"collect_m":[5,17,12],"n":null,"collect_n":[10,6,9]}
 {"_time":"1996-12-19T16:40:02.000","_subsort":5,"_key_hash":12960666915911099378,"_key":"A","m":null,"collect_m":[5,17,12],"n":null,"collect_n":[10,6,9]}
+{"_time":"1996-12-19T16:40:04.000","_subsort":6,"_key_hash":12960666915911099378,"_key":"A","m":null,"collect_m":[5,17,12],"n":null,"collect_n":[10,6,9]}
```
jordanrfrazier, line 7: These additional rows are because I added a new input row

python/pytests/golden/collect_test/test_collect_bool_trailing_window_1s.jsonl
```diff
@@ -1,6 +1,14 @@
...
+{"_time":"1996-12-19T16:39:58.000","_subsort":0,"_key_hash":12960666915911099378,"_key":"A","b":null,"collect_b":[]}
 {"_time":"1996-12-19T16:39:58.000","_subsort":1,"_key_hash":2867199309159137213,"_key":"B","b":true,"collect_b":[true]}
-{"_time":"1996-12-19T16:39:59.000","_subsort":2,"_key_hash":12960666915911099378,"_key":"A","b":null,"collect_b":[true]}
+{"_time":"1996-12-19T16:39:59.000","_subsort":1,"_key_hash":2867199309159137213,"_key":"B","b":null,"collect_b":[]}
+{"_time":"1996-12-19T16:39:59.000","_subsort":2,"_key_hash":12960666915911099378,"_key":"A","b":null,"collect_b":[]}
+{"_time":"1996-12-19T16:40:00.000","_subsort":2,"_key_hash":12960666915911099378,"_key":"A","b":null,"collect_b":[]}
 {"_time":"1996-12-19T16:40:00.000","_subsort":3,"_key_hash":12960666915911099378,"_key":"A","b":false,"collect_b":[false]}
-{"_time":"1996-12-19T16:40:01.000","_subsort":4,"_key_hash":12960666915911099378,"_key":"A","b":false,"collect_b":[false,false]}
-{"_time":"1996-12-19T16:40:02.000","_subsort":5,"_key_hash":12960666915911099378,"_key":"A","b":true,"collect_b":[false,true]}
```
jordanrfrazier, line 6: These lists are now correctly cleared on the second

python/pytests/golden/collect_test/test_collect_bool_trailing_window_1s.jsonl
```diff
@@ -1,6 +1,14 @@
 {"_time":"1996-12-19T16:39:57.000","_subsort":0,"_key_hash":12960666915911099378,"_key":"A","b":true,"collect_b":[true]}
+{"_time":"1996-12-19T16:39:58.000","_subsort":0,"_key_hash":12960666915911099378,"_key":"A","b":null,"collect_b":[]}
```
jordanrfrazier, line 2: Empty list rows like this represent the "clearing" of the window after `duration` has passed.

python/pysrc/kaskada/_timestream.py
```diff
@@ -1149,9 +1149,23 @@ def _aggregation(
...
+        #
+        # Note: Assumes the "input" is discrete. Can probably
+        # use a transform to make it discrete (eg., `input.filter(True)`)
+        # or a special function to do that.
+        #
+        # HACK: This places an extra null row in the input to `collect`
+        # which allows us to "clear" the window when the appropriate 
+        # `duration` has passed with no "real" inputs.
+        merged_input = record({ "input": input, "shift": input_shift }).col("input")
+        return Timestream._call("collect", merged_input, *args, None, trailing_ns)
```
bjchambers, line 1168: Don't we need to do the filtering here to ensure the right number of rows? I guess the answer is no since those rows are the clearing, and we actually want those.

bjchambers (approved):

python/pysrc/kaskada/_timestream.py
jordanrfrazier: Yeah, it's odd. Similar to how a `tick` introduces a row, the `trailing` window introduces a row at the time of the last event + duration. However, this row is really intended to be a "clearing" of the buffer, meaning we shouldn't need it unless a whole `duration` of time elapses. 

It would be easier to manage this behavior if all inputs were known to be "real" inputs -- we could compare the time to the next input and insert a "clearing" row if necessary. 

For now, I think it's okay for our specific use case, but it will introduce some noisy output (doubles the output rows) for larger inputs.

PR #661, opened 2023-08-15
ci: Stub out CI for new python package

bjchambers:
Initial stubs were generated by `maturin`.

Updated (I think) to reflect the location in `python`.


python/pyproject.toml
```diff
@@ -121,3 +129,20 @@ show_error_context = true
 testpaths = [
     "pytests",
 ]
+
+[tool.semantic_release]
```
bjchambers, line 133: With this configuration, the `python-semantic-release` tool can be used to generate version numbers, tag branches, and push changelogs to GitHub. Need to verify / test. It will use the `-a` prerelease token for the `main` branch, allowing us to (I hope) do continuous deployment using auto-incrementing alpha versions off main.

.github/workflows/ci_python.yml
```diff
@@ -0,0 +1,175 @@
...
+          command: sdist
+          args: --out dist
+          working-directory: ${{ github.workspace }}/python
+      - name: Upload sdist
+        uses: actions/upload-artifact@v3
+        with:
+          name: wheels
+          path: ${{ github.workspace }}/python/dist
+
+  release:
```
bjchambers, line 160: The release downloads the wheel we produced on each OS and then publishes to PyPI using `maturin upload`.

.github/workflows/ci_python.yml
```diff
@@ -0,0 +1,175 @@
...
+      - master
+    tags:
+      - '*'
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
```
bjchambers, line 21: I think each of these flows will need to run the `python-semantic-release` tool to update the version in the local file to reflect the release we want to publish. Not sure enough about Python versioning yet.

.github/workflows/ci_python.yml
```diff
@@ -0,0 +1,175 @@
...
+  linux:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target: [x86_64, x86, aarch64, armv7, s390x, ppc64le]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
```
bjchambers, line 31: We may want to have a step after this that runs tests (and type-checking, etc.) in multiple versions.

.github/workflows/ci_python.yml
```diff
@@ -0,0 +1,175 @@
...
+        with:
+          name: wheels
+          path: ${{ github.workspace }}/python/dist
+      - name: pytest
+        if: ${{ startsWith(matrix.target, 'x86_64') }}
+        shell: bash
+        working-directory: ${{ github.workspace }}/python
+        run: |
+          set -e
+          pip install kaskada --find-links dist --force-reinstall
```
bjchambers, line 51: I need to verify -- the intention is to build the wheel, and then install using that, which shouldn't require any building and should let us test things.

epinzur (approved):
mostly want to make sure we set appropriate versions... we already have many 0.1.0 alpha versions that maybe we don't want to clobber.

python/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "kaskada"
 authors = ["Kaskada Developers"]
 edition = "2021"
 license = "Apache-2.0"
-version = "0.6.0"
+version = "0.1.0-a.1"
```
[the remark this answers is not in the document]
bjchambers: Yeah, this is a combination of (a) the rust version (which we may decide doesn't matter) and (b) the `python-semantic-release` not detecting a previous tag.

Do you think this should be `0.6.0-a.1` (> past Python versions) or `0.12.0.a.1` (> past Python *and* engine versions)? I think either could make sense.

jordanrfrazier (approved):

PR #664, opened 2023-08-16
feat: Support floating point time columns

bjchambers:
Also allow specifying the time unit for int and float.


crates/sparrow-runtime/src/prepare/preparer.rs
```diff
@@ -40,14 +44,17 @@ impl Preparer {
         key_column_name: String,
         prepared_schema: SchemaRef,
         prepare_hash: u64,
-    ) -> Self {
-        Self {
+        time_unit: Option<&str>,
```
bjchambers, line 47: Rust code changed to:

a) Accept a `time_unit` (to use for int64 and float64 times). It's optional, and currently defaults to `ns`. We could make it required (when time is a number) and/or change the default if we wish.
b) Instead of just casting the time column, it applies some logic to perform the multiplication.

crates/sparrow-session/src/session.rs
```diff
@@ -117,7 +119,7 @@ impl Session {
             })
             .clone();
 
-        Ok(Table::new(table_info, key_hash_inverse, key_column, expr))
+        Table::new(table_info, key_hash_inverse, key_column, expr, time_unit)
```
bjchambers, line 122: Plumb the time_unit in the "sparrow-session" (native rust session crate).

python/pysrc/kaskada/_ffi.pyi
```diff
@@ -47,6 +47,7 @@ class Table(Expr):
         schema: pa.Schema,
         subsort_column_name: Optional[str],
         grouping_name: Optional[str],
+        time_unit: Optional[str],
```
bjchambers, line 50: Add the option to the Python type annotations for the code exposed from Rust.

python/pysrc/kaskada/sources/source.py
```diff
@@ -60,6 +61,7 @@ def fix_field(field: pa.Field) -> pa.Field:
             schema,
             subsort_column_name,
             grouping_name,
+            time_unit,
```
bjchambers, line 64: Add the keyword argument to the Python `Source` class.

python/pytests/conftest.py
```diff
@@ -30,7 +30,7 @@ def jsonl(self, data: Union[kd.Timestream, pd.DataFrame]) -> None:
         filename = self._filename("jsonl")
 
         if self._save:
-            df.to_json(filename, orient="records", lines=True, date_format="iso")
+            df.to_json(filename, orient="records", lines=True, date_format="iso", date_unit="ns")
```
bjchambers, line 33: Tweak tests to use `ns` as the unit -- this changes the precision in a lot of golden files, hence the churn, but it was necessary for the golden test comparison to pass in the new tests.

python/pytests/pandas_source_test.py
```diff
@@ -1,37 +1,41 @@
-import random
-
 import kaskada as kd
 import pandas as pd
 
+def test_float_microseconds(golden) -> None:
```
bjchambers, line 4: New tests taking floating point numbers in different precisions. Only this test changed -- and only this tests golden files are substantively different.

python/pytests/pandas_source_test.py
```diff
@@ -1,37 +1,41 @@
...
+    data = {
+        'time': [1671477472.026119],
```
epinzur, line 6: seems like this is actually a seconds timestamp.  if it was microseconds it would be `1671477472026119.123`

python/pytests/pandas_source_test.py
epinzur: the actual time from this should be something like `Mon Dec 19 2022 19:17:52 GMT+0000`

epinzur (approved):
pls update new test

PR #665, opened 2023-08-16
ref: rename time_of to time

jordanrfrazier:
Could probably go ahead and rename the sparrow function as well if we want.


bjchambers (approved):
Probably wouldn't rename Sparrow yet. It makes sense to write `time_of(X)` but not `X.time_of()` (eg., the way chaining works in Pyhton, `time` made sense, but Fenl is a functon not a method).

bjchambers:
Is this ready to merge?

PR #669, opened 2023-08-17
feat: add other aggregations

kevinjnguyen:
Adds the following aggregations: count_if, max, mean, min, stddev, and variance


python/pysrc/kaskada/_timestream.py
```diff
@@ -1013,6 +1013,126 @@ def count(self, window: Optional[kd.windows.Window] = None) -> Timestream:
         """
         return _aggregation("count", self, window)
 
+    def count_if(self, window: Optional[kd.windows.Window] = None) -> Timestream:
+        """
+        Create a Timestream containing the count value if true in the `window`.
```
bjchambers, line 1018: "count of `true` values in `window`"?

(eg., "count value if true" could suggest that it is `.count().if_(v)`, rather than counting the `true` values).

python/pysrc/kaskada/_timestream.py
```diff
@@ -1013,6 +1013,126 @@ def count(self, window: Optional[kd.windows.Window] = None) -> Timestream:
...
+        Parameters
+        ----------
+        window : Optional[Window]
+            The window to use for the aggregation.
+            If not specified, the entire Timestream is used.
+
+        Returns
+        -------
+        Timestream
+            Timestream containing the count value if true for the key in the window for
```
bjchambers, line 1031: .

python/pysrc/kaskada/_timestream.py
```diff
@@ -1013,6 +1013,126 @@ def count(self, window: Optional[kd.windows.Window] = None) -> Timestream:
...
+        -------
+        Timestream
+            Timestream containing the mean value for the key in the window for
+            each point.
+        """
+        return _aggregation("mean", self, window)
+
+    def stddev(self, window: Optional[kd.windows.Window] = None) -> Timestream:
+        """
+        Create a Timestream containing the standard deviation value in the `window`.
```
bjchambers, line 1098: I don't know that we need to say "standard deviation value" -- just "standard deviation" should do?

python/pysrc/kaskada/_timestream.py
```diff
@@ -1013,6 +1013,126 @@ def count(self, window: Optional[kd.windows.Window] = None) -> Timestream:
...
+        -------
+        Timestream
+            Timestream containing the standard deviation for the key in the window for
+            each point.
+        """
+        return _aggregation("stddev", self, window)
+
+    def variance(self, window: Optional[kd.windows.Window] = None) -> Timestream:
+        """
+        Create a Timestream containing the variance value in the `window`.
```
bjchambers, line 1118: s/variance value/variance?

python/pytests/aggregation/count_if_test.py
```diff
@@ -0,0 +1,63 @@
+import pytest
+import kaskada as kd
+
+
+@pytest.fixture
```
bjchambers, line 5: Since this is used multiple times, there would be a benefit to using `@pytest.fixture(scope = "module")`. Specifically, that would share the source across all test in this file (module) rather than creating (and preparing / loading) a new source for each test.

python/pytests/aggregation/count_if_test.py
```diff
@@ -0,0 +1,63 @@
...
+            {
+                "is_valid": is_valid,
+                "count_if": is_valid.count_if(),
+                "m": m,
+            }
+        )
+    )
+
+
+# TODO: This test doesn't appear correct.
```
bjchambers, line 37: What doesn't appear correct?

python/pytests/aggregation/max_test.py
```diff
@@ -0,0 +1,33 @@
+import kaskada as kd
+
+
+def test_max_unwindowed(source, golden) -> None:
```
bjchambers, line 4: How are these executing? It doesn't look like they have a source defined?

kevinjnguyen: Yup, I addressed these comments. Please take another look

bjchambers (approved):

PR #674, opened 2023-08-17
feat: add udf sparrow trait 

jordanrfrazier:
Adds the sparrow trait for user defined functions. This will allow us to both keep python dependencies in `python` crates and implement evaluators from other `crates`.

### DIFF
--- a/crates/sparrow-instructions/src/inst.rs
+++ b/crates/sparrow-instructions/src/inst.rs
@@ -264,7 +264,7 @@ impl PartialEq for InstKind {
 impl fmt::Display for InstKind {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            InstKind::Simple(instruction) => write!(f, "{instruction:?}"),
+            InstKind::Simple(instruction) => write!(f, "{instruction}"),
             InstKind::FieldRef => write!(f, "field"),
             InstKind::Cast(data_type) => write!(f, "cast:{data_type}"),
             InstKind::Record => write!(f, "record"),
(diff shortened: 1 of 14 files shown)


bjchambers:
> Adds the sparrow trait for user defined functions. This will allow us to keep python dependencies in python crates, while passing

While passing *what*?

crates/sparrow-instructions/src/inst.rs
```diff
@@ -236,15 +232,43 @@ pub enum InstKind {
...
+            Self::Simple(arg0) => Self::Simple(arg0.clone()),
+            Self::FieldRef => Self::FieldRef,
+            Self::Cast(arg0) => Self::Cast(arg0.clone()),
+            Self::Record => Self::Record,
+            Self::Udf(arg0) => Self::Udf(arg0.clone()),
+        }
+    }
+}
+
+impl PartialEq for InstKind {
```
jordanrfrazier, line 251: I can't derive `PartialEq` -- something about it moving `self` out. Can look at this together if you think we can get around that.

crates/sparrow-instructions/src/inst.rs
```diff
@@ -236,15 +232,43 @@ pub enum InstKind {
...
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Simple(l0), Self::Simple(r0)) => l0 == r0,
+            (Self::FieldRef, Self::FieldRef) => true,
+            (Self::Cast(l0), Self::Cast(r0)) => l0 == r0,
+            (Self::Record, Self::Record) => true,
+            (Self::Udf(l0), Self::Udf(r0)) => l0.eq(r0),
```
bjchambers, line 258: I think we should be able to write this `l0 == r0`

crates/sparrow-instructions/src/inst.rs
bjchambers: that's somewhat surprising -- I can go either way.

bjchambers (approved):
Some clippy errors you should fix.

PR #681, opened 2023-08-19
feat: udf node creation in dfg

jordanrfrazier:
A little bit more plumbing for UDFs. Adds ability for `python.Expr` to call into the `sparrow-session::Session::add_udf_to_dfg` method, which adds the udf directly to the dfg.

### DIFF
--- a/crates/sparrow-compiler/src/ast_to_dfg.rs
+++ b/crates/sparrow-compiler/src/ast_to_dfg.rs
@@ -96,6 +96,7 @@ pub(super) fn ast_to_dfg(
 }
 
 pub fn add_udf_to_dfg(
+    location: &Located<String>,
     udf: Arc<dyn Udf>,
     dfg: &mut Dfg,
     arguments: Resolved<Located<AstDfgRef>>,
@@ -105,13 +106,8 @@ pub fn add_udf_to_dfg(
     let argument_types = arguments.transform(|i| i.with_value(i.value_type().clone()));
     let signature = udf.signature();
 
-    // TODO: No locations anymore
-    let fake_location = Located::new("fake".to_owned(), Location::internal_str("fake"));
-
-    // TODO: Ask about what python handles for us in regards to type checking, what we can pull
-    // out of here re: diagnostics since they're not reported(?)
     let (instantiated_types, instantiated_result_type) =
-        match instantiate(&fake_location, &argument_types, signature) {
+        match instantiate(&location, &argument_types, signature) {
             Ok(result) => result,
             Err(diagnostic) => {
                 diagnostic.emit(diagnostics);
@@ -125,7 +121,7 @@ pub fn add_udf_to_dfg(
     let grouping = verify_same_partitioning(
         data_context,
         diagnostics,
-        &fake_location.with_value(fake_location.inner().as_str()),
+        &location.with_value(location.inner().as_str()),
         &arguments,
     )?;
 
@@ -149,7 +145,7 @@ pub fn add_udf_to_dfg(
 
     let time_domain_check = TimeDomainCheck::Compatible;
     let time_domain =
-        time_domain_check.check_args(fake_location.location(), diagnostics, &args, data_context)?;
+        time_domain_check.check_args(location.location(), diagnostics, &args, data_context)?;
 
     Ok(Arc::new(AstDfg::new(
         value,
@@ -157,7 +153,7 @@ pub fn add_udf_to_dfg(
         instantiated_result_type,
         grouping,
         time_domain,
-        fake_location.location().clone(),
+        location.location().clone(),
         None,
     )))
 }
--- a/crates/sparrow-compiler/src/ast_to_dfg.rs
+++ b/crates/sparrow-compiler/src/ast_to_dfg.rs
@@ -17,7 +17,7 @@ use itertools::{izip, Itertools};
 use record_ops_to_dfg::*;
 use smallvec::{smallvec, SmallVec};
 use sparrow_arrow::scalar_value::ScalarValue;
-use sparrow_instructions::CastEvaluator;
+use sparrow_instructions::{CastEvaluator, Udf};
 use sparrow_instructions::{GroupId, InstKind, InstOp};
 use sparrow_syntax::{
     Collection, ExprOp, FenlType, FormatDataType, LiteralValue, Located, Location, Resolved,
@@ -29,7 +29,7 @@ use crate::dfg::{Dfg, Expression, Operation};
 use crate::diagnostics::DiagnosticCode;
 use crate::time_domain::TimeDomain;
 use crate::types::inference::instantiate;
-use crate::{DataContext, DiagnosticBuilder, DiagnosticCollector};
+use crate::{DataContext, DiagnosticBuilder, DiagnosticCollector, TimeDomainCheck};
 
 /// Convert the `expr` to corresponding DFG nodes.
 pub(super) fn ast_to_dfg(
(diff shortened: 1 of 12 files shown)


crates/sparrow-compiler/src/ast_to_dfg.rs
```diff
@@ -95,6 +95,73 @@ pub(super) fn ast_to_dfg(
...
+    dfg: &mut Dfg,
+    arguments: Resolved<Located<AstDfgRef>>,
+    data_context: &mut DataContext,
+    diagnostics: &mut DiagnosticCollector<'_>,
+) -> anyhow::Result<AstDfgRef> {
+    let argument_types = arguments.transform(|i| i.with_value(i.value_type().clone()));
+    let signature = udf.signature();
+
+    // TODO: No locations anymore
+    let fake_location = Located::new("fake".to_owned(), Location::internal_str("fake"));
```
jordanrfrazier, line 109: Okay for now?

crates/sparrow-compiler/src/ast_to_dfg.rs
```diff
@@ -95,6 +95,73 @@ pub(super) fn ast_to_dfg(
     )
 }
 
+pub fn add_udf_to_dfg(
```
jordanrfrazier, line 98: This code is duplicated from the `function` call, but a lot of pieces from that are unnecessary given that this is exclusively called from python client, meaning args are already flattened and locations not applicable. So, it can probably be refactored to share some methods but not worth the redirection imo.

crates/sparrow-compiler/src/types/instruction.rs
```diff
@@ -33,6 +32,16 @@ pub(crate) fn typecheck_inst(
 
             validate_instantiation(&argument_types, signature)
         }
+        InstKind::Udf(udf) => {
+            let signature = udf.signature();
+            let argument_types = Resolved::new(
+                Cow::Owned(signature.parameters().names().to_owned()),
```
jordanrfrazier, line 38: I don't think I can use `Borrowed` because `udf` is an `Arc`. I thought I would've been able to clone the `Arc`, then use borrowed from that clone's signature, but I was getting `udf not static lifetime`.

crates/sparrow-session/src/session.rs
```diff
@@ -277,6 +277,58 @@ impl Session {
...
+    #[allow(unused)]
+    fn add_udf_to_dfg(
+        &mut self,
+        udf: Arc<dyn Udf>,
+        args: Vec<Expr>,
+    ) -> error_stack::Result<AstDfgRef, Error> {
+        let signature = udf.signature();
+        // TODO: Why do I have to do this?
+        // Why lifetime on arc now happy?
+        let arg_names = signature.arg_names().to_owned();
```
jordanrfrazier, line 293: Same problem on lifetimes for this

crates/sparrow-compiler/src/types/instruction.rs
jordanrfrazier: Actually..I guess that makes sense? `Resolved` requires 
```
        names: Cow<'static, [Located<String>]>,
```
so even if I clone the `Arc`, it is dereferenced at the end of this function, so there's no guarantee the `names()` lives longer than this function call?

crates/sparrow-compiler/src/ast_to_dfg.rs
bjchambers: `Located::builder()` is used elsewhere, and exists for this case. Possibly pass the location in so that the session code is what decides to use that "fake" location?

crates/sparrow-compiler/src/ast_to_dfg.rs
```diff
@@ -95,6 +95,73 @@ pub(super) fn ast_to_dfg(
...
+    diagnostics: &mut DiagnosticCollector<'_>,
+) -> anyhow::Result<AstDfgRef> {
+    let argument_types = arguments.transform(|i| i.with_value(i.value_type().clone()));
+    let signature = udf.signature();
+
+    // TODO: No locations anymore
+    let fake_location = Located::new("fake".to_owned(), Location::internal_str("fake"));
+
+    // TODO: Ask about what python handles for us in regards to type checking, what we can pull
+    // out of here re: diagnostics since they're not reported(?)
```
bjchambers, line 112: We currently have to put things in diagnostics in order for the python code to actually report an error. Ideally, this would get much simplified, and just return the (error or errors) from the builder method without first creating the diagnostic and then extracting the message.

crates/sparrow-compiler/src/dfg.rs
```diff
@@ -125,6 +125,14 @@ impl Dfg {
         self.add_expression(Expression::Inst(InstKind::Simple(instruction)), children)
     }
 
+    pub(super) fn add_udf(
```
bjchambers, line 128: Do we need this? Should `ast_to_dfg` just call `add_expression` directly?

crates/sparrow-compiler/src/types/instruction.rs
bjchambers: Yeah -- it would require knowing some lifetime (of the arc) and then using `Cow<'arc, ...>`.

I think if we continue down this road, the better option be for this to just expect `Arc<Signature>` or `Arc<ThingThatHasSignature>` and use the signature from behind the arc.This is where `ArcRef` comes up... but that has issues. 

The other option which may be better is if this doesn't need to return an owned signatuer. If we could instead have:

```
fn signature(&self) -> &Signature { ... }
```

then we can just use the `Arc` that we have to get the signature, and be ok

crates/sparrow-session/src/session.rs
bjchambers: see previous -- we can discuss further.

it should be possible to resolve, possibly by allowing the `Resolved` to take a lifetime? not sure if we want to fix now or just really tear into this later

bjchambers (approved):

crates/sparrow-compiler/src/dfg.rs
jordanrfrazier: We could, but I like keeping it. It abstracts away the need for the caller to know the types of the `Expression` and `InstKind` in order to create the UDF. Similar to above, in `add_instruction`:

```
    pub(super) fn add_instruction(
        &mut self,
        instruction: InstOp,
        children: ChildrenVec,
    ) -> anyhow::Result<Id> {
        self.add_expression(Expression::Inst(InstKind::Simple(instruction)), children)
    }
```

crates/sparrow-instructions/src/inst.rs
```diff
@@ -261,11 +262,12 @@ impl PartialEq for InstKind {
     }
 }
 
-use std::hash::Hash;
 impl Hash for InstKind {
```
bjchambers, line 265: Should include discriminant first. This is the macro expansion:

```
impl core::hash::Hash for InstKind {
    fn hash<H: core::hash::Hasher>(&self, ra_expand_state: &mut H) {
        core::mem::discriminant(self).hash(ra_expand_state);
        match self {
            InstKind::Simple(f0) => {
                f0.hash(ra_expand_state);
            }
            InstKind::FieldRef => {}
            InstKind::Cast(f0) => {
                f0.hash(ra_expand_state);
            }
            InstKind::Record => {}
            InstKind::Udf(f0) => {
                f0.hash(ra_expand_state);
            }
        }
    }
}
```

Consider something like:
```
enum Either {
  A(i64),
  B(i64),
}
```

You would likely want `A(5)` to hash differently from `B(5)`. You do that by including the discriminant.

PR #682, opened 2023-08-21
added test of else method using differently ordered columns

epinzur:
i was having trouble shortening this test, while still being able to reproduce the bug.  

feel free to shorten if it still reproduces the issue.

note that I haven't fixed the issue yet. only added this failing test.


python/pytests/golden/else_test/test_else_unordered_record.jsonl
```diff
@@ -0,0 +1,4 @@
+{"_time":"1970-01-01T00:00:00.000000001","_subsort":0,"_key_hash":17095134351192101601,"_key":"dev","joined":{"text":"Thread 1","user":"UCZ4","time":1,"thread_ts":1.0,"key":"dev"},"threads":{"text":"Thread 1","user":"UCZ4","time":1.0,"thread_ts":1.0,"key":"dev"},"non_threads":null}
```
kevinjnguyen, line 1: Can you elaborate on what is the expected results here? These appear to be correct. If `threads` is null then `non_threads` is used as the value of `joined`.

epinzur: the expected results are correct.  the test currently fails.  https://github.com/kaskada-ai/kaskada/actions/runs/5926509057/job/16068060314?pr=682#step:6:211

epinzur: i made you a failing test.

python/pytests/else_test.py
```diff
@@ -63,3 +63,36 @@ def test_else_debug(record_source, golden) -> None:
             }
         )
     )
+
+
+@pytest.fixture(scope="module")
+def record_source_slack() -> kd.sources.JsonlString:
+    content = "\n".join(
+        [
+            """{"text":"Thread 1","user":"UCZ4","time":1,"thread_ts":1,"key":"dev"}""",
```
bjchambers, line 72: Shouldn't use `"""..."""` here -- those are "docstrings". Lint will likely suggest changing it.

PR #683, opened 2023-08-21
Edit the front page a bit.

kerinin:
This is intended to improve alignment with the current "PMF Theory".


python/docs/source/index.md
```diff
@@ -3,94 +3,119 @@ hide-toc: true
...
 ---
+# Introduction
 
 <div class="px-4 py-5 my-5 text-center">
     <img class="d-block mx-auto mb-4" src="_static/kaskada.svg" alt="" width="auto">
-    <h1 class="display-5 fw-bold">Event-processing for AI applications.</h1>
+    <h1 class="display-5 fw-bold">Real-Time AI without the fuss.</h1>
     <div class="col-lg-7 mx-auto">
-      <p class="lead mb-4">Next-generation, real-time and historic event processing.
+      <p class="lead mb-4">Kaskada is a simple streaming engine that connects AI models to real-time & historical data.
```
bjchambers, line 12: Any risk that "simple" is received as "limited"?

Overall, I think emphasizing the use for Real-Time AI is good, but I worry that this loses the explanation of what it is (event-processing / temporal queries / compute / etc.)

python/docs/source/index.md
```diff
@@ -3,94 +3,119 @@ hide-toc: true
...
       </p>
     </div>
 </div>
 
+(what-it-provides)=
```
bjchambers, line 17: Unless you're relying on these links, you shouldn't need the `(anchor-name)=` on headings. Sphinx will automatically generate an Anchor based on the heading for linking to (and in fact, this will add a second anchor, that won't show up automatically while scrolling). Probably only necessary if you want to link to it and be able to change the heading without having the link stop working (but sphinx will complain about internal links if the heading isn't there).

python/docs/source/index.md
```diff
@@ -3,94 +3,119 @@ hide-toc: true
...
 
-Kaskada goes beyond the columnar representation, by introduce a Timestream -- a columnar representation of events, ordered by time and grouped by key.
-This representation is a perfect fit for all kinds of events, modern event streams as well as events stored in a database.
-Specializing for Timestreams allows Kaskada to optimize temporal queries and execute them much faster.
+(real-time-ai)=
+### Real-time AI in minutes
 
-(python)=
-# Python-native
+Connect and compute from streaming data _or_ databases, and dynamically load data using Python.
```
bjchambers, line 35: Connect and compute over databases, streaming data, _and_ data loaded dynamically using Python.

python/docs/source/index.md
```diff
@@ -3,94 +3,119 @@ hide-toc: true
...
-Kaskada lets you do it all in one place.
+(scalable-and-reliable)=
+### Built for scale and reliability
 
-(get_started)=
-# Get Started
-
-With no infrastructure to deploy, get started processing events immediately.
-Check out the [Quick Start](quickstart) now!
+Implemented in [Rust](https://www.rust-lang.org/) using [Apache Arrow](https://arrow.apache.org/), Kaskada's compute engine uses columnar data to efficiently execute large historic or streaming queries.
```
bjchambers, line 44: "large historic and high-throughput streaming queries"?

python/docs/source/index.md
```diff
@@ -3,94 +3,119 @@ hide-toc: true
...
+Every operation in Kaskada is implemented incrementally, allowing automatic recovery if the process is terminated or killed.
 
-(local-and-distributed)=
-# Local, Remote and Distributed
+Kaskada makes most jobs fast enough to run locally, so it's easy to build and test your real-time queries.
+As your needs grow, Kaskada's cloud-native design and support for partitioned exeuction gives you the volume and throughput you need to scale.
```
bjchambers, line 48: "exeuction" -> "execution"

python/docs/source/index.md
```diff
@@ -3,94 +3,119 @@ hide-toc: true
...
+)
+
+# Send each Slack message to Kaskada
+def handle_message(client, req):
+    messages.add_rows(req.payload["event"])
+slack.socket_mode_request_listeners.append(handle_message)
+slack.connect()
+
+# Aggregate multiple messages into a "conversation"
+conversations = messages \
```
bjchambers, line 79: You can use parentheses if you want to omit the slashes:

```
conversations = (messages
  .select("user", "text")
  .collect(max=20)
)
```

python/docs/source/index.md
kerinin: Changed "simple" to "next-generation".

python/docs/source/index.md
kerinin: Cool - wasn't sure if they were important

python/docs/source/index.md
```diff
@@ -3,94 +3,119 @@ hide-toc: true
...
-- header: "{fas}`fast-forward;pst-color-primary` Real-time, Batch and Streaming"
-  content: "Execute large-historic queries or materialize in real-time. Or both."
-  link: ".#real-time-and-historic"
-- header: "{fas}`rocket;pst-color-primary` Local, Remote and Distributed"
-  content: "Develop and test locally. Deploy to Docker, K8s or a service for production."
-  link: ".#local-and-distributed"
-- header: "{fas}`backward;pst-color-primary` Time-travel"
-  content: "Generate training examples from the past to predict the future."
-  link: ".#time-travel"
+- header: "{fas}`timeline;pst-color-primary` Stream Processing"
```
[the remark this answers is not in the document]
kerinin: (1): I experimented with some variations along these lines, but I feel like "features" is too data-science-specific, and "prompts" or "contexts" also feel like jargon.  What do you think of "Real-time Aggregation"? It's a more general term than the other three, and also communicates that we're doing more than just event-level mapping.

(2): Agreed

(3): My concern is that we haven't found a good catchphrase for this. "Time-travel" is all over the place and doesn't really mean what we mean. "Historic computation" is better, but still seems sort of vague. What I like about "historic replay" is that it's describing the process and I feel like it does a better job suggesting the potential applications.

I asked ChatGPT what it thought the two terms meant:

> "Historic computation" and "historic replay" seem to be related terms, but they likely have distinct meanings based on their usage. Here's how they might differ:
> 
> 1. **Historic Computation**:
>    - This term could refer to the process of performing computations or calculations using methods and techniques that were prevalent in the past. It might involve using historical computing devices, algorithms, and tools that were available before the modern digital era.
>    - Historic computation could also encompass the study of how computations were done in the past, including the evolution of mathematical and computational techniques.
> 
> 2. **Historic Replay**:
>    - This term could refer to the act of replaying or recreating historical events, scenarios, or processes using computational models, simulations, or data analysis techniques.
>    - Historic replay might involve using available historical data to simulate and visualize past events, allowing researchers to gain insights into the dynamics and outcomes of those events.
>    - In certain contexts, historic replay could also refer to the process of reviewing and analyzing historical records, logs, or data to reconstruct past sequences of events, often in fields like computer science, where system behavior needs to be understood retrospectively.
> 
> In summary, while both terms involve the concept of history and computation, "historic computation" seems more focused on the computation methods of the past, while "historic replay" appears to center on recreating or analyzing historical events using computational techniques. The exact meanings would depend on the specific context in which these terms are used.

python/docs/source/index.md
kerinin: Works for me - we should probably standardize on one way or another

python/docs/source/index.md
```diff
@@ -3,94 +3,116 @@ hide-toc: true
...
-- header: "{fas}`rocket;pst-color-primary` Local, Remote and Distributed"
-  content: "Develop and test locally. Deploy to Docker, K8s or a service for production."
-  link: ".#local-and-distributed"
-- header: "{fas}`backward;pst-color-primary` Time-travel"
-  content: "Generate training examples from the past to predict the future."
-  link: ".#time-travel"
+- header: "{fas}`timeline;pst-color-primary` Real-time Aggregation"
+  content: "Precompute model inputs from streaming data with robust data connectors, transformations & aggregations."
+- header: "{fas}`binoculars;pst-color-primary` Event Detection"
+  content: "Trigger pro-active AI behaviors by identifying significant changes in your data, when they happen."
```
bjchambers, line 25: Significant events may not be changes? Not sure what would be better though ("occurrences"?)

python/docs/source/index.md
```diff
@@ -3,94 +3,116 @@ hide-toc: true
...
+- header: "{fas}`backward;pst-color-primary` History Replay"
```
bjchambers, line 26: What if this was "Replay & Backfill" or something like that.

python/docs/source/index.md
```diff
@@ -3,94 +3,116 @@ hide-toc: true
...
-(python)=
-# Python-native
+There's no infrastructure to provision (and no JVM hiding under the covers), so you can jump right in - check out the [Quick Start](quickstart).
 
-Connect to existing data in streams or databases, or load data using Python.
-Wherever your events are stored, Kaskada can help you process them.
 
-Build temporal queries and process the results using Python.
-Connect straight to your visualizations, dashboards or machine learning systems.
+(scalable-and-reliable)=
```
bjchambers, line 39: Maybe drop this anchor as well? Don't see it being used.

bjchambers (approved):

python/docs/source/index.md
kerinin: "Occurrences" feels wordy to me. I was going to say that all events are changes in your data, but I suppose maybe timers aren't data per se? We could go with "identifying significant events" - I didn't use that originally because it suggests that we can only react to input events.

python/docs/source/index.md
kerinin: doh

python/docs/source/index.md
kerinin: Going to try "Trigger pro-active AI behaviors by identifying important activies, as they happen". Starting to feel like word salad - think I've been looking at it too much.

PR #684, opened 2023-08-21
Add google tracking to Sphynx docs

kerinin:


python/docs/source/_templates/layout.html
```diff
@@ -0,0 +1,14 @@
+{% extends "!layout.html" %}
```
bjchambers, line 1: Should change the template -- the theme already supports this.

python/docs/source/conf.py
```diff
@@ -69,6 +69,7 @@
     "github_repo": "kaskada",
     "github_version": "main",
     "doc_path": "kaskada/docs/source",
+    "google_analytics": "G-HR9E2E6TG4",
```
bjchambers, line 72: https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/analytics.html

```
html_theme_options["analytics"] = {
    "google_analytics_id": "G-XXXXXXXXXX",
}
```

OR

```
html_theme_options = {
  ...
  "analytics": {
    "google_analytics_id": "...",
  }
}
```

bjchambers, line 72: We may also want this to be an environment variable, so that it is only on in the deployed docs?

kerinin: Let's hard code it for now - it's what we're doing in a couple other places.

bjchambers (approved):

PR #685, opened 2023-08-21
feat: Setup release process

bjchambers:
This should do the following:

1. Use `release-drafter` to maintain a draft of a `v###` release.
2. Build wheels with a version based on either the current state of the file or using the version from the tag.
3. When we publish that release, it should publish to PyPi.


.github/release-drafter-python-client.yml
```diff
@@ -0,0 +1,9 @@
+_extends: kaskada:.github/release-drafter.yml
```
bjchambers, line 1: Some renaming to "free up" the name "python" for now.

I think in the next week or two it would be good if we stopped drafting releases for the python client and engine, and just had the single "kaskada release draft", but we can revisit once we get this out.

.github/release-drafter-python.yml
```diff
@@ -1,9 +1,5 @@
...
-tag-template: python@v$RESOLVED_VERSION
-tag-prefix: python@v
-
-# Only include PRs with one of these labels
-include-labels:
-  - python
\ No newline at end of file
+name-template: Kaskada $RESOLVED_VERSION-a.0
+tag-template: v$RESOLVED_VERSION-a.0
+tag-prefix: v
```
bjchambers, line 5: Since there are no `include-labels` specified, this will include all PRs in the repo.

.github/workflows/ci_python.yml
```diff
@@ -43,6 +46,7 @@ jobs:
   #
   # Also, only do it on one machine.
   lint:
+    if: github.event_name == 'pull_request' || github.event_name == 'merge_group'
```
bjchambers, line 49: Only lint on pull requests and merge queue.

.github/workflows/ci_python.yml
```diff
@@ -75,6 +79,10 @@ jobs:
           poetry run pydocstyle pysrc
 
   debug:
+    if: |
```
bjchambers, line 82: Only debug on pull requests, merge queue, and pushes to main (since it is needed for docs).

.github/workflows/ci_python.yml
```diff
@@ -135,11 +143,42 @@ jobs:
           # Automatically uploads an artifact from the './_site' directory by default
           path: ${{ github.workspace }}/python/docs/_build
 
+  version:
```
bjchambers, line 146: New job (version) to extract the version we're building. This is mostly used for releasing, since we need to update the file and put it in the wheels we build.

.github/workflows/ci_python.yml
```diff
@@ -345,20 +420,23 @@ jobs:
...
+    if: github.event_name == 'release' && github.event.action == 'published'
+    steps:
+      - uses: actions/download-artifact@v3
+        with:
+          # unpacks default artifact into dist/
+          # if `name: artifact` is omitted, the action will create extra parent dir
+          name: artifact
+          path: dist
+
+      # - uses: pypa/gh-action-pypi-publish@release/v1
```
bjchambers, line 439: Currently, I have the actual publishing disabled. I'll probably add it once I get things working.

.github/workflows/ci_python.yml
```diff
@@ -345,20 +420,23 @@ jobs:
...
-  #       env:
-  #         MATURIN_PYPI_TOKEN: ${{ secrets.PYPI_API_TOKEN }}
-  #       with:
-  #         command: upload
-  #         args: --skip-existing *
-  #         working-directory: python
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    environment: pypi
```
bjchambers, line 426: This should use a GitHub deploy environment. I think that means we need to setup an environment. The main benefit is that since it's considered a "deployment" we can configure manual approval if we want, etc.

.github/workflows/ci_python.yml
```diff
@@ -345,20 +420,23 @@ jobs:
...
+        with:
+          # unpacks default artifact into dist/
+          # if `name: artifact` is omitted, the action will create extra parent dir
+          name: artifact
+          path: dist
+
+      # - uses: pypa/gh-action-pypi-publish@release/v1
+        # To Test
+        #with:
+          # To test: repository_url: https://test.pypi.org/legacy/
```
bjchambers, line 442: I think after the push, we'd ideally have this either create a PR or make a commit that updates the local version to something like "<next_patch>-dev` or something like that, so the wheels built during development reflect the *next* version. But, I also don't think that's strictly necessary yet. Thoughts?

.github/workflows/ci_python.yml
epinzur: linting on commit was helpful for me today... maybe don't add this?

.github/workflows/ci_python.yml
epinzur: nice but not required.

epinzur (approved):

.github/workflows/ci_python.yml
bjchambers: It will still happen on commit -- that's part of the `pull_request` event (eg., when you update the pull request it will run).

Issue #697, opened 2023-08-22
Get existing ticks and window functionality available


bjchambers:
Thought for how we can “hack” ticks for now.

Problem: We previously relied on re-parsing the AST and binding an “input” so they could implicitly tick over that. That won’t work with the new Python builders.

Idea: Allow things (especially windows) to take expressions that need the “input” as a lambda.

```
@dataclass(frozen=True)
class Since(Window):
    predicate: ... | Callable[Timestream, Timestream]
Then, define the ticks as things like:
def hourly(input: Timestream) -> Timestream:
  ...
```

Then aggregations like `sum(window=Since(hourly))` should work by adding a little code to the `_aggregation` helper. Specifically, for each kind of window, if the predicate is a callable they can pass the input to the function to generate the domain to tick over.
It’s not super pretty, but it doesn’t seem horrible. And the ability to configure things referencing the input is generally useful, even if/when we fix ticks.
Thoughts?

bjchambers:
An alternative would be to expose these more like the windows we think we'd like to have. For instance, we could expose:

```
e.sum(window=kd.windows.hourly())

e.sum(window=kd.windows.sliding(3, kd.windows.hourly()))

etc.
```

This would involve some changes in the Python side, but wouldn't differ significantly in terms of complexity. It could be a way to prototype some of the ideas for the windowing API, especially if we could use some hacks in the Python side to provide "windows with a value at the end of the window" based on the tick columns.

@jordanrfrazier @kerinin any thoughts on this?

PR #702, opened 2023-08-22
feat: Expose tick functions

bjchambers:
This introduces methods to `Timestream` to create tick-streams, and allows passing a function to construct such a tickstream as the predicate for since and sliding windows.


crates/sparrow-compiler/src/ast_to_dfg.rs
```diff
@@ -242,20 +243,7 @@ pub fn add_to_dfg(
...
-                        tick_node,
-                        tick_node,
-                        FenlType::Concrete(DataType::Boolean),
-                        agg_input.grouping(),
-                        agg_input.time_domain().clone(),
-                        function_name.location().clone(),
-                        None,
-                    ));
-                    return Ok(tick_node);
+                    return add_tick(dfg, behavior, agg_input.value());
```
jordanrfrazier, line 246: Ah okay -- so this is still valid for coming from fenl, doing the weird condition input binding, but through python we pass the input directly. Nice.

jordanrfrazier:
This looks like a good pattern

bjchambers:
@jordanrfrazier 

> This looks like a good pattern

Yeah -- I think this makes sense for the Rust side. As noted in #697, we can discuss encapsulating the ticks themselves in the windows:

```
e.sum(window = kd.windows.hourly())
e.sum(window = kd.windows.Session(3, kd.windows.hourly()))
```

as a way of reducing the number of concepts (no such thing as ticks are exposed) and a way to let us experiment with richer windows. For instance, I *think* we can define both `since hourly` and `outputting once at the end of the hour` nearly in Python (using some of the shift-by-1-nanosecond style tricks), and it would let us experiment with the way those concepts fit together over time.

python/pysrc/kaskada/_timestream.py
```diff
@@ -1463,9 +1463,15 @@ def _aggregation(
...
+        predicate = window.predicate
+        if callable(predicate):
+            predicate = predicate(input)
+        return Timestream._call(op, input, *args, predicate, None)
     elif isinstance(window, kd.windows.Sliding):
-        return Timestream._call(op, input, *args, window.predicate, window.duration)
+        predicate = window.predicate
+        if callable(predicate):
+            predicate = predicate(input)
+        return Timestream._call(op, input, *args, predicate, window.duration)
```
jordanrfrazier, line 1474: We'll probably have to do something similar to this if we want to support ticks in other functions like `filter/if`. I could work on that as a follow-up to this

jordanrfrazier (approved):
Looks good, the lambda is really clever. I still think we'll want to go the route of offering explicit parameters to the windows themselves, like `tumbling(units, duration, offset)`, which I suppose we could do in a builder pattern 
```
kd.window.Tumbling().duration(5).units(kd.units.Second)
```

bjchambers:
> Looks good, the lambda is really clever. I still think we'll want to go the route of offering explicit parameters to the windows themselves, like `tumbling(units, duration, offset)`, which I suppose we could do in a builder pattern
> 
> ```
> kd.window.Tumbling().duration(5).units(kd.units.Second)
> ```

I think if we go that route, I'd be more inclined to do `kd.window.Tumbling(seconds = 5)` or taking a timedelta, etc. Specifically, we shouldn't define units ourselves.

PR #703, opened 2023-08-23
feat: python udf implementation 

### LINKED ISSUE #698: Python UDF

jordanrfrazier:
Adds the ability for users to define Python user-defined-functions that operate on Pandas Series, and interoperate with Timestreams. For example,

```
@kd.udf("add<N: number>(x: N, y: N) -> N")
def add(x: pd.Series, y: pd.Series) -> pd.Series:
    x + y
```

Which can be called simply as:
```
add(Foo.m, Foo.n)
```

There are some considerations:
1. Only allows for operations on the `Series` and not per-element. 
2. Requires knowing Fenl syntax/types to place the annotation. 

However, improvements can be made where only the return type is supplied, and the argument types checked at runtime. This is similar to how PySpark handles it. 

Polars offers another option, where they split udf calls into `apply` (working per-element) and `map` (over arrays). It's noted that the per-element operations are slow, which is expected. 

Closes #698


crates/sparrow-runtime/src/execute.rs
```diff
@@ -21,6 +21,7 @@ use crate::execute::output::Destination;
 use crate::key_hash_inverse::{KeyHashInverse, ThreadSafeKeyHashInverse};
 use crate::stores::ObjectStoreRegistry;
 use crate::RuntimeOptions;
+pub use operation::WorkArea;
```
jordanrfrazier, line 24: remove

crates/sparrow-runtime/src/execute/operation.rs
```diff
@@ -66,6 +66,7 @@ use crate::execute::Error;
 use crate::key_hash_inverse::ThreadSafeKeyHashInverse;
 use crate::stores::ObjectStoreRegistry;
 use crate::Batch;
+pub use expression_executor::WorkArea;
```
jordanrfrazier, line 69: remove

crates/sparrow-session/src/lib.rs
```diff
@@ -9,3 +9,5 @@ pub use execution::Execution;
 pub use expr::{Expr, Literal};
 pub use session::{ExecutionOptions, Session};
 pub use table::Table;
+
+// pub type Udf = dyn sparrow_instructions::Udf;
```
jordanrfrazier, line 13: The type alias was not allowed when implementing the trait. Perhaps was doing it incorrectly, though

python/pysrc/kaskada/__init__.py
```diff
@@ -21,5 +22,6 @@
     "Result",
     "sources",
     "Timestream",
+    "udf",
     "windows",
 ]
```
jordanrfrazier, line 27: I'm not quite 100% on what _needs_ to be here or not

python/pysrc/kaskada/_timestream.py
```diff
@@ -4,6 +4,8 @@
 
 import sys
 import warnings
+import inspect
```
jordanrfrazier, line 7: remove

python/pysrc/kaskada/_timestream.py
```diff
@@ -65,6 +67,63 @@ def _literal(value: Literal, session: _ffi.Session) -> Timestream:
...
         return Timestream(_ffi.Expr.literal(session, value))
 
+    @staticmethod
+    def _call_udf(
+        udf: _ffi.Udf,
+        *args: Union[Timestream, Literal],
+        session: Optional[_ffi.Session] = None,
+    ) -> Timestream:
+        """
+        Construct a new Timestream by calling the given function.
```
jordanrfrazier, line 77: update docs

python/pysrc/kaskada/_timestream.py
```diff
@@ -199,6 +257,16 @@ def pipe(
         else:
             return func(self, *args, **kwargs)
 
+# TODO: How does udf work? We kind of want it like... apply some function on an input?
```
jordanrfrazier, line 260: remove

python/src/udf.rs
```diff
@@ -1,32 +1,131 @@
...
+}
+
+struct PyUdfEvaluator {
+    args: Vec<ValueRef>,
+    callable: Py<PyAny>,
+}
+
+#[derive(Clone)]
+#[pyclass]
+// TODO: Does it need to extend Expr for some reason?
```
jordanrfrazier, line 28: I ask this because I don't understand why `Table` is a subclass of `Expr`

bjchambers: I think table did because we wanted to allow every source to be used as an expression. We don't *need* to model that in the Rust code -- the `source` could hold both... but at the least, we needed the sources to extend `Timestream`. I'm not sure if it would be clearer to have them separate in the Rust code -- we could say:

* `Timestream has the _ffi.Expr (and/or rename that to Timestream)`
* `Source extends Timestream (so it has an _ffi.Expr) and it has a _ffi.Table (which we should rename to Source)`.

The nice thing if we split them like that is that the Rust code doesn't deal with subtyping, but also it keeps the separation between expressions and sources more distinct in rust.

crates/sparrow-compiler/src/ast_to_dfg.rs
```diff
@@ -103,6 +103,7 @@ pub fn add_udf_to_dfg(
     data_context: &mut DataContext,
     diagnostics: &mut DiagnosticCollector<'_>,
 ) -> anyhow::Result<AstDfgRef> {
+    println!("sparrow_compiler::add_udf_to_dfg");
```
bjchambers, line 106: printlns (won't comment on them in rest of PR)

crates/sparrow-session/src/lib.rs
bjchambers: Ah, yeah -- the reason is that it changes `impl Udf (the trait) for ...` into `imp dyn Udf for ...`... but you can only implement traits.

crates/sparrow-session/src/session.rs
```diff
@@ -25,6 +25,13 @@ pub struct Session {
     data_context: DataContext,
     dfg: Dfg,
     key_hash_inverse: HashMap<GroupId, Arc<ThreadSafeKeyHashInverse>>,
+    /// Keeps track of the uuid mapping.
+    ///
+    /// This is a bit of a hack that allows to serialize the uuid instead of the
+    /// `Arc<dyn Udf` into the compute plan, then look up that mapping during plan
+    /// construction. Once we run on multiple machines we'll need to serialize the
+    /// udf.
+    udfs: HashMap<Uuid, Arc<dyn Udf>>,
```
bjchambers, line 34: Even when we do, I'd expect us to serialize a map from trait to UDF -- specifically, the same UDF may appear multiple times, and we wouldn't want to pickle it each time it occurs. Using the UUID for indirection is a reasonable pattern for serializations.

python/Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
```
bjchambers, line 1: what is this file for?

python/pysrc/kaskada/__init__.py
bjchambers: This is the list of "public" things that are available as `kd.<name>` when you `import kaskada as kd`. So, the `udf` annotatino probably needs to be here at this point. Not sure the `Udf` needs to, if the user isn't implementing the class directly yet.

python/pysrc/kaskada/_timestream.py
```diff
@@ -65,6 +67,63 @@ def _literal(value: Literal, session: _ffi.Session) -> Timestream:
...
+            If the argument types are invalid for the given function.
+        ValueError
+            If the argument values are invalid for the given function.
+        """
+        if session is None:
+            session = next(
+                arg._ffi_expr.session() for arg in args if isinstance(arg, Timestream)
+            )
+
+        def make_arg(arg: Union[Timestream, Literal]) -> _ffi.Expr:
```
bjchambers, line 108: I wonder if this can either use more of the code from `Expr.call`? Isn't this the same as `call` but the name is now `str | Udf`?

python/pysrc/kaskada/_timestream.py
bjchambers: I think the udf should work as a function, so `my_udf(a, b)` should work. At which point, `a.pipe(my_udf, b)` should automatically work. You shouldn't need to change `pipe` to do that -- at that point, the `Udf` is a callable so it can be applied as the `func`, which `pipe` already does.

python/pysrc/kaskada/udf.py
```diff
@@ -1,57 +1,54 @@
...
-        if isinstance(pd_result, pd.Series):
-            return pa.Array.from_pandas(pd_result, type=result_type)
-        else:
-            raise TypeError(f"Unsupported result type: {type(pd_result)}")
+        self._ffi = _ffi.Udf(signature, func)
 
+    def __call__(self, *args: Timestream | Literal) -> Timestream:
+        """Apply the UDF to the given arguments."""
+        print("Calling udf")
+        return Timestream._call_udf(self._ffi, *args)
```
bjchambers, line 26: I think this can just use `Timestream._call(self, *args)`, then in timestream you can change the `name: str` parameter to be `func: str | Udf`, and then do a `if isinstance(Udf, func)`

python/src/udf.rs
```diff
@@ -1,32 +1,134 @@
...
+        //     .collect::<anyhow::Result<_>>()?;
+
+        // let locals = [("args", args)].into_py_dict(py);
+        // py.run(&self.code, None, Some(&locals))?;
+        // let result_py: PyObject = locals.get_item("ret").unwrap().into();
+
+        // let result_array_data: ArrayData = ArrayData::from_pyarrow(result_py.as_ref(py))?;
+
+        // // We can't control the returned type, but we can refuse to accept the "wrong" type.
+        // ensure!(
```
bjchambers, line 96: We can also pass the expected type to Python and do some of the coercion in Python. That is sometimes beneficial -- consider we get a "numeric" Pandas array. If we say "convert to PyArrow" it will, but it may pick `f64`, `i64` or whatever type it wants. If we instead say "convert to a PyArrow array of type `i32`", it will try to do that.

This is why the UDF pattern had "calling" pass the expected type to Python, since it lets it try to satisfy the expectation. If we wait til here, the only thing we can do is fail if it isn't what we expect (or possibly cast).

python/src/execution.rs
```diff
@@ -54,13 +54,20 @@ impl Execution {}
...
+        let batches = py.allow_threads(move || {
+            execution.collect_all_blocking()
+        })?;
+
         let results = batches
             .into_iter()
             .map(|batch| batch.to_pyarrow(py))
             .collect::<PyResult<Vec<_>>>()?;
         Ok(results)
+        // result.map_err(|_| error_stack::report!(ErrorContext::Python).into())
```
jordanrfrazier, line 70: remove

crates/sparrow-runtime/src/execute.rs
```diff
@@ -326,5 +342,9 @@ pub async fn materialize(
     // TODO: the `execute_with_progress` method contains a lot of additional logic that is theoretically not needed,
     // as the materialization does not exit, and should not need to handle cleanup tasks that regular
     // queries do. We should likely refactor this to use a separate `materialize_with_progress` method.
-    execute_new(plan, destination, data_context, options, None).await
+
+    // TODO: Unimplemented feature - UDFs
+    let udfs = HashMap::new();
```
jordanrfrazier, line 347: This is for materializations

python/pysrc/kaskada/__init__.py
```diff
@@ -10,6 +10,8 @@
 from ._timestream import Literal
 from ._timestream import Timestream
 from ._timestream import record
+from .udf import udf
+from .udf import Udf
```
jordanrfrazier, line 14: Exposing the `Udf` class so we can test `isinstance(.., kd.Udf)`

python/pytests/udf_test.py
```diff
@@ -1,26 +1,63 @@
...
+    return (x + y) * 2
 
-    x = pa.array([1, 12, 17, 23, 28], type=pa.int8())
-    y = pa.array([1, 13, 18, 20, 4], type=pa.int8())
-    result = add.run_pyarrow(pa.int8(), x, y)
-    assert result == pa.array([2, 25, 35, 43, 32], type=pa.int8())
 
+@pytest.mark.skip(reason="currently not working")
+def test_docstring() -> None:
+    assert add.__doc__ == "Use Pandas to add two numbers."
```
jordanrfrazier, line 21: I haven't looked into this yet. Will look more tomorrow

python/src/udf.rs
jordanrfrazier: Okay, I think that makes sense. `Table` extends `Expr`, which is the object within `Timestream` currently. Yeah, if we need that separation, that would be a good idea.

python/src/udf.rs
jordanrfrazier: Cool -- I believe we are doing this in the `_converted_func` call, where we pass the `result_type` in now.

crates/sparrow-runtime/src/execute/operation/expression_executor.rs
```diff
@@ -95,15 +98,23 @@ impl ExpressionExecutor {
                 Operator::Instruction(inst) => {
                     // TODO: This could probably be cleaned up, possibly by eliminating
                     // `inst kind` entirely.
+                    println!("Inst: {:?}", inst);
```
jordanrfrazier, line 101: .

crates/sparrow-runtime/src/execute/operation.rs
```diff
@@ -175,6 +178,7 @@ impl OperationExecutor {
         max_event_time_tx: tokio::sync::mpsc::UnboundedSender<Timestamp>,
         late_bindings: &EnumMap<LateBoundValue, Option<ScalarValue>>,
         stop_signal_rx: Option<tokio::sync::watch::Receiver<bool>>,
+        udfs: &HashMap<Uuid, Arc<dyn Udf>>,
```
bjchambers, line 181: Can this be put in the `OperationContext` rather than adding an argument?

python/pysrc/kaskada/__init__.py
bjchambers: We shouldn't need to expose that externally -- the place where we do the `isinstance` is internal, so it can just import `Udf` from `udf`.

python/pysrc/kaskada/_timestream.py
```diff
@@ -111,10 +111,18 @@ def make_arg(arg: Union[Timestream, Literal]) -> _ffi.Expr:
...
         try:
-            return Timestream(
-                # TODO: FRAZ - so I need a `call` that can take the udf
-                _ffi.Expr.call(session=session, operation=func, args=ffi_args)
-            )
+            if isinstance(func, str):
+                return Timestream(
+                    _ffi.Expr.call(session=session, operation=func, args=ffi_args)
+                )
+            elif isinstance(func, _ffi.Udf):
```
bjchambers, line 118: Thoughts on whether this should be `_ffi.Udf` or `Udf`?

python/pysrc/kaskada/udf.py
```diff
@@ -1,57 +1,63 @@
...
+    """User defined function."""
 
-        signature : str
-            The Kaskada function signature for this UDF. Will be used to check
-            types of calls to this function and propagate type information for
-            the rest of the query.
-        """
+    def __init__(self, signature: str, func: Callable[..., pa.Array]) -> None:
+        """Create a UDF with the given signature."""
         functools.update_wrapper(self, func)
```
bjchambers, line 24: I suspect this is why you're not getting the docstring through -- I think you need to pass the original function, not the wrapped function -- this basically tells `functools` that "self is a wrapper around `func`", but that's the lambda we made.

You can probably move this out of the `__init__` and call `functools.update_wrapper` below.

python/pysrc/kaskada/udf.py
```diff
@@ -1,57 +1,63 @@
...
-            raise TypeError(f"Unsupported result type: {type(pd_result)}")
 
-
-def fenl_udf(name: str, signature: str):
+def udf(signature: str):
     """Decorate a function for use as a Kaskada UDF."""
 
-    def decorator(func: FuncType):
-        return Udf(name, func, signature)
+    def decorator(func_type: FuncType):
```
bjchambers, line 35: It's not really a `func_type` -- maybe `original_func` or `user_func` or something?

python/pysrc/kaskada/udf.py
```diff
@@ -1,57 +1,63 @@
...
+        # 1. Convert the `FuncType` to the type expected by Udf.
+        # This needs to take the PyArrow result type and PyArrow inputs,
+        # convert them to Pandas, call the function, and convert the result
+        # back to PyArrow.
+        def func(result_type, *args):
+            return _converted_func(func_type, result_type, *args)
+
+        # 2. Create the UDF object.
+        return Udf(signature, func)
```
bjchambers, line 44: For functools, consider:

```
udf = Udf(signature, func)
functools.update_wrapper(udf, func_type)
```

That way you connect the original function (with any docs) to the UDF.

python/src/expr.rs
```diff
@@ -44,6 +44,34 @@ impl Expr {
...
+            ));
+        }
+
+        let mut rust_session = session.rust_session()?;
+        let args: Vec<_> = args.into_iter().map(|e| e.rust_expr).collect();
+
+        let rust_expr = match rust_session.add_udf_to_dfg(sparrow_udf.clone(), args) {
+            Ok(node) => node,
+            Err(e) => {
+                // DO NOT SUBMIT: Better error handling.
```
bjchambers, line 66: .

bjchambers (approved):

crates/sparrow-runtime/src/execute/operation.rs
jordanrfrazier: Done

python/pysrc/kaskada/_timestream.py
jordanrfrazier: I'm not sure I understand the question, or maybe the distinction well enough. the `_ffi.Udf` class is the rust-built class available in python that we're operating on in the `_ffi.Expr` class. 

Where `Udf` is the python class that wraps the `_ffi.Udf`, allows a decorator annotation, and defines how to execute the callable. The `__call__` function is defined as 
```
    def __call__(self, *args: Timestream | Literal) -> Timestream:
        """Apply the UDF to the given arguments."""
        return Timestream._call(self._ffi, *args)

```

which passes in the `self._ffi`. This is the `_ffi.Udf`. Are you asking here whether we should pass in `self` instead, then get out the `_ffi.Udf` within `TImestream._call`?

PR #706, opened 2023-08-23
feat: Introduce crate for standalone signatures

bjchambers:
This provides a standalone signature parser (without location tracking) and instantiation. The intention is to provide the core type-checking component from the compiler in a stand-alone component for `sparrow-session` (and possibly checking logical/physical plans).

This doesn't use any fancy lexer/parser -- instead it just parses the string directly. This also doesn't use names in for the function, assuming that isn't necessary in the new direction. This also doesn't track locations, etc.


crates/sparrow-types/src/instantiate/type_constraints.rs
```diff
@@ -0,0 +1,169 @@
...
+        (UInt64, UInt64 | Float64) => true,
+        // Floats can be widened.
+        (Float16, Float16 | Float32 | Float64) => true,
+        (Float32, Float32 | Float64) => true,
+        (Float64, Float64) => true,
+        // Other promotions that we allow implicitly.
+        (Utf8, Timestamp(TimeUnit::Nanosecond, None)) => true,
+        // Other promotions must be explicitly requested.
+        (_, _) => false,
+    }
```
jordanrfrazier, line 168: This is missing a few casts that are supported from the sparrow-compiler's version. On purpose?

jordanrfrazier (approved):

jordanrfrazier:
Assuming that this was copied over and pruned of locations and unnecessary code. Let me know if I should go over specific files closer

bjchambers:
> Assuming that this was copied over and pruned of locations and unnecessary code. Let me know if I should go over specific files closer

Yep, that and converting to error_stack.

crates/sparrow-types/src/instantiate/type_constraints.rs
bjchambers: Unintentional -- probably reflects when I grabbed that code.

crates/sparrow-types/src/instantiate.rs
```diff
@@ -0,0 +1,125 @@
...
+    let solutions: HashMap<TypeVariable, DataType> =
+        type_constraints.solve(signature.type_parameters.iter())?;
+
+    // Instantiate the actual types for the signature.
+    let arguments = signature
+        .iter_parameters(arguments_len)?
+        .map(|parameter| {
+            parameter
+                .ty
+                .instantiate(Some(&solutions))
```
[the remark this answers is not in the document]
bjchambers: Currently, `instantiate` is used in two places:

1. When trying to get the "Arrow" type out, if possible. This needs to support both concrete primitives `FenlType::Concrete` and collections like `FenlType::List(FenlType::Concrete(...))`. We can do this by using `instantiate` (which extracts the `DataType` and passing no bindings). If it encounters a `Generic` / Type Variable, it will return `None` or fail.
2. When extracting the result, in which case it shouldn't fail.

I could see a case for making it an error, but the error is different in the two cases. In the first -- it means something went wrong and it's not concrete, like we had expected. In the second, it means something internal went wrong.

crates/sparrow-types/src/fenl_type.rs
```diff
@@ -0,0 +1,317 @@
...
+            DataType::Float16 => fmt.write_str("f16"),
+            DataType::Float32 => fmt.write_str("f32"),
+            DataType::Float64 => fmt.write_str("f64"),
+            DataType::Timestamp(unit, _) => fmt.write_str(match unit {
+                TimeUnit::Second => "timestamp_s",
+                TimeUnit::Millisecond => "timestamp_ms",
+                TimeUnit::Microsecond => "timestamp_us",
+                TimeUnit::Nanosecond => "timestamp_ns",
+            }),
+            DataType::Time32(_) => todo!(),
```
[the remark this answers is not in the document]
bjchambers: This type isn't currently supported by Kaskada so figured we should revisit separately.

crates/sparrow-types/src/instantiate/type_constraints.rs
bjchambers: Pulled in the latest.

PR #707, opened 2023-08-25
ref: Make parameters to sources explicit

bjchambers:
This improves the documentation and code-completion experience.


python/pysrc/kaskada/sources/arrow.py
```diff
@@ -11,13 +11,22 @@
...
 
 class Pandas(Source):
     """Source reading data from Pandas dataframe."""
 
     def __init__(
-        self, dataframe: pd.DataFrame, *, schema: Optional[pa.Schema] = None, **kwargs
+        self,
+        dataframe: pd.DataFrame,
+        *,
+        time_column_name: str,
```
bjchambers, line 24: `time_column_name` is long -- thoughts on `time_column` or `time_name`?

@kerinin @epinzur @jordanrfrazier

python/pysrc/kaskada/sources/source.py
```diff
@@ -3,13 +3,16 @@
...
+from typing import Literal
+from typing_extensions import TypeAlias
 
 from .._session import _get_session
 from .._timestream import Timestream
 
 
 _TABLE_NUM: int = 0
 
+TimeUnit: TypeAlias = Optional[Literal["ns", "us", "ms", "s"]]
```
jordanrfrazier, line 15: I think I'd prefer removing the `Optional` here and explicitly writing `Optional[TimeUnit]` where needed, so we can clearly see in the signature whether it's required. I acknowledge that it's generally got a default value to the parameters where it's used, so not a big deal.

jordanrfrazier (approved):

epinzur (approved):

python/pysrc/kaskada/sources/source.py
bjchambers: I like that. Making that change.

PR #709, opened 2023-08-28
version: update to 0.6.0-a.2

jordanrfrazier:


python/docs/source/examples/time_centric.ipynb
```diff
@@ -81,7 +81,7 @@
     "## Kaskada Client Setup\n",
     "\n",
     "```\n",
-    "%pip install kaskada>=0.6.0-a.1\n",
+    "%pip install kaskada>=0.6.0-a.2\n",
```
bjchambers, line 84: We shouldn't need to explicitly do this unless we want to -- I think `>=` will take the latest pre-release, which would be `0.6.0-a.2` once it's available.

python/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "kaskada"
-version = "0.6.0-a.1"
+version = "0.6.0-a.2"
```
bjchambers, line 3: There are two version numbers in this -- should update both.

Is there a reason we're bumping this?

PR #710, opened 2023-08-28
echo versions in python release

jordanrfrazier:


.github/workflows/release_python.yml
```diff
@@ -59,7 +59,9 @@ jobs:
       - name: Determine version (release)
         if: github.event_name == 'release'
         run: |
+          echo "GITHUB_REF_VERSION=${GITHUB_REF#refs/tags/v}" >> $GITHUB_OUTPUT
```
bjchambers, line 62: Would just echo -- this adds is at output for a later step.

```
echo "GITHUB_REF=${GITHUB_REF}"
VERSION=${GITHUB_REF#refs/tags/v}
echo "VERSION=${VERSION}"
```

bjchambers (approved):

PR #721, opened 2023-08-29
release: Install wheels by path

bjchambers:
Also clears the working-directory for the publish command.


.github/workflows/ci_python.yml
```diff
@@ -105,15 +105,24 @@ jobs:
             dnf -y install clang protobuf-devel lld
             clang --version
             protoc --version
+      - name: Determine Version
+        if: github.event_name != 'release'
+        run: |
+          pip install tomlkit
+          VERSION=$(python ../scripts/get_version.py pyproject.toml project.version)
+          echo "WHEEL=dist/kaskada-${VERSION}-cp38-abi3-manylinux_2_28_x86_64.whl"
```
bjchambers, line 113: In CI, extract the `project.version` from the `pyproject.toml` and use that to generate the path to the `WHEEL`.

.github/workflows/ci_python.yml
```diff
@@ -105,15 +105,24 @@ jobs:
...
+        # This ensures that we don't accidentally install the version from pypi.
         run: |
           for V in 3.8 3.9 3.10 3.11; do
             echo "::group::Install for Python $V"
             poetry env use $V
             poetry env info
             source $(poetry env info --path)/bin/activate
             poetry install --only=test --only=typecheck
-            pip install 'kaskada[plot]>=0.6.0-a.0' --find-links dist --force-reinstall
+            pip install ${WHEEL} --force-reinstall
```
bjchambers, line 125: Install the wheel by path.

.github/workflows/release_python.yml
```diff
@@ -77,7 +77,11 @@ jobs:
       matrix:
         include:
           - target: x86_64
+            wheel_suffix: macosx_11_0_x86_64
+          - target: arm64
+            wheel_suffix: macosx_11_0_arm64
```
bjchambers, line 82: Build an extra `arm64` wheel -- if that is more specific it can avoid downloading the larger universal binary.

.github/workflows/release_python.yml
```diff
@@ -77,7 +77,11 @@ jobs:
...
           - target: universal2
+            wheel_suffix: macosx_11_0_x86_64.macosx_11_0_arm64.macosx_11_0_universal2
```
bjchambers, line 84: Apparently the `universal2` includes the suffix for everything it includes. This is also why it takes twice as long to build because it has to compile both versions.

.github/workflows/release_python.yml
```diff
@@ -304,6 +315,10 @@ jobs:
         #   repository-url: https://test.pypi.org/legacy/
 
       - name: Publish release
+        # We don't need the source for the release, so we don't currently check it out.
```
bjchambers, line 318: This should fix the failure we observed @jordanrfrazier. We should probably do an alpha release soon after this goes in to verify it works / flush out any surprises.

epinzur (approved):

.github/workflows/ci_python.yml
```diff
@@ -105,15 +105,25 @@ jobs:
...
             clang --version
             protoc --version
+      - name: Determine Version
+        if: github.event_name != 'release'
+        run: |
+          pip install tomlkit
+          VERSION=$(python ../scripts/get_version.py pyproject.toml project.version)
+          WHEEL=dist/kaskada-${VERSION}-cp38-abi3-manylinux_2_28_x86_64.whl
+          echo "WHEEL: ${WHEEL}"
+          echo "WHEEL=${WHEEL}" >> $GITHUB_ENV
```
bjchambers, line 115: Ugh. If version is `0.6.0-a.2` then the actual wheel is `dist/kaskada-0.6.0a2-cp38-abi3-manylinux_2_28_x86_64.whl`.

So we need to do some post-processing on the pre-release version. This is going to get verbose.

Issue #722, opened 2023-08-30
Implement existing operations in Python client


epinzur:
As a new iteration of Kaskada written in Python, I want to add functional and aggregation parity with the existing Kaskada service so that I can enable a transition story from existing users of Kaskada.

Acceptance Criteria:
* Determine the subset of functions required for implementation from the existing [Function Catalog ](https://kaskada.io/docs-site/kaskada/main/fenl/catalog.html). These include but are not limited to:

- [x] lower
- [x] round
- [x] sqrt
- [x] upper
- [x] zip_max (greatest)
- [x] zip_min (least)
- [x] hash
- [x] substring
- [x] len

* There are additional time functions that need to be considered as well, but could be separated out as an individual ticket. See #729

PR #723, opened 2023-08-30
implemented numerous existing operations

epinzur:
progress on #722 

Added:
* hash
* lower
* round
* sqrt
* upper
* zip_max
* zip_min


python/pytests/golden/hash_test/test_hash_integer.jsonl
```diff
@@ -0,0 +1,6 @@
+{"_time":"2021-01-01T00:00:00.000000000","_key":"Ben","m":5.0,"hash_m":16461383214845928621}
+{"_time":"2021-01-01T00:00:00.000000000","_key":"Ryan","m":8.0,"hash_m":6794973171266502674}
+{"_time":"2021-01-02T00:00:00.000000000","_key":"Ryan","m":9.0,"hash_m":15653042715643359010}
+{"_time":"2021-01-03T00:00:00.000000000","_key":"Ben","m":8.0,"hash_m":6794973171266502674}
+{"_time":"2021-01-04T00:00:00.000000000","_key":"Ben","m":null,"hash_m":0}
+{"_time":"2021-01-04T00:00:00.000000000","_key":"Ryan","m":9.0,"hash_m":15653042715643359010}
```
epinzur, line 6: these values aren't what I expected... old docs show different values:
 https://kaskada.io/docs-site/kaskada/main/fenl/catalog.html#hash

also I wasn't expecting `null` to hash to `0`

python/pytests/golden/hash_test/test_hash_string.jsonl
```diff
@@ -0,0 +1,6 @@
+{"_time":"2021-01-01T00:00:00.000000000","_key":"Ben","m":"hello","hash_m":1472103086483932002}
+{"_time":"2021-01-01T00:00:00.000000000","_key":"Ryan","m":null,"hash_m":0}
+{"_time":"2021-01-02T00:00:00.000000000","_key":"Ryan","m":"world","hash_m":8057155968893317866}
+{"_time":"2021-01-03T00:00:00.000000000","_key":"Ben","m":"hi","hash_m":2460612554838835252}
+{"_time":"2021-01-04T00:00:00.000000000","_key":"Ben","m":null,"hash_m":0}
+{"_time":"2021-01-04T00:00:00.000000000","_key":"Ryan","m":"earth","hash_m":14489671231712828724}
```
epinzur, line 6: these values aren't what I expected... old docs show different values:
 https://kaskada.io/docs-site/kaskada/main/fenl/catalog.html#hash

also I wasn't expecting `null` to hash to `0`

python/pytests/golden/sqrt_test/test_sqrt.jsonl
```diff
@@ -0,0 +1,3 @@
+{"_time":"2021-01-01T00:00:00.000000000","_key":"A","m":5.7,"sqrt_m":2.3874672773}
+{"_time":"2021-01-01T00:00:00.000000000","_key":"A","m":6.3,"sqrt_m":2.5099800796}
+{"_time":"2021-01-02T00:00:00.000000000","_key":"B","m":null,"sqrt_m":null}
```
epinzur, line 3: seeing lower precision results than old tests: https://kaskada.io/docs-site/kaskada/main/fenl/catalog.html#sqrt

is this due to some rounding in python?

python/pytests/golden/zip_max_test/test_zip_max.jsonl
```diff
@@ -0,0 +1,4 @@
+{"_time":"2021-01-01T00:00:00.000000000","_key":"A","a":5.7,"b":1.2,"a_zip_max_b":5.7}
+{"_time":"2021-01-01T00:00:00.000000000","_key":"A","a":6.3,"b":0.4,"a_zip_max_b":6.3}
+{"_time":"2021-01-02T00:00:00.000000000","_key":"B","a":null,"b":3.7,"a_zip_max_b":null}
+{"_time":"2021-01-03T00:00:00.000000000","_key":"A","a":13.2,"b":null,"a_zip_max_b":null}
```
epinzur, line 4: based on our docs (not the tests), I would have expected the result on line 3 to be `3.7` and not null.

https://kaskada.io/docs-site/kaskada/main/fenl/catalog.html#zip_max  Specfically: 
> If a or b are null, then b will be returned.

Are the docs wrong or is there a bug in the code?

python/pysrc/kaskada/_timestream.py
```diff
@@ -880,6 +893,84 @@ def record(self, fields: Callable[[Timestream], Mapping[str, Arg]]) -> Timestrea
...
+              if the value was `null` at that row. Otherwise the row
+              contains the square root the value.
+        """
+        return Timestream._call("sqrt", self)
+
+    def upper(self) -> Timestream:
+        """Return a Timestream with all values converted to upper case."""
+        return Timestream._call("upper", self)
+
+    def zip_max(self, rhs: Union[Timestream, Literal]) -> Timestream:
```
bjchambers, line 928: SQL calls this `greatest` and `zip_min` is `least`. We probably want to follow suit:

https://www.tutorialsteacher.com/postgresql/greatest-least-function#:~:text=Json%20Type-,PostgreSQL%3A%20GREATEST()%20and%20LEAST(),take%20any%20number%20of%20arguments.

python/pysrc/kaskada/_timestream.py
```diff
@@ -264,6 +264,19 @@ def floor(self) -> Timestream:
         """Return a Timestream of self rounded down to the nearest integer."""
         return Timestream._call("floor", self)
 
+    def hash(self) -> Timestream:
+        """Return a Timestream containing the hash of the input.
+
+        Notes:
+            Unlike many functions which return `null` if an input is `null`,
```
bjchambers, line 271: The note reads weird. Maybe:

> `hash` will only return `null` when interpolated at points the input is not defined. At other points, it will return the hash of `null`, which is not `null`.

python/pysrc/kaskada/_timestream.py
```diff
@@ -880,6 +893,84 @@ def record(self, fields: Callable[[Timestream], Mapping[str, Arg]]) -> Timestrea
         """
         return record(fields(self))
 
+    def round(self) -> Timestream:
+        """Return a Timestream with all values rounded to the nearest integer.
+
+        Returns:
+            A Timestream of the same type as `self`. The result contains `null`
```
bjchambers, line 900: I've been using "at that point" instead of "position", trying to line the terms up more with timelines/timestreams.

python/pysrc/kaskada/_timestream.py
```diff
@@ -880,6 +893,84 @@ def record(self, fields: Callable[[Timestream], Mapping[str, Arg]]) -> Timestrea
...
+              if the value was `null` at that position. Otherwise, it contains
+              the result of rounding the value to the nearest integer.
+
+        Notes:
+            This method may be applied to any numeric type. For anything other
+              than `float32` and `float64` it has no affect since the values
+              are already integers.
+
+        See Also:
+            `ceil` and `floor`
```
bjchambers, line 910: Double check the rendering (`nox -s docs`) to make sure both of these are linked. I'm not sure. You may need to do:

```
See Also:
  ceil
  floor
```

or possibly comma separated? Not sure what Sphinx expects here.

python/pysrc/kaskada/_timestream.py
```diff
@@ -880,6 +893,84 @@ def record(self, fields: Callable[[Timestream], Mapping[str, Arg]]) -> Timestrea
...
+        """
+        return Timestream._call("round", self)
+
+    def sqrt(self) -> Timestream:
+        """Return a Timestream with the square root of all values.
+
+        Returns:
+            A Timestream of type `float64`. The result contains `null`
```
bjchambers, line 918: I wonder if we should simplify and/or omit this. We probably should just put in the docs somewhere that when we say "returns a timestream with X" we mean "that applies X to each point in the input". Not sure.

python/pysrc/kaskada/_timestream.py
```diff
@@ -880,6 +893,84 @@ def record(self, fields: Callable[[Timestream], Mapping[str, Arg]]) -> Timestrea
...
+            rhs: The Timestream or literal value to compare to this.
+
+        Returns:
+            A numeric Timestream of the promoted type. Each row
+              contains the value from `self` if `self` is greater
+              than `rhs`, otherwise it contains `rhs`. If `self` or
+              `rhs` is `NaN` or `null`, then `rhs` will be returned.
+
+        Notes:
+            Both `self` and `rhs` are promoted to a compatible ordered
```
bjchambers, line 941: Does this note seem excessive or helpful? If we like it, should we add it to other two-argument functions that promote (eg., arithmetic, etc.)

python/pysrc/kaskada/_timestream.py
```diff
@@ -880,6 +893,84 @@ def record(self, fields: Callable[[Timestream], Mapping[str, Arg]]) -> Timestrea
...
+              type following the docs:data-model#numeric-type-coercion-table[numeric type coercion rules].
+
+        See Also:
+            This returns the maximum of two values. See the
```
bjchambers, line 945: Confirm this links to `max`. Otherwise, you may need:

```
See Also:
  max: ...
```

or something like that.

python/pytests/golden/hash_test/test_hash_integer.jsonl
bjchambers: The behavior of hash has changed, but the catalog hasn't been updated since the last time we pushed to read the docs from the old kaskada repo (so basically, really old). The *content* in the repo is likely more up to date:

https://github.com/kaskada-ai/kaskada/blob/main/crates/sparrow-catalog/catalog/hash.toml#L1

We switched to using a hash function based on DataFusion which let us hash structs and things, but it changed the results when applied to `null` to return `0`.

python/pytests/golden/zip_max_test/test_zip_max.jsonl
bjchambers: I'm not sure? My gut says the docs are wrong, since `null` could be interpreted as "unknown" (any value) in SQL, which means that we don't know what the maximum is. On the other hand, I could see value in "take the greatest of these 3, ignoring any nulls". What does SQL do with `least` and `greatest`? Let's name it based on those and file a bug if the behavior doesn't match?

python/pysrc/kaskada/_timestream.py
epinzur: I copied the node directly from the old docs.  but with the new implementation of `hash`, I'm going to change it to: 

> `hash` will only return `null` when interpolated at points the input is not defined. At other points, it will return the hash of `null`, which is `0` (not `null`).

python/pytests/golden/sqrt_test/test_sqrt.jsonl
bjchambers: Could be lower precision in Python, or how the JSON represents it. I'm not sure I'd worry about it now.

bjchambers: My guess is this is related to Pandas and how it prints the JSON:

https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_json.html

Note the default precision is 10 digits.

python/pysrc/kaskada/_timestream.py
epinzur: needs to be like this:
```
        See Also:
            :func:`ceil`
            :func:`floor`
```

python/pysrc/kaskada/_timestream.py
epinzur: ill drop it

python/pytests/golden/zip_max_test/test_zip_max.jsonl
epinzur: per this text, https://modern-sql.com/caniuse/greatest-least the sql standard says that nulls should always pass through.
 
however seems like numerous dialects ignore them. our code currently follows the standard, so let's leave it as is for now.  eventually we might want to add an option to ignore them.

python/pysrc/kaskada/_timestream.py
epinzur: `:func:\`max\``  works.  

I also added see-also on `max` and `min`, pointing to `greatest` and `least`

python/docs/source/reference/timestream/string.md
```diff
@@ -0,0 +1,11 @@
...
+
+```{eval-rst}
+.. currentmodule:: kaskada
+
+.. autosummary::
+   :toctree: ../apidocs/
+
+    Timestream.lower
+    Timestream.upper
+```
```
epinzur, line 11: new bucket? make sense?

Will also add other string functions here later (like `substring`)

python/docs/source/reference/timestream/comparison.md
```diff
@@ -21,7 +21,9 @@ To respect the semantics of `__eq__` and `__ne__`, `a == b` and `a != b` are *no
     Timestream.eq
     Timestream.ge
     Timestream.gt
+    Timestream.greatest
     Timestream.le
+    Timestream.least
```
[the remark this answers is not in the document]
bjchambers: Seems OK for now. Otherwise I could see arithmetic if we wanted to keep this "things that return boolean" and those "things that return numbers"?

python/pysrc/kaskada/_timestream.py
```diff
@@ -760,6 +774,12 @@ def max(self, window: Optional[kd.windows.Window] = None) -> Timestream:
 
         Args:
             window: The window to use for the aggregation. Defaults to the entire Timestream.
+
+        See Also:
+            This returns the the maximum of values in a
```
bjchambers, line 779: s/the the/the

python/pysrc/kaskada/_timestream.py
```diff
@@ -760,6 +774,12 @@ def max(self, window: Optional[kd.windows.Window] = None) -> Timestream:
...
+            column up to and including the current row. See the
+            comparison :func:`greatest` to get the greatest value
```
bjchambers, line 781: Maybe just `See :func:`greatest` for the greatest value between ...` (specifically, not calling it a comparison). Maybe also put `Timestreams` in source ticks.

python/pysrc/kaskada/_timestream.py
```diff
@@ -770,6 +790,12 @@ def min(self, window: Optional[kd.windows.Window] = None) -> Timestream:
 
         Args:
             window: The window to use for the aggregation. Defaults to the entire Timestream.
+
+        See Also:
+            This returns the the minimum of values in a
+            column up to and including the current row. See the
+            comparison :func:`least` to get the least value
```
bjchambers, line 797: comparison

python/pysrc/kaskada/_timestream.py
```diff
@@ -770,6 +790,12 @@ def min(self, window: Optional[kd.windows.Window] = None) -> Timestream:
 
         Args:
             window: The window to use for the aggregation. Defaults to the entire Timestream.
+
+        See Also:
+            This returns the the minimum of values in a
```
bjchambers, line 795: the the

python/pysrc/kaskada/_timestream.py
bjchambers: Maybe also call it "to get the maximum value ..." so it isn't "greatest returns the greatest"?

python/pysrc/kaskada/_timestream.py
```diff
@@ -880,6 +906,71 @@ def record(self, fields: Callable[[Timestream], Mapping[str, Arg]]) -> Timestrea
...
+            if the value was `null` at that point. Otherwise, it contains
+            the result of rounding the value to the nearest integer.
+
+        Notes:
+            This method may be applied to any numeric type. For anything other
+            than `float32` and `float64` it has no affect since the values
+            are already integers.
+
+        See Also:
+            :func:`ceil`
```
bjchambers, line 923: May want to use bullets so this renders as a list:

```
See Also:
  - :func:...
  - :func:...
```

python/pysrc/kaskada/_timestream.py
```diff
@@ -880,6 +906,71 @@ def record(self, fields: Callable[[Timestream], Mapping[str, Arg]]) -> Timestrea
...
+        Returns:
+            Each point contains the value from `self` if `self`
+            is greater than `rhs`, otherwise it contains `rhs`.
+            If any input is `null` or `NaN`, then that will be
+            the result.
+
+        See Also:
+            This returns the greatest of two values. See the
+            aggregation :func:`max` for the maximum of values in
+            a column up to and including the current row.
```
bjchambers, line 951: Suggest dropping the "up to and including the current row" since that will depend on the windowing, etc.

python/pysrc/kaskada/_timestream.py
```diff
@@ -880,6 +906,71 @@ def record(self, fields: Callable[[Timestream], Mapping[str, Arg]]) -> Timestrea
...
+        Returns:
+            Each point contains the value from `self` if `self`
+            is less than `rhs`, otherwise it contains `rhs`.
+            If any input is `null` or `NaN`, then that will be
+            the result.
+
+        See Also:
+            This returns the least of two values. See the
+            aggregation :func:`min` for the minimum of values in
+            a column up to and including the current row.
```
bjchambers, line 970: same

bjchambers (approved):

python/noxfile.py
```diff
@@ -133,7 +133,7 @@ def docs_build(session: nox.Session) -> None:
 @nox.session(python=python_versions[0])
 def docs(session: nox.Session) -> None:
     """Build and serve the documentation with live reloading on file changes."""
-    args = ["--open-browser", "docs/source", "docs/_build", "-j", "auto", "--ignore", "*/apidocs/*"]
+    args = ["--open-browser", "docs/source", "docs/_build", "-j", "auto", "--ignore", "*/apidocs/*", "--watch", "pysrc/kaskada"]
```
epinzur, line 136: now rebuilds after edit of `_timestream.py` and similar files :))

python/pysrc/kaskada/_timestream.py
epinzur: i ended up removing `the comparison`.  now it is just `see *least* to get minimum value...`

PR #728, opened 2023-08-30
feat: introduce new execution API

bjchambers:
This doesn't fully wire up the new functionality (snapshots, changed since, etc.) but it adapts existing functionality to the new API.


python/pysrc/kaskada/results.py
```diff
@@ -0,0 +1,36 @@
...
+
+@dataclass
+class History:
+  """Execution options for queries producing all historic points."""
+
+  #: If set, only returns points after this time.
+  #:
+  #: Setting this allows incremental execution to use a checkpoint
+  #: from a time before the `changed_since` time.
+  changed_since: Optional[datetime] = None
```
kerinin, line 15: Should we find a name for this that's different than `changed_since`, since the meaning is different than in `Snapshot`? Maybe `start` or similar?

bjchambers: I'm OK either way. Maybe it would just be `since`, but that runs into "since windows". Other options:

- `from` and `to`
- `start_from` and `up_to`

bjchambers: Or `min_time` and `max_time`?

PR #731, opened 2023-08-31
feat: allow comparisons against literal timedeltas

jordanrfrazier:
This adds in a way to compare `timedeltas (durations)`. Arrow does not currently support casting any duration types, so this hacks in a manual math + cast to convert from less granular to more granular duration types. 

Note that you cannot perform operations between `timedeltas` and `timestamps`. This is expected, as they are different concepts (one is the time between two points, one is an absolute point in time)


crates/sparrow-instructions/src/evaluators/cast.rs
```diff
@@ -46,6 +50,9 @@ impl CastEvaluator {
                 ),
                 ty,
             ) if CastEvaluator::is_supported_interval_cast(ty) => true,
+            // Support casting from a duration to a duration with a more granular time unit
+            // However, arrow doesn't currently support this directly, so this hacks it a bit.
+            (DataType::Duration(TimeUnit::Second), DataType::Duration(_)) => true,
```
jordanrfrazier, line 55: Oops. Add the rest

crates/sparrow-instructions/src/evaluators/cast.rs
```diff
@@ -89,6 +96,34 @@ impl CastEvaluator {
...
+                }
+                TimeUnit::Nanosecond => {
+                    let input: &DurationSecondArray = input.as_primitive();
+                    let input: Int64Array =
+                        arrow::compute::kernels::arity::unary(input, |n| n * 1_000_000_000);
+                    let result = arrow::compute::cast(&input, to_type)?;
+                    Ok(Arc::new(result))
+                }
+                _ => panic!("unexpected cast from duration('s') to {:?}", tu),
+            },
```
jordanrfrazier, line 126: This is the main "hack"

python/pysrc/kaskada/_timestream.py
```diff
@@ -59,11 +59,19 @@ def __init__(self, ffi: _ffi.Expr) -> None:
...
-        return Timestream(_ffi.Expr.literal(session, value))
+
+        if isinstance(value, timedelta):
+            # The smallest unit stored in timedelta is us
+            us = value / timedelta(microseconds=1)
+            # Get the seconds as an integer
+            seconds = int(us / 1_000_000)
+            # Get the leftover nanoseconds
+            nanoseconds = int((us % 1_000_000) * 1_000)
+            return Timestream(_ffi.Expr.literal_datetime(session, seconds, nanoseconds))
```
jordanrfrazier, line 72: I _could_ just convert to pure nanos here, which would avoid the math in sparrow conversion. Thoughts? `Timedelta` can only represent down to `us`, so we don't have that nanosecond granularity, but that's a different consideration anyway.

python/pytests/golden/time_test/test_time_of_point.jsonl
```diff
@@ -1,4 +1,7 @@
...
+{"_time":"1996-12-19T16:39:57.000000000","_key":"A","m":5.0,"time_of_m":"1996-12-19T16:39:57.000000000","n":10.0,"time_of_n":"1996-12-19T16:39:57.000000000"}
+{"_time":"1996-12-19T16:39:58.000000000","_key":"B","m":24.0,"time_of_m":"1996-12-19T16:39:58.000000000","n":3.0,"time_of_n":"1996-12-19T16:39:58.000000000"}
+{"_time":"1996-12-19T16:39:59.000000000","_key":"A","m":17.0,"time_of_m":"1996-12-19T16:39:59.000000000","n":6.0,"time_of_n":"1996-12-19T16:39:59.000000000"}
+{"_time":"1997-01-18T16:40:00.000000000","_key":"A","m":null,"time_of_m":"1997-01-18T16:40:00.000000000","n":9.0,"time_of_n":"1997-01-18T16:40:00.000000000"}
+{"_time":"1997-01-18T16:40:01.000000000","_key":"A","m":3.0,"time_of_m":"1997-01-18T16:40:01.000000000","n":9.0,"time_of_n":"1997-01-18T16:40:01.000000000"}
+{"_time":"1997-01-18T16:40:03.000000000","_key":"A","m":9.0,"time_of_m":"1997-01-18T16:40:03.000000000","n":1.0,"time_of_n":"1997-01-18T16:40:03.000000000"}
+{"_time":"1997-01-18T16:40:04.000000000","_key":"A","m":5.0,"time_of_m":"1997-01-18T16:40:04.000000000","n":null,"time_of_n":"1997-01-18T16:40:04.000000000"}
```
jordanrfrazier, line 7: Added new rows which is where these test files all changed

crates/sparrow-compiler/src/types/inference.rs
```diff
@@ -489,6 +489,20 @@ fn least_upper_bound_data_type(a: DataType, b: &DataType) -> Option<DataType> {
         (Null, other) => Some(other.clone()),
         (other, Null) => Some(other),
 
+        // Duration types can cast to more granular duration types
+        // s -> any
+        // ms -> us,ns
+        // us -> ns
+        // nanosecond is omitted, as the catch all should find that lhs == rhs
+        (Duration(TimeUnit::Second), other @ Duration(_)) => Some(other.clone()),
```
bjchambers, line 497: Could probably clean this up to be:

```
// Duration types cast to the more granular TimeUnit
(Duration(unit1), Duration(unit2)) => Some(Duration(std::cmp::max(unit1, unit2)))
```

The benefit is that that you don't need to explain as much since it's (a) fewer cases and (b) less code.

crates/sparrow-instructions/src/evaluators/cast.rs
```diff
@@ -46,6 +50,17 @@ impl CastEvaluator {
                 ),
                 ty,
             ) if CastEvaluator::is_supported_interval_cast(ty) => true,
+            // Support casting from a duration to a duration with a more granular time unit
+            // However, arrow doesn't currently support this directly, so this hacks it a bit.
+            (DataType::Duration(TimeUnit::Second), DataType::Duration(_)) => true,
```
bjchambers, line 55: For casting, it probably makes sense to allow both ways. We have a separate "can implicitly cast" which is the thing that should be limited to "only make it more granular". So here, this can just be -- if they're both durations then we can cast".

crates/sparrow-instructions/src/evaluators/cast.rs
```diff
@@ -89,6 +104,34 @@ impl CastEvaluator {
                     input.data_type()
                 )),
             },
+            // Arrow does not support casting directly from duration to duration types, but it does
+            // support casting to the underlying data type (int64) then back to any duration type.
+            // This hacks that cast by doing the multiplication manually then casting back to the
+            // desired type.
+            (DataType::Duration(TimeUnit::Second), to_type @ DataType::Duration(tu)) => match tu {
```
bjchambers, line 111: These can probably be simplified in a few ways:

1. We can cast from duration to int64array and vice versa for free.
2. This means instead of using `arity::unary` to do the multiplication we can just use the multiply by a scalar kernel.
3. Instead of special casing, we can compute the multiplier -- if we know how many nanoseconds are in each unit, then `from_nanos / to_nanos` gives us the conversion. We could maybe check which is greater and use division if going the other direction.

I think putting that together gives us something like:

```
fn time_unit_nanos(unit: &TimeUnit) -> i64 {
  match unit {
    TimeUnit::Second => 1_000_000_000,
    TimeUnit::Millisecond => 1_000_000,
    TimeUnit::Microsecond => 1_000,
    TimeUnit::Nanosecond => 1
  }
}

fn cast_duration(input: &dyn Array, from: &TimeUnit, to: &TimeUnit) -> ArrayRef {
  let input: ArrayRef = arrow::compute::cast(input, DataType::Int64Array)?;
  let input: &Int64Array = input.as_primitive();
  
  let from = time_unit_nanos(from);
  let to = time_unit_nanos(to);
  let result: ArrayRef = if from < to {
    arrow::compute::kernels::numeric::mul(&input, to / from)?
  } else if from > to {
    arrow::compute::kernels::numeric::div(&input, from / to)?
  };
  let result = arrow::compute::cast(result.as_ref(), DataType::Duration(*to))?;
  Ok(result)
}
```

Then you can put more of your comment on the `cast_duration` function. That will avoid needing as many special cases.

crates/sparrow-session/src/expr.rs
```diff
@@ -29,4 +29,6 @@ pub enum Literal {
     Int64(i64),
     UInt64(u64),
     Float64(f64),
+    /// time in s, nanos
+    Timedelta(i64, i64),
```
bjchambers, line 33: Make it a record and name the fields? `TimeDelta { seconds: i64, nanos: i64 }` is a lot clearer.

crates/sparrow-syntax/src/syntax/literal.rs
```diff
@@ -22,6 +22,8 @@ pub enum LiteralValue {
     /// loss of precision we retain the original string.
     Number(String),
     String(String),
+    /// Time in s, nanos
+    Timedelta(i64, i64),
```
bjchambers, line 26: Record.

python/pysrc/kaskada/_ffi.pyi
```diff
@@ -34,6 +34,12 @@ class Expr:
         session: Session,
         value: int | float | str | None,
     ) -> Expr: ...
+    @staticmethod
+    def literal_datetime(
```
bjchambers, line 38: Why is this called `_datetime` when it creates a `timedelat`? What will happen in the future if we want to support datetime?

python/pysrc/kaskada/_timestream.py
bjchambers: I can see either option making sense. Probably OK to keep the split representation for now (main loss to standardizing on nanos is not being able to represent deltas or date times greater than about 292 years.

python/src/expr.rs
```diff
@@ -91,6 +92,19 @@ impl Expr {
         Ok(Self { rust_expr, session })
     }
 
+    #[staticmethod]
+    #[pyo3(signature = (session, s, ns))]
+    fn literal_datetime(session: Session, s: i64, ns: i64) -> PyResult<Self> {
```
bjchambers, line 97: datetime makes timedelta?

python/src/expr.rs
```diff
@@ -91,6 +92,19 @@ impl Expr {
...
+        let mut rust_session = session.rust_session()?;
+
+        let rust_expr = rust_session
+            .add_literal(RustLiteral::Timedelta(s, ns))
```
bjchambers, line 101: Should this be `TimeDelta`?

bjchambers (approved):

crates/sparrow-compiler/src/types/inference.rs
jordanrfrazier: Is the ordering of the enum an invariant we can rely on?

crates/sparrow-instructions/src/evaluators/cast.rs
jordanrfrazier: Added the implicit cast

python/pysrc/kaskada/_ffi.pyi
jordanrfrazier: That is because it definitely should not be called _datetime

python/src/expr.rs
jordanrfrazier: Hm..I'm not sure. I usually think of it as one word, but it probably is two words. I like the way it looks as is, but feel free to disagree

PR #732, opened 2023-08-31
docs: Use pydata theme and setup ablog

bjchambers:


kerinin (approved):

kerinin (commented):
Can Sphinx handle redirects from the old URLs, or does that need to happen somewhere else?

PR #733, opened 2023-09-01
feat: Add support for results and time ranges

bjchambers:
This adds the plumbing to use History and Snapshot result configurations. It adds the necessary decoration code to the session.


crates/sparrow-session/src/session.rs
```diff
@@ -42,6 +49,16 @@ pub struct ExecutionOptions {
     pub max_batch_size: Option<usize>,
     /// Whether to run execute as a materialization or not.
     pub materialize: bool,
+    // History or Snapshot results.
+    pub results: Results,
```
jordanrfrazier, line 53: ///

crates/sparrow-session/src/session.rs
```diff
@@ -42,6 +49,16 @@ pub struct ExecutionOptions {
...
+    /// The changed since time. This is the minimum timestamp of changes to events.
+    /// For historic queries, this limits the output points.
+    /// For snapshot queries, this limits the set of entities that are considered changed.
+    pub changed_since_time_s: Option<i64>,
+    /// The final at time. This ist he maximum timestamp output.
+    /// For historic queries, this limits the output points.
+    /// For snapshot queries, this determines the time at which the snapshot is produced.
+    pub final_at_time_s: Option<i64>,
```
jordanrfrazier, line 61: Maybe: "The time to produce the snapshot at. This _is the_* maximum timestamp output." 

We don't allow `final_at_time` with `historic` queries. Right now, we have three entity behaviors: `ALL, FINAL, FINAL_AT_TIME`, and the last implies `snapshot` behavior.

crates/sparrow-session/src/session.rs
```diff
@@ -345,31 +362,49 @@ impl Session {
         }
     }
 
+    /// Execute the query.
+    ///
+    /// It is unnfortunate this requires `&mut self` instead of `&self`. It relates to the
+    /// fact that the decorations may require mutating the DFG, which in turn requires
+    /// mutability. In practice, the decorations shouldn't mutate the DFG and/or that
+    /// shouldn't require mutating the session.
```
jordanrfrazier, line 370: What are the other options? Could we refactor it such that the decorations create their own DFG, then we connect the original + decoration in different step?

crates/sparrow-runtime/src/execute.rs
```diff
@@ -118,10 +118,18 @@ impl ExecutionOptions {
...
+    }
+
     pub fn set_final_at_time(&mut self, final_at_time: Timestamp) {
         self.final_at_time = Some(final_at_time);
     }
 
+    pub fn set_final_at_time_s(&mut self, seconds: i64) {
+        self.final_at_time = Some(Timestamp { seconds, nanos: 0 });
+    }
+
```
jordanrfrazier, line 132: nit: now that we have multiple methods that set this with different time units, I'd think it better practice to change these two parameters to private, and add access methods. i.e. I wouldn't want to be able to bypass these methods and directly mutate the variable, in case I set it using the wrong time unit somehow. *Admittedly low risk since you have to convert to `Timestamp` before anyway, so not a big deal.

python/pysrc/kaskada/_execution.py
```diff
@@ -11,19 +11,33 @@
...
-    """
+    """Execution options passed to the FFI layer."""
 
+    #: The maximum number of rows to return.
+    #: If not specified, all rows are returned.
     row_limit: Optional[int] = None
+
+    #: The maximum batch size to use when returning results.
+    #: If not specified, the default batch size will be used.
     max_batch_size: Optional[int] = None
```
jordanrfrazier, line 22: Do you want to explain what "batch size in results" indicates here? Or too descriptive for this param

python/pysrc/kaskada/_execution.py
```diff
@@ -11,19 +11,33 @@
...
+    #: The earliest time of changes to include in the results.
+    #: For history, this limits the points output.
+    #: For snapshots, this limits the entities that are output.
+    changed_since: Optional[int] = None
+
+    #: The last time to process.
+    #: If not set, defaults to the current time.
+    #: For history, this limits the points output.
+    #: For snapshots, this determines when the snapshot is produced.
+    final_at: Optional[int] = None
```
jordanrfrazier, line 39: Same doc update -- only snapshots

python/pytests/conftest.py
```diff
@@ -27,8 +27,10 @@ def jsonl(self, data: Union[kd.Timestream, pd.DataFrame]) -> None:
         """Golden test against newline-delimited JSON file (json-lines)."""
         df = _data_to_dataframe(data)
         filename = self._filename("jsonl")
+        print(f"Results: {df}, {filename}")
```
jordanrfrazier, line 30: Keeping this?

python/pytests/conftest.py
```diff
@@ -27,8 +27,10 @@ def jsonl(self, data: Union[kd.Timestream, pd.DataFrame]) -> None:
...
 
         if self._save:
+            print("Saving")
```
jordanrfrazier, line 33: .

jordanrfrazier (approved):

python/pysrc/kaskada/_execution.py
bjchambers: This is more of an internal option any way -- it lets us say "when iterating over the 6 elements in this CSV, give me the results in batches of 2 elements". It lets us test the iterator APIs by controlling the batch size coming out.

crates/sparrow-runtime/src/execute.rs
bjchambers: Maybe? I look at this as a class that should probably go away once we have the Python Session stuff cleaned up, so don't know that we want to maintain the old execution path that much.

crates/sparrow-session/src/session.rs
bjchambers: Not really, since the decorations take the original as their argument. The best option would be if we had logical plans -- then this isn't mutating any DFG, it's just creating a new logical plan node applying the decoration. Then, even if we create a DFG it would be part of compiling the logical plan. Not something I think we should do here.

crates/sparrow-session/src/session.rs
bjchambers: Ah. We should probably allow with history it as a way to limit the upper bound of times. This is documenting the internal options, so I'm OK with it describing how we intend it to work.

python/pysrc/kaskada/_execution.py
bjchambers: Going to keep the options (this is part of the internal API anyway). We should probably just make it a bug. It wouldn't be hard to have the scan limit itself at the final-at time, and that could definitely be useful.

PR #734, opened 2023-09-04
created timestream module and split class up into many files

epinzur:
no longer have a monster `_timestamp.py` file that is hard to work with...

everything is passing except `mypy`

docs look good, wheel works in jupyter notebook.


python/.flake8
```diff
@@ -4,4 +4,5 @@ ignore = E203,E501,W503
 max-line-length = 100
 max-complexity = 10
 rst-roles = class,const,func,meth,mod,ref
-rst-directives = deprecated
\ No newline at end of file
+rst-directives = deprecated
+exclude = pysrc/kaskada/timestream/__init__.py
```
bjchambers, line 8: We shouldn't need to exclude `__init__.py` -- what was `flake8` complaining about?

python/docs/source/reference/execution.md
```diff
@@ -8,4 +8,5 @@
 
     Execution
     ResultIterator
+    record
```
bjchambers, line 11: `record` is not part of execution -- it should be part of `records.md`.

python/docs/source/reference/timestream/aggregation.md
```diff
@@ -9,7 +9,7 @@ It is important to remember that aggregations are partitioned by entity and wind
 ## Aggregation Methods
 
 ```{eval-rst}
-.. currentmodule:: kaskada
+.. currentmodule:: kaskada.timestream
```
bjchambers, line 12: I'm not sure I like exposing this as a separate module -- `kd.timestream` is pretty long. On the other hand, I could be convinced that the user generally doesn't interact with that module directly. But, this was why it was previously `_timestream` -- there doesn't seem to be a reason to expose this as a public module. Why not keep it `_timestream`?

python/pysrc/kaskada/timestream/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+from ._execution import preview, to_pandas, run_iter, write
+from ._grouping import lookup, with_key
+from ._logical import and_, or_, not_
+from ._misc import cast, coalesce, else_, filter, hash, if_, lag, null_if, pipe
+from ._records import col, extend, _record, remove, select
+from ._string import lower, upper
+from ._time import shift_by, shift_to, shift_until, time, seconds_since, seconds_since_previous
+
+# aggregation
+Timestream.collect = collect
```
bjchambers, line 22: This seems pretty messy, and I'm not sure how idiomatic it is.

Looking around, it seems it's not particularly suggested to split up a single class like this. But if we need to, https://stackoverflow.com/questions/47561840/how-can-i-separate-the-functions-of-a-class-into-multiple-files may be a bit cleaner:

```
class Timestream:
   def __init__(self,whatever):
     ...

     # Imported methods
     from ._arithmetic import add, sub, mul, div
```

The benefit being that you don't need to separately import them and then set them on the class -- you can just import them directly into the class.

python/pysrc/kaskada/timestream/_aggregation.py
```diff
@@ -0,0 +1,215 @@
...
+
+from typing import Optional, Union
+
+import kaskada as kd
+import pyarrow as pa
+
+from . import LiteralValue, Timestream, record
+
+
+def _aggregation(
```
bjchambers, line 11: If we're going to group things like this, I wonder if we should also group and organize the tests similarly.

python/pysrc/kaskada/timestream/__init__.py
epinzur: I tried this first, but it doesn't work due to circular references.

python/pysrc/kaskada/timestream/_aggregation.py
epinzur: I think this would be a very good idea.

python/pysrc/kaskada/timestream/__init__.py
epinzur: I realize this is messy, and maybe we shouldn't do this, but I also really dislike the `_timestreams.py` file being so huge and unorganized.

python/.flake8
epinzur: it doesn't like that `Arg, LiteralValue, record, Literal` are imported into `__init__.py` but not used there.

I needed to import them there to resolve the circular reference issues.  basically need all of them + Timestream to exist before the remainder of things.

python/docs/source/reference/execution.md
epinzur: there is `record` and `Timestream.record`... 

seems like `record` should not be under the `Timestream` docs... tossed it here but maybe it should be elsewhere?

python/docs/source/reference/timestream/aggregation.md
epinzur: I don't think I can make a module that starts with an underscore.  I tried that at first but I wasn't having much luck getting it working.

epinzur: looking at docs, seems like I should be able to start a module with an underscore. Maybe something else was the issue at that point. I can try to make it private again.

epinzur: resolved.

python/.flake8
epinzur: fixed this, removed the flake8 exclude.

python/docs/source/reference/timestream/aggregation.md
epinzur: I was also concerned about the longer name length, but users shouldn't ever use the module directly.

epinzur:
hmm, this change seems to break intellisense. i think it means that I should stop working on this.

python/docs/source/reference/execution.md
bjchambers: They're both there because `record` creates a timestream and `Timestream.record` is a short-hand for doing the same thing, but computing the fields from an initial timestream. I think they both belong in `record.md` since they are both "functions for creating a recrod timestream"

bjchambers:
> hmm, this change seems to break intellisense. i think it means that I should stop working on this.

Interesting. I wonder why.

python/pysrc/kaskada/timestream/__init__.py
bjchambers: Hmm. There are ways to avoid that -- I can take a look tomorrow. Usually, the trick is to not import them via `__init__.py` and rather import them from the file they are defined in. Basically, `__init__.py` is just used to create the public / external modules.

See for instance https://medium.com/brexeng/avoiding-circular-imports-in-python-7c35ec8145ed

PR #735, opened 2023-09-05
Adding example showing real-time event handling from API requests

kerinin:


examples/event-api/server.py
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env python
+
+import json, asyncio, time, uuid, asyncio
+import kaskada as kd
+from aiohttp import web
+
+async def main():
```
bjchambers, line 7: I wonder if we should make a page in the docs for this? We can either put it in a notebook and have it show up that way, or we can use [literalinclude](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude) to pull in the code and document it in a page on the docs.

kerinin: We could pull this into the docs. I just needed to get it accessible quickly, but we can definitely tweak now that it's available.

PR #736, opened 2023-09-05
bug: fix set version release script

jordanrfrazier:


scripts/version.py
```diff
@@ -84,7 +85,7 @@ def main() -> None:
             version = normalize_version(version)
         print(version)
     elif args.command == "set":
-        set_version(args.toml_file, args.key, args.version)
+        update_versions(args.entries, args.version)
```
bjchambers, line 88: May want to call `update` `set`, but seems fine.

bjchambers (approved):

scripts/version.py
jordanrfrazier: I'm working on another branch now so will just add as is, but can change later on

PR #737, opened 2023-09-05
bug: fix normalization of versions when setting toml

jordanrfrazier:
I _believe_ we only want the normalized versions when using them in the wheels. When we set them in the toml files, we want to retain the non-normalized format.

### DIFF
--- a/.github/workflows/release_python.yml
+++ b/.github/workflows/release_python.yml
@@ -54,7 +54,7 @@ jobs:
         if: github.event_name != 'release'
         run: |
           pip install tomlkit packaging
-          VERSION=$(python ../scripts/version.py get --normalize pyproject.toml project.version)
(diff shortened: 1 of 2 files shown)


.github/workflows/release_python.yml
```diff
@@ -54,15 +54,14 @@ jobs:
...
           pip install tomlkit packaging
-          VERSION=$(python ../scripts/version.py get --normalize pyproject.toml project.version)
+          VERSION=$(python ../scripts/version.py get pyproject.toml project.version)
           echo "VERSION=$VERSION" >> $GITHUB_ENV
       - name: Determine version (release)
         if: github.event_name == 'release'
         run: |
           echo "GITHUB_REF=${GITHUB_REF}"
           VERSION=${GITHUB_REF#refs/tags/v}
           pip install tomlkit packaging
```
jordanrfrazier, line 64: I can probably remove this line then too?

PR #744, opened 2023-09-07
feat: Configurable retention on PyDict sources

bjchambers:
This renames the `PyList` source to `PyDict`.

This allows the in-memory retention to be disabled for any source, but exposes this for the `PyDict` source, which is often used with materializations.


crates/sparrow-merge/src/in_memory_batches.rs
```diff
@@ -53,7 +55,9 @@ impl InMemoryBatches {
             let (version, old) = &*write;
             let version = *version;
 
-            let merged = if old.num_rows() == 0 {
+            let merged = if !self.retained {
+                old.clone()
```
bjchambers, line 59: This keeps the empty batch forever. We could probably put this in a struct to allow updating just the version (rather than using a tuple. Or we could modify the fields. @jordanrfrazier for thoughts on if it is worth doing to clean thsi up.

crates/sparrow-merge/src/in_memory_batches.rs
```diff
@@ -53,7 +55,9 @@ impl InMemoryBatches {
...
+            } else if old.num_rows() == 0 {
                 batch.clone()
             } else {
                 homogeneous_merge(&self.schema, vec![old.clone(), batch.clone()])
```
bjchambers, line 63: Note on below: The entire batch (including potentially late data) is sent to the materialization. I believe the materialization will then be able to choose how to handle that data.

examples/event-api/server.py
```diff
@@ -6,16 +6,17 @@
...
-    # Initialize event source with historical data
-    events = kd.sources.PyList(
+    # Initialize event source with schema from historical data.
+    events = kd.sources.PyDict(
         rows = [{"ts": start, "user": "user_1", "request_id": "12345678-1234-5678-1234-567812345678"}],
         time_column = "ts",
         key_column = "user",
-        time_unit = "s"
+        time_unit = "s",
+        retained=False,
```
bjchambers, line 19: disable retention in the server example, since we only care about the materialization. note that the `rows` above will then be used only for schema inference.

python/pysrc/kaskada/sources/arrow.py
```diff
@@ -81,6 +82,7 @@ def __init__(
             rows: One or more rows represented as dicts.
             time_column: The name of the column containing the time.
             key_column: The name of the column containing the key.
+            retained: Whether added rows should be retained for queries.
```
bjchambers, line 85: We could (possibly) document the interaction with `rows` / materializations.

kerinin: This might benefit from more explanation, and some help on when to use it vs when not to.

crates/sparrow-merge/src/in_memory_batches.rs
kerinin: Probably worth a comment at least

kerinin (approved):

kerinin:
I get an error when I run the script:

```
(sparrow-py) % ./server.py
Waiting for events...
Traceback (most recent call last):
  File "/Users/ryan.michael/work/kaskada/examples/event-api/./server.py", line 91, in <module>
    loop.run_until_complete(main())
  File "/Users/ryan.michael/.pyenv/versions/3.11.3/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/Users/ryan.michael/work/kaskada/examples/event-api/./server.py", line 67, in main
    async for row in output.run_iter():
  File "/Users/ryan.michael/work/kaskada/python/pysrc/kaskada/_execution.py", line 108, in __anext__
    next_batch = await self._ffi_execution.next_pyarrow_async()
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: error in kaskada Rust code
├╴at src/error.rs:54:21
│
├─▶ execute query
│   ╰╴at /Users/ryan.michael/work/kaskada/crates/sparrow-session/src/session.rs:488:28
│
├─▶ internal compute error: failed to join compute threads
│   ╰╴at /Users/ryan.michael/work/kaskada/crates/sparrow-runtime/src/execute/compute_executor.rs:192:22
│
├─▶ internal compute error: no additional context
│   ╰╴at /Users/ryan.michael/work/kaskada/crates/sparrow-runtime/src/execute/operation.rs:345:58
│
├─▶ internal compute error: invalid input
│   ╰╴at /Users/ryan.michael/work/kaskada/crates/sparrow-runtime/src/execute/operation/scan.rs:200:26
│
╰─▶ Unable to create batch from empty data
    ╰╴at /Users/ryan.michael/work/kaskada/crates/sparrow-runtime/src/execute/operation/scan.rs:199:26
```

bjchambers:
Hmm, yeah. Empty batches posed a problem. I think that should be fixed.

PR #746, opened 2023-09-08
feat: Async back-pressure on adding to a source

### LINKED ISSUE #743: improve backpressure mechanism for in-memory batch channel
The in-memory bounded channel is lagging behind during load tests. A couple options:

1. Block python on send, wait for a space to send a batch to the channel.
2. Assemble larger batches in the consumer to speed up rate of consumption 
3. Instead of blocking, concatenate batches in the sender to send less frequent, bigger batches

bjchambers:
This closes #743


python/pysrc/kaskada/sources/arrow.py
```diff
@@ -109,15 +105,61 @@ def __init__(
...
+        """Create a source reading from rows represented as dicts.
+
+        Args:
+            time_column: The name of the column containing the time.
+            key_column: The name of the column containing the key.
+            retained: Whether added rows should be retained for queries.
+              If True, rows (both provided to the constructor and added later) will be retained
+              for interactive queries. If False, rows will be discarded after being sent to any
+              running materializations. Consider setting this to False when the source will only
+              be used for materialization to avoid unnecessary memory usage.
```
jordanrfrazier, line 129: I think it's good the default is to retain input data, but I wonder if we need to more explicitly say that if you want to run queries over your data, it has to be retained, otherwise we'll discard them. People may associate "retained" as an additional layer of persistence that they shouldn't need for local querying, where they just want to add some rows to in-memory storage then query over them a few times to try things out.

python/pysrc/kaskada/sources/arrow.py
```diff
@@ -109,15 +105,61 @@ def __init__(
...
+            grouping_name: The name of the group associated with each key.
+              This is used to ensure implicit joins are only performed between data grouped
+              by the same entity.
+            time_unit: The unit of the time column. One of `ns`, `us`, `ms`, or `s`.
+              If not specified (and not specified in the data), nanosecond will be assumed.
+        """
+        if schema is None:
+            if rows is None:
+                raise ValueError("Must provide schema or rows")
+            schema = pa.Table.from_pylist(rows).schema
```
jordanrfrazier, line 142: I think this was mentioned in a conversation earlier but -- would it be possible to defer schema inference to when rows _are_ added?

bjchambers: Consider:

```
source = kd.sources.PyDict(...)
source.col("x") # does this throw an error? we don't know
source.add_row({ y: 5 }) # does this throw an error?
```

jordanrfrazier: > `source.col("x") # does this throw an error? we don't know`

Yes: "Source has not defined a schema" 

> `source.add_row({ y: 5 }) # does this throw an error?`

No, but this sets the source schema to `{y: int}`

Just an idea -- I'm sure there may be examples that make it confusing, and I think going explicit for now isn't a bad idea.

bjchambers: > Yes: "Source has not defined a schema"

So this forces you to add a row before you use it:

```
source = kd.sources.PyDict(...)
source.add_row({ y: 5 })
source.col("x")
```

This is just longer and more error-prone than allowing it to be combined like below:

```
source = kd.sources.PyDict({ y: 5 }, ...)
source.col("x")
```

python/pytests/execution_test.py
```diff
@@ -32,15 +34,15 @@ def test_iter_pandas(golden, source_int64) -> None:
...
+async def test_iter_rows(golden, source_int64) -> None:
     results: Iterator[dict] = source_int64.run_iter("row", row_limit=2)
     assert next(results)["m"] == 5
     assert next(results)["m"] == 24
     with pytest.raises(StopIteration):
         next(results)
 
 
-@pytest.mark.asyncio
+@pytest.mark.skip("hanging?")
```
jordanrfrazier, line 45: These are hanging now. Unsure how `AsyncIterator` now plays into this given it's an async source?

python/pytests/pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
+asyncio_mode = auto
```
jordanrfrazier, line 2: Asyncio

python/pysrc/kaskada/sources/arrow.py
```diff
@@ -109,35 +143,120 @@ def __init__(
...
-        self.add_rows(rows)
 
-    def add_rows(self, rows: dict | list[dict]) -> None:
+    @staticmethod
+    async def create(
+        rows: Optional[dict | list[dict]] = None,
+        *,
+        time_column: str,
+        key_column: str,
+        retained: bool = True,
```
jordanrfrazier, line 153: I didn't add the `retained` in all sources. That can be a follow up when we figure out the language

python/pytests/pytest.ini
bjchambers: We don't have a pytest.ini -- we're following the newer pattern of putting the config in the `pyproject.toml`. See here https://github.com/kaskada-ai/kaskada/blob/main/python/pyproject.toml#L155

jordanrfrazier (approved):

PR #749, opened 2023-09-12
feat: add parquet source to python ffi

jordanrfrazier:
Adds the ability to pass the parquet path to sparrow and read from the file directly rather than reading in python and transforming to a record batch.


crates/sparrow-session/src/table.rs
```diff
@@ -79,12 +126,36 @@ impl Table {
...
 
-        self.in_memory_batches
+        source
             .add_batch(prepared)
             .await
             .change_context(Error::Prepare)?;
         Ok(())
     }
+
+    pub async fn add_parquet(&self, path: String) -> error_stack::Result<(), Error> {
```
bjchambers, line 136: should probably switch to `path` to use `&std::path::Path`

crates/sparrow-compiler/src/data_context.rs
```diff
@@ -303,6 +303,52 @@ pub struct GroupInfo {
     key_type: DataType,
 }
 
+/// Concurrent file sets allows us to update the file sets from the
+/// python library and see updates reflected within the data context.
+#[derive(Debug, Default, Clone)]
+pub struct ConcurrentFileSets {
```
bjchambers, line 309: Maybe just call this "FileSets"?

Why do we need multiple file sets, vs adding to the existing file set?

Could this just be `Arc<Mutex<compute_table::FileSet>>`? Doesn't that alreday have a vector inside it?

crates/sparrow-compiler/src/data_context.rs
```diff
@@ -303,6 +303,52 @@ pub struct GroupInfo {
...
+}
+
+impl ConcurrentFileSets {
+    pub fn new(file_sets: Vec<compute_table::FileSet>) -> Self {
+        Self {
+            file_sets: Arc::new(Mutex::new(file_sets)),
+        }
+    }
+
+    pub fn push(&mut self, file_set: compute_table::FileSet) {
```
bjchambers, line 320: I think having this be `&mut self` is misleading -- it makes us think the ownership is what let's us modify. Generally, I'd suggest:

1. having `file_sets` be `Mutex<...>` rather than `Arc<Mutex<...>>`
2. having these take `&self` to indicate "interior mutability" (eg., you can mutate this with only a reference because ti has a mutex inside it)
3. having users choose to have a `Arc<ConcurrentFileSets>` for sharing -- this makes it clear to them that it has an arc and is cheap to clone.

But, I'm not 100% sure that's the best pattern -- just what I've tended towards. We can discuss further.

crates/sparrow-compiler/src/data_context.rs
```diff
@@ -303,6 +303,52 @@ pub struct GroupInfo {
...
+                slice_plan,
+                prepared_files: prepared,
+            });
+        }
+    }
+
+    // TODO: Cloning is bad, but since it's locked behind a mutex, it's necessary.
+    // Can we do better?
+    pub fn get_clone(&self) -> Vec<compute_table::FileSet> {
+        self.file_sets.lock().unwrap().clone()
```
bjchambers, line 344: This should only be necessary if we need ownership of the file sets. We should be able to do something like:

```
pub fn file_sets(&self) -> MutexGuard<'_, Vec<compute_table::FileSet>> {
  self.file_sets.lock().unwrap()
}
```


Note `unwrap` here is a potential source of a panic if the mutex has been poisoned. Probably OK (and should just not let the lock get poisoned).

bjchambers: Hmm. The problem is at least partly that the consumer wants to have the owned file sets and work with them (and probably for lifetime reasons, the guarded reference wouldn't work). I wonder if we'd be better off with a `Mutex<RefCell<Arc<[FileSet]>>>`. Specifically, when we add files, we could create a new Arc but we could then create a copy of the fileset rather than mutating it. This would allow us to just copy the current file set and compute on it.

crates/sparrow-runtime/src/prepare/preparer.rs
```diff
@@ -24,55 +34,148 @@ pub enum Error {
...
+    /// - This produces multiple parts if the input file is large.
+    /// - This produces metadata files alongside data files.
+    pub async fn prepare_parquet(
+        &self,
+        path: &std::path::Path,
+    ) -> error_stack::Result<Vec<PreparedFile>, Error> {
+        // TODO: Support Slicing
+
+        // Prepared files are stored in the following format:
+        // file:///<cwd>/tables/<table_uuid>/prepared/<prepare_version>/<uuid>/part-<n>.parquet
```
bjchambers, line 93: I don't think we *need* them to be stored in that format anymore. Without Wren we don't have tables anymore, etc.

crates/sparrow-runtime/src/prepare/preparer.rs
```diff
@@ -24,55 +34,148 @@ pub enum Error {
...
+        let cur_dir = std::env::current_dir().expect("current dir");
+        let cur_dir = cur_dir.to_string_lossy();
+
+        let uuid = Uuid::new_v4();
+        let output_path_prefix = format!(
+            "file:///{}/tables/{}/prepare/{PREPARE_VERSION}/{uuid}/",
```
bjchambers, line 99: We should track a follow-up to support preparing to S3 prefixes, etc.

crates/sparrow-runtime/src/prepare/preparer.rs
```diff
@@ -24,55 +34,148 @@ pub enum Error {
...
+            source: Some(
+                SourceData::try_from_local(path)
+                    .into_report()
+                    .change_context(Error::Internal)?,
+            ),
+        };
+
+        // TODO: Slicing
+        let mut prepare_stream =
+            prepared_batches(&self.object_stores, &source_data, &self.table_config, &None)
```
bjchambers, line 122: How does this size the file? We may want to do some batching or something, so that multiple prepared batches (or even all prepared batches) become part of the same prepared file.

crates/sparrow-compiler/src/data_context.rs
jordanrfrazier: > Why do we need multiple file sets, vs adding to the existing file set?

If and when we do support slicing, each file set would correspond to a specific slice plan

> Could this just be Arc<Mutex<compute_table::FileSet>>? Doesn't that alreday have a vector inside it?

Could, just wanted to abstract a little bit away

crates/sparrow-compiler/src/data_context.rs
jordanrfrazier: Ah, yep. That makes sense, made that change.

crates/sparrow-runtime/src/prepare/preparer.rs
jordanrfrazier: Added a task

crates/sparrow-runtime/src/prepare/preparer.rs
jordanrfrazier: We still have a table uuid id though, right, and can have multiple tables per session? What parts of this could we drop?

crates/sparrow-runtime/src/prepare/preparer.rs
jordanrfrazier: Yeah, it looks like prepare currently just makes a new file per batch. Probably agree we'd want to batch up batches to some desired size. 

But, seems outside of this PR, so will add a task to follow up with that.

jordanrfrazier: It uses.. this? 
```
fn get_num_files(file_size: usize) -> usize {
    // The number of files is the ceiling of the number of gigabytes in the file
    // e.g. 2.5 gb -> 3 files or 1 gb -> 1 file
    (file_size + GIGABYTE_IN_BYTES - 1) / GIGABYTE_IN_BYTES
}

fn get_batch_size(num_rows: usize, num_files: usize) -> usize {
    // To get the files ~1GB in size, we assume that the ceiling of the total number
    // of rows divided by the number of files.
    (num_rows + num_files - 1) / num_files
}
```

crates/sparrow-compiler/src/data_context.rs
jordanrfrazier: Hm..that still complains for me -- if the intention is to be able to do 
```
    pub fn get(&self) -> Ref<'_, Vec<compute_table::FileSet>> {
        self.file_sets.lock().unwrap().borrow()
    }
```

It's creating a temporary variable with the `lock().unwrap()`.

I can do:
```
    pub fn get(&mut self) -> Ref<'_, Vec<compute_table::FileSet>> {
        self.file_sets.get_mut().unwrap().borrow()
    }
```
but the `&mut self` is perhaps not ideal.

jordanrfrazier: We do want to be able to mutate the `FileSet` within the `Mutex`, which I believe you can't do if it's wrapped by an `Arc`?

crates/sparrow-runtime/src/prepare/preparer.rs
jordanrfrazier: Afaict, the batch size should be a gb, unless I misunderstand.

crates/sparrow-runtime/src/execute/operation/scan.rs
```diff
@@ -156,6 +156,20 @@ impl ScanOperation {
                 ))?,
         );
 
+        // Figure out the projected columns from the table schema.
+        //
+        // TODO: Can we clean anything up by changing the table reader API
```
bjchambers, line 161: I'm sure... but the scan needs some attention in a follow-up. Probably not needed *now*.

crates/sparrow-runtime/src/prepare/preparer.rs
bjchambers: I mean, in theory we could (for now) put them all in a just a "<prepare_prefix>/<uuid>.parquet" path. No reason to associate them with specific prepare versions, etc.

At the least, we could maybe do `<cwd>/tables/<table_uuid>/prepared/<uuid>.parquet` may be enough? 

I think the first UUID was so that we could have data tokens exclude the file, which we're currently not anticipating supporting. Additionally, changing prepare version may not matter since this is ephemeral.

python/pysrc/kaskada/sources/arrow.py
```diff
@@ -445,7 +446,8 @@ async def create(
         """Create a Parquet source.
 
         Args:
-            path: The path to the Parquet file to add.
+            path: The path to the Parquet file to add. This can be relative to the current
```
bjchambers, line 449: I wonder if we should just require these be URLs (so `file://`)

bjchambers (approved):

crates/sparrow-runtime/src/prepare/preparer.rs
jordanrfrazier: Removed the `prepare_version`. 

I think there is benefit in keeping the `table_uuid` and `<uuid>/part_n.parquet` format. For debugging purposes, the user is able to (more easily, though not entirely) inspect their prepared files. Probably nice if we would include the actual table name in the path, but for now I'm fine with it as is until we decide during state management design.

crates/sparrow-compiler/src/data_context.rs
```diff
@@ -303,6 +303,83 @@ pub struct GroupInfo {
...
+            //         "Table '{}' missing file set with requested slice {:?}",
+            //         "asdf",
+            //         requested_slice
+            //     )
+            // }).unwrap();
+            &mut file_set.prepared_files
+        }))
+    }
+
+    pub fn metadata_for_files(&self) -> Vec<String> {
```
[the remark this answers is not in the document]
jordanrfrazier: That's an interesting question. This is how it worked previously, but when I look into it, I would agree that we perhaps only need the fileset corresponding to the requested slice plan? I don't know enough to make that change now without looking into the possible ramifications, so I'll leave it as is with a note to investigate and make a task so it doesn't get lost.

crates/sparrow-compiler/src/data_context.rs
```diff
@@ -314,7 +391,11 @@ pub struct TableInfo {
     ///
     /// Each file set corresponds to the files for the table with a specific
     /// slice configuration.
-    file_sets: Vec<compute_table::FileSet>,
+    ///
+    /// TODO: A lot of existing code assumes this is a non-optional field, so
+    /// leaving this as is for now in the assumption that it'll change rapidly
```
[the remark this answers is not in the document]
jordanrfrazier: 🤷

python/pysrc/kaskada/sources/arrow.py
jordanrfrazier: Resolving so I can merge but tagging @epinzur , thoughts?

epinzur: probably should be urls so we can add support to load from object storage by just changing the url (and maybe setting some additional auth params)

bjchambers: Agreed with Eric that these should be URLs. Also should probably move to a different file (`_parquet.py`) since this is no longer using PyArrow under the hood to load the data in memory.

PR #750, opened 2023-09-13
Doc fixes to align to new API and release versions

kerinin:


python/docs/source/guide/installation.md
```diff
@@ -5,7 +5,7 @@ We suggest using 3.11 or newer, since that provides more precise error locations
 
 ```{code-block} bash
 :caption: Installing Kaskada
-pip install kaskada>=0.6.0-a.0
+pip install kaskada>=0.6.0-a.3
```
bjchambers, line 8: FWIW, the `>=0.6.0-a.0` should still pick the "newest" version greater than that, so we shouldn't need to mention `a.3` specifically. The fact it includes `-a.0` means it will allow pre-release versions.

This should all be simplified once we call this `0.6.0` and start using non-pre-release versions. Then we could just say `pip install>=0.6.0` and rely on the version resolution to ensure it gets the newest (the greater than equal just ensures it doesn't install one of the "old" clients).

python/docs/source/guide/quickstart.md
```diff
@@ -14,10 +14,10 @@ This shows the bare minimum needed to get started with Kaskada.
 ## Install
 
 Install the latest version.
-This uses `kaskdaa>=0.6.0-a.0` to ensure the pre-release version is installed.
+This uses `kaskada>=0.6.0-a.3` to ensure the pre-release version is installed.
```
bjchambers, line 17: Not sure we want to get into the habit of changing these all the time. Should probably either make it a variable, or accept that `>=0.6.0-a.0` will ensure the latest version is installed.

python/docs/source/guide/quickstart.md
```diff
@@ -42,6 +42,6 @@ content = "\n".join(
         "1996-12-19T16:40:02,A,,",
     ]
 )
-source = await kd.sources.CsvString.create(content, time_column="time", key_column="key")
+source = await kd.sources.CsvString(content, time_column="time", key_column="key")
```
bjchambers, line 45: I don't think this will work -- the constructor isn't async, so you can't `await` it, and the constructor doesn't accept content.

python/docs/source/guide/timestreams.md
```diff
@@ -34,7 +34,7 @@ data = "\n".join(
         "1996-12-23T16:40:01,A,12",
     ]
 )
-multi_entity = await kd.sources.CsvString.create(data, time_column="time", key_column="key")
+multi_entity = await kd.sources.CsvString(data, time_column="time", key_column="key")
```
bjchambers, line 37: ditto

python/docs/source/guide/tour.md
```diff
@@ -37,7 +37,7 @@ single_entity = "\n".join(
         "1996-12-24T16:40:02,A,,",
     ]
 )
-single_entity = await kd.sources.CsvString.create(single_entity, time_column="time", key_column="key")
+single_entity = await kd.sources.CsvString(single_entity, time_column="time", key_column="key")
```
bjchambers, line 40: ditto

bjchambers (approved):

PR #751, opened 2023-09-13
Add an example showing how to consume from BlueSky

kerinin:


python/pyproject.toml
```diff
@@ -34,6 +34,7 @@ python = ">=3.9,<4.0"
 pyarrow = "^12.0.1"
 typing-extensions = "^4.7.1"
 plotly = {version = "^5.16.1", optional = true}
+flake8 = "^6.1.0"
```
bjchambers, line 37: We shouldn't add flake8 here -- it is already listed in the optional group for linting.

kerinin: I don't think I added this - it must have been added by some of the tooling

PR #755, opened 2023-09-14
ref: replace first/last with generic aggregators

bjchambers:
This allows us to eliminate the "pushdown" code for applying first/last to records, which were exceedingly complicated.


crates/sparrow-instructions/src/evaluators/aggregation/last_evaluator.rs
```diff
@@ -0,0 +1,396 @@
+use crate::ValueRef;
+use crate::{
+    AggregationArgs, ArrayRefAccumToken, Evaluator, EvaluatorFactory, RuntimeInfo, StateToken,
+    StaticInfo,
+};
+use arrow::array::{Array, ArrayRef, BooleanArray, PrimitiveArray, UInt32Array};
+
+/// Evaluator for the `First` instruction.
```
jordanrfrazier, line 8: last

crates/sparrow-instructions/src/evaluators/aggregation/first_evaluator.rs
```diff
@@ -0,0 +1,410 @@
...
+                    None
+                }
+            })
+            .collect();
+
+        let mut take_output_builder = UInt32Array::builder(input.len());
+        for input_index in 0..input.len() {
+            let entity_index = key_indices.value(input_index) as usize;
+            if input.is_valid(input_index) && take_new_state[entity_index].is_none() {
+                take_new_state[entity_index] = Some((input_index + take_new_state.len()) as u32);
```
jordanrfrazier, line 160: I think we need to follow the same pattern as the above aggregate loop, with the `not_taken`, right?

bjchambers: I don't think so -- `take_new_state` will only be `none` if there isn't a previous value, so it does the equivalent job.

jordanrfrazier (approved):

PR #756, opened 2023-09-14
ref: Introduce logical plan creation

bjchambers:
This introduces a simple, reference counted logical plan.

Future changes:
- Implement compilation from logical plans to physical plans
- Connect new execution to physical plans
- Time domain checking
- Compilation from Fnel to logical plans

### DIFF
--- a/crates/sparrow-logical/src/functions.yml
+++ b/crates/sparrow-logical/src/functions.yml
@@ -1,0 +1,6 @@
+- name: add
+  signature: "<T: number>(x: T, y: T) -> T"
+- name: sub
+  signature: "<T: number>(x: T, y: T) -> T"
+- name: eq
+  signature: "<T: any>(x: T, y: T) -> bool"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4524,6 +4524,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "sparrow-logical"
+version = "0.11.0"
+dependencies = [
+ "arrow-schema",
+ "derive_more",
+ "error-stack",
+ "hashbrown 0.14.0",
+ "insta",
+ "itertools 0.11.0",
+ "serde",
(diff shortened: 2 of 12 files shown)


crates/sparrow-logical/src/functions.yml
```diff
@@ -0,0 +1,6 @@
+- name: add
```
bjchambers, line 1: Thinking it might be nice if this was just a data file like this, rather than the complexity of function registration in the old codebase. I suspect with the new approach, there shouldn't *need* to be much here (possibly just a flag indicating which of these are aggregations).

I'm happy to make this a struct and have a rust file with all of them, but thought I'd prototype it as a file to see if we like this better. Thoughts?

jordanrfrazier: The old pattern requires a lot of additional flags -- time domain checks, is_new patterns, rewrites, etc. I don't think it would be _messier_ to have it in a yml vs. registration as rust, so no strong opinion here.

jordanrfrazier: I suppose we'd miss some type checking / compilation benefits of doing it in rust, but we could still share shared patterns like `AGGREGATION_IS_NEW`, so :shrug:

jordanrfrazier (approved):

crates/sparrow-logical/src/functions.yml
bjchambers: Yep. My hope is that we can make it simpler, and possibly just have something like `is_aggregation` if needed. If this gets to be too messy, I'd lean towards going back to a big list of structs, but possibly all in one place so we can just see the list of entries. But, one benefit of doing this in YML is it may force us to keep it simpler. I'm happy to make it a struct right now, but maybe it's worth *trying* YML to see if it ends up better than what we had?

PR #758, opened 2023-09-17
feat: allow periodic filtering of aggregations using a tumbling window

### LINKED ISSUE #759: Filter periodically (with ticks)
We want to support filtering on predicates (or ticks). The plan is to 

1. Expose a new python window type that is rewritten to use `when` in sparrow -- this allows for `windowed_agg | when(..)` without exposing ticks to timestreams
2. Build up aggregations with the new design when swapping over to partitioned execution codebase

### LINKED ISSUE #766: Select filter not working as intended
Converting sliding windows to use `filter()` to produce values only on the predicate have unearthed a strange behavior of select. 

Preliminary findings show that we may not be respecting interpolation correctly, or the DFG is simplifying / placing merges in a strange place.

jordanrfrazier:
tumbling windows are hacked using `filter`. This adds performance concerns regarding additional merges and filter, but it's okay for now. 

Closes #759 
Closes #766


python/pysrc/kaskada/windows.py
```diff
@@ -16,10 +16,15 @@ class Window(object):
 
 @dataclass(frozen=True)
 class Since(Window):
-    """Window since the last time a predicate was true.
+    """A series of non-overlapping windows that reset each time a predicate evaluates to true.
 
     Aggregations will contain all values starting from the last time the predicate
     evaluated to true (inclusive).
```
bjchambers, line 22: This sounds incorrect. Shouldn't it be *exclusive* of the last time the predicate was true?

python/pysrc/kaskada/windows.py
```diff
@@ -16,10 +16,15 @@ class Window(object):
...
+
+    Aggregations will produce a new value for each input value and additionally when
+    the current window closes.
```
bjchambers, line 25: Not sure about the term "closes". Maybe "and when the predicate evaluates to true causing the window to reset (including all of the results up to that point)".

python/pysrc/kaskada/windows.py
```diff
@@ -55,9 +60,60 @@ def yearly() -> Since:
         return Since(predicate=lambda domain: Timestream._call("yearly", domain))
 
 
+@dataclass(frozen=True)
+class Tumbling(Window):
+    """A series of non-overlapping windows that reset each time a predicate evaluates to true.
```
bjchambers, line 65: Change the terms here to match more of the tumbling window concepts. I don't think that is normally described as "reset each time a predicate evaluates to true"

python/pysrc/kaskada/windows.py
jordanrfrazier: Yeah, I thought it sounded weird too. But it's more so "If an `input` exists in the same row that the `predicate` is true, it will be added to the previous window". 

So maybe it should be simply reworded to "Aggregations are exclusive on start and inclusive on end". Then the next sentence should clarify whether it emits for all inputs or on window boundaries.

python/pysrc/kaskada/windows.py
jordanrfrazier: What do you dislike about "close"? 

Window "open" and window "close" sound like good terms to adopt. (I like it better than "start/end"). 

"Reset" feels more like an implementation strategy than a user-facing explanation, sometimes. The "reset" or "clear" window language feels more like "restarting the window", but without implying that the window is also shifting forward when it's reset. "Open" and "Close" imply to a user that the existing window is "closed", meaning no more values can be added to the window in this time frame, and then a new one in a new time boundary is "opened".

jordanrfrazier: Note: I know "reset" is used elsewhere. Petitioning to change the language everywhere to use "open" and "close" instead of "reset"

python/pysrc/kaskada/windows.py
```diff
@@ -55,9 +62,73 @@ def yearly() -> Since:
         return Since(predicate=lambda domain: Timestream._call("yearly", domain))
 
 
+@dataclass(frozen=True)
+class Tumbling(Window):
+    """Window producing a single value at the time of a predicate.
+
+    "Tumbling" windows are a series of non-overlapping windows producing a single value
```
bjchambers, line 69: Close. It seems like we're missing a detail -- "producing a single value at the time the predicate evaluates to true".

1. What happens if the predicate evaluates to true multiple times
2. How do we determine what values are included?

Maybe this is similar to Since windows? We need to do something to explain the window is between two successvie evaluations of the predicate to true.

Also note: This isn't *truly* tumbling windows. With tumbling windows the window is always the same size, and once it "ends" it just flips/shifts over and starts again. Specifically, with a predicate, these are likely different sizes. Even with something like "monthly", the "tumbling" equivalent would be fixed size (30 days or 31 days, but never changing) while our "monthly" is more of a "calendar monthly window".

python/pysrc/kaskada/windows.py
```diff
@@ -55,9 +62,73 @@ def yearly() -> Since:
...
+    If multiple windows exist, the value produced at any given time will be the cumulative
+    value in the most recently created window.
+
+    Args:
+        duration: the number of active windows at any given time.
+        predicate: the condition used to determine when the oldest window ends and a new
+          window starts.
+
+    Returns:
+        Overlapping windows for aggregating cumulative values since the predicate.
```
bjchambers, line 130: Since the N-th predicate evaluation to true?

I wonder if these would be easier if we described this as a "boundary" or "segment" or something. Basically, window boundaries exist when the predicate evaluates to true. Since windows produce cumulative values since the most recent boundary. Sliding windows produce cumulative values since the Nth most recent boundary. Tumbling windows produce the value since the last boundary at each boundary.

python/pysrc/kaskada/windows.py
jordanrfrazier: Yeah, that's true. Maybe we shouldn't use `tumbling` then, since it deviates from the "standard" definition.

bjchambers: Yeah, but I'm not sure what we *should* call them then.

1. In Dataflow, they had "Calendar Windows" which makes sense for things like monthly. It's a bit odd for "hourly", but it can still (somewhat) make sense when you factor in leap seconds and such.
2. We could maybe call them tumbling if we had a way of explaining the difference. Perhaps it's something like "Dynamic Tumbling" or "Calendar Tumbling", where the property is that they are non-overlapping and when one window ends the next starts. That would let us have "Fixed tumbling" (exactly 3600 seconds for an hour) in the future.
3. These also feel kind of like a "session" between a predicate.

python/pysrc/kaskada/windows.py
```diff
@@ -31,33 +38,97 @@ class Since(Window):
...
 @dataclass(frozen=True)
 class Sliding(Window):
-    """Window for the last `duration` intervals of some `predicate`."""
+    """Overlapping windows producing the latest cumulative value since a predicate.
+
+    "Sliding" windows are a series of overlapping windows producing cumulative values
+    since each time the predicate evaluates to true. Each window is inclusive of the time
+    it starts at and exclusive of the end time.
+
+    If multiple windows exist, the value produced at any given time will be the cumulative
```
bjchambers, line 121: These docs all still say `cumulative` for sliding?

python/pysrc/kaskada/windows.py
```diff
@@ -31,33 +38,97 @@ class Since(Window):
...
+
+    "Tumbling" windows are a series of non-overlapping windows that produce values
+    each time the predicate evaluates to true. Each window is inclusive of the time
+    it starts at and exclusive of the end time.
+
+    Args:
+        predicate: the condition used to determine when the window resets.
+
+    Returns:
+        Window for aggregating cumulative values since the predicate.
```
bjchambers, line 77: Maybe add a:

```
Note:
    Like other systems, Kaskada treats tumbling windows as non-overlapping.
    When one window ends the next starts.
     
    Unlike other systems, the tumbling windows do not need to be of a fixed size.
    Instead, they may be determined by a fixed duration, a calendar duration (such as a "month")
    or even a predicate.
```

python/pysrc/kaskada/windows.py
jordanrfrazier: Just going to leave it as `overlapping windows for aggregation values` -- I highly anticipate we'll change the parameters for these

python/pysrc/kaskada/windows.py
```diff
@@ -31,33 +38,102 @@ class Since(Window):
...
 
+@dataclass(frozen=True)
+class Tumbling(Window):
+    """Window producing a single value at each predicate.
+
+    "Tumbling" windows are a series of non-overlapping windows that produce values
+    each time the predicate evaluates to true. Each window is exclusive of the time
+    it starts at and inclusive of the end time.
+
+    Note:
```
bjchambers, line 73: I think it may want the `Note` after `Args` and `Returns`. If not this is fine. Maybe check how it renders.

python/pysrc/kaskada/windows.py
```diff
@@ -31,33 +38,102 @@ class Since(Window):
...
+
+        Unlike other systems, tumbling windows do not need to be of a fixed size.
+        Instead, they may be determined by a fixed duration, a calendar duration (such as a "month"),
+        or even a predicate.
+
+    Args:
+        predicate: the condition used to determine when the window resets.
+
+    Returns:
+        Window for aggregating cumulative values since the predicate.
```
bjchambers, line 85: Tumbling windows don't return cumulative values. This should be `Window for aggregating values since the predicate`

bjchambers (approved):

jordanrfrazier:
TODO: Investigate https://github.com/kaskada-ai/kaskada/issues/766 before merging

bjchambers (approved):

PR #762, opened 2023-09-18
benchmarks: Flesh out scripts for the SQL benchmark

bjchambers:


benchmarks/sql-benchmark/queries_duckdb.sql
```diff
@@ -1,28 +1,34 @@
 PRAGMA enable_profiling;
 
--- Create (L2.87s)
+-- Temp directory is necessary for larger queries.
+PRAGMA temp_directory='tmp';
+
+-- Set this to the input directory to use
+-- PRAGMA file_search_path='6x_1';
```
bjchambers, line 7: I *think* I added this because I was planning on making a bunch of different data sets of different sizes (eg., 6x_1, 6x_2, 6x_3, 12x_1, 12x_2, 12x_3, etc.) to measure (a) average time on a given size and (b) scaling of different systems. Hadn't finished that.

benchmarks/sql-benchmark/kaskada.ipynb
```diff
@@ -58,6 +58,9 @@
    "source": [
     "from kaskada import table\n",
     "\n",
+    "# Change this to the (relative) path to the generated input.\n",
+    "INPUT_PATH = '6x_1'\n",
```
bjchambers, line 62: This is the equivalent. May be worth modifying the generator script to include an "output prefix" or something to more easily work with these "input paths".

jordanrfrazier (approved):

PR #763, opened 2023-09-20
bug: drains output channel when batching results 

### LINKED ISSUE #775: Preview blocks on excess batches (not rows)
Solution is to drain output channel to unblock `preview()` queries.

jordanrfrazier:
Drains the output channel. Previously was causing queries to block on send. 

Puts the tokio runtime into an `Arc` to share it -- the `handle()` was not working like I wanted since it was being dropped (a blocking method) from an asynchronous context. Unsure why, but swapping to use `Arc`, as is also recommended, works fine. 

https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#sharing

Closes #775


crates/sparrow-session/src/execution.rs
```diff
@@ -1,26 +1,28 @@
...
 use futures::{StreamExt, TryStreamExt};
 use sparrow_api::kaskada::v1alpha::ExecuteResponse;
 
 use crate::Error;
 
 pub struct Execution {
-    /// Tokio runtme managing this execution.
-    rt: tokio::runtime::Runtime,
+    /// Tokio runtime managing this execution.
+    pub rt: Arc<tokio::runtime::Runtime>,
```
jordanrfrazier, line 13: don't need pub

crates/sparrow-session/src/execution.rs
```diff
@@ -56,13 +54,13 @@ impl Execution {
...
         let result = match &mut self.status {
-            Status::Running(future) => {
+            Status::Running(progress) => {
                 // Based on the implementation of `FutureExt::now_or_never`:
                 let noop_waker = futures::task::noop_waker();
                 let mut cx = std::task::Context::from_waker(&noop_waker);
 
-                match future.as_mut().poll(&mut cx) {
-                    std::task::Poll::Ready(x) => x,
+                match progress.try_poll_next_unpin(&mut cx) {
```
jordanrfrazier, line 62: I'm pretty sure this is the equivalent behavior

jordanrfrazier:
Current status with runtime in session and using handles: the runtime is closing unexpectedly when a query is being made, resulting in an error. Need to investigate deeper into why this is happening.

crates/sparrow-session/src/execution.rs
```diff
@@ -1,47 +1,43 @@
...
+    /// Future which resolves to the first error or None.
     status: Status,
     /// Stop signal. Send `true` to stop execution.
     stop_signal_rx: tokio::sync::watch::Sender<bool>,
     pub schema: SchemaRef,
 }
 
 enum Status {
-    Running(BoxFuture<'static, error_stack::Result<(), Error>>),
+    // Running(BoxFuture<'static, error_stack::Result<(), Error>>),
```
bjchambers, line 22: Delete old code? Any reason this needed to switch to a stream?

crates/sparrow-session/src/execution.rs
bjchambers: I'm not sure. I don't think so -- previously it ran in a loop to wait until the end of the stream. Now it seems to return sooner.

crates/sparrow-session/src/execution.rs
```diff
@@ -99,29 +95,51 @@ impl Execution {
...
 
-        Ok(self.output.collect().await)
+        let mut outputs = Vec::new();
+        loop {
+            tokio::select! {
+                // Poll futures in the order listed.
+                biased;
+
+                progress = progress.next() => {
+                    if let Some(execute_response) = progress {
```
bjchambers, line 119: If this is `None`, then the stream is done. Do we need to do anything special in that case? Guessing no -- we just need to wait until the output stream finishes.

crates/sparrow-session/src/execution.rs
```diff
@@ -99,29 +95,51 @@ impl Execution {
...
+                        // Propagate errors
+                        let _ = execute_response?;
+                    }
+                }
+                batch = self.output.next() => {
+                    if let Some(batch) = batch {
+                        outputs.push(batch);
+                    } else {
+                        // Output stream is empty, break out and return
```
bjchambers, line 128: Not just empty. We also need to make sure that it is *done* (eg., nothing will be added). If we get here and this gets `None`, then it is reasonable to assume the senders are dropped, so we shouldn't get more batches in the future.

But consider that they were dropped due to an error. If we finish here, we may fail to report the error if the progress channel isn't done.

bjchambers (changes requested):

crates/sparrow-session/src/execution.rs
```diff
@@ -99,29 +95,51 @@ impl Execution {
 
     pub fn next_blocking(&mut self) -> error_stack::Result<Option<RecordBatch>, Error> {
         self.is_done()?;
```
bjchambers, line 97: Thought: We can probably cleanup a lot of this if we treat this as "only use the execution once".

1. `is_done()` doesn't really need to do any async stuff. It could just check to see if we've "used" up the future. Then `next_blocking` just needs to make sure it's `Running` (and/or throw an error if we want it to).
2. `collect_all` can just iterate collecting things until the stream is empty, and then it can run down the progress updates to see if it ends in `Ok()` in which case it returns the collected batches or `Err(...)` in which case it returns the error. May be simpler than trying to use the `select`, and handles the race condition noted.

bjchambers (approved):

PR #767, opened 2023-09-21
added substring, json, len, is_valid methods

### LINKED ISSUE #722: Implement existing operations in Python client

epinzur:
this closes #722


python/docs/source/reference/timestream/misc.md
```diff
@@ -12,6 +12,7 @@
     Timestream.filter
     Timestream.hash
     Timestream.if_
+    Timestream.is_valid
```
bjchambers, line 15: `is_valid` was added as `is_not_null` in `logical.md`. I don't think we need to add it.

python/docs/source/reference/timestream/string.md
```diff
@@ -6,6 +6,9 @@
 .. autosummary::
    :toctree: ../apidocs/
 
+    Timestream.json
```
bjchambers, line 9: If we can't get `json` to work as it was, we can either drop it for now, or expose the primitive `json_field` which takes the field name: `e.json_field("name")`. The support was previously relying on detecting `json(e).name` and converting it to that instruction, but that is only part of the Fenl compilation path, not what the Python stuff goes through.

I think maybe we should skip JSON for now.

python/pytests/substring_test.py
```diff
@@ -0,0 +1,35 @@
...
+            '1996-12-19T16:40:01,B,,2,"salutations"',
+            '1996-12-19T16:40:02,B,"goodbye",,',
+        ]
+    )
+    return await kd.sources.CsvString.create(
+        content, time_column="time", key_column="key"
+    )
+
+
+async def test_substring(source, golden) -> None:
```
bjchambers, line 23: We may also want to test with negative numbers (I think substring uses that to reflect "distance from the end"). So `start=-1` would capture the last character, etc.

python/pysrc/kaskada/sources/arrow.py
```diff
@@ -246,6 +246,10 @@ def __init__(
             strings_can_be_null=True,
         )
 
+    _parse_options = pyarrow.csv.ParseOptions(
+        escape_char="\\",
+    )
+
```
epinzur, line 252: this was done to allow escaping comma's in CSV to handle JSON.  should I remove for now since we aren't adding json anymore?

python/pytests/substring_test.py
epinzur: this already exists via line 12, 31, 32

python/docs/source/reference/timestream/string.md
epinzur: removed. issue made to do later #768

python/pysrc/kaskada/_timestream.py
```diff
@@ -321,6 +321,14 @@ def hash(self) -> Timestream:
         """
         return Timestream._call("hash", self)
 
+    def len(self) -> Timestream:
+        """Return a Timestream with the length of input string.
```
bjchambers, line 325: May want this to support lists as well. Not sure if we need to do anything *now*.

python/pysrc/kaskada/sources/arrow.py
bjchambers: seems OK to leave. fine either way.

bjchambers (approved):

PR #772, opened 2023-09-26
removed ci workflows for the engine, python client, and cli

epinzur:
unfortunately there is no easy way to disable these from the config files. 

We have this tag if we want to reference these or bring them back: https://github.com/kaskada-ai/kaskada/releases/tag/before_engine_decomission

Before we move ahead with this change, we will also need to remove the associated checks to merge to main: 
https://github.com/kaskada-ai/kaskada/settings/branch_protection_rules/34344582


.github/workflows/ci_engine.yml
bjchambers, line 1: Before we remove the `ci_engine` we should add this to the `ci_python`. Specifically, that already has to do a build for `maturin` to test the Python code. So, we should have that *also* do the Rust lint, and Rust unit tests (since it will already have fetched the packages and such).

.github/workflows/release_engine.yml
bjchambers, line 1: We likely want some of the Rust tests to be part of `release_python`.

.github/workflows/reusable_ci_rust.yml
```diff
@@ -1,212 +0,0 @@
-on:
```
epinzur, line 1: maybe keep this one?

.github/release-drafter-engine.yml
```diff
@@ -1,11 +0,0 @@
-_extends: kaskada:.github/release-drafter.yml
```
bjchambers, line 1: We can probably combine the general `release-drafter.yml` and the `release-drafter-python.yml` since we don't need the "sharing".

.github/workflows/ci_notebooks.yml
```diff
@@ -0,0 +1,25 @@
+# TODO: Use Makefile provided in CI (might have to move to docker builds in CI or drop docker in Makefile)
```
bjchambers, line 1: Can probably clean this up (`ent` no longer exists, etc.)

PR #773, opened 2023-09-26
added new jsonl-file source

epinzur:
In my notebooks, I'm tired of loading `jsonl` files into arrays before inputting them into kaskada. So I made a new source to handle `jsonl` files directly.

### DIFF
--- a/python/pysrc/kaskada/sources/arrow.py
+++ b/python/pysrc/kaskada/sources/arrow.py
@@ -309,6 +309,97 @@ class CsvString(Source):
         for batch in content.to_batches():
             await self._ffi_table.add_pyarrow(batch)
 
+class JsonlFile(Source):
+    """Source reading data from line-delimited JSON files using PyArrow."""
+
+    def __init__(
+        self,
+        *,
+        time_column: str,
+        key_column: str,
+        schema: pa.Schema,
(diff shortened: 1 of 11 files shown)


python/pysrc/kaskada/sources/arrow.py
```diff
@@ -310,6 +309,99 @@ async def add_string(self, csv_string: str | BytesIO) -> None:
             await self._ffi_table.add_pyarrow(batch)
 
 
+class JsonlFile(Source):
```
bjchambers, line 312: I wonder if we want these to be separate sources (for files and strings) or a single `Jsonl` source thaht has `add_files` and `add_jsonl` (or uses `file://` prefixes to determine the kind of strings).

Unrelated to this PR, but wondering if there would be value (in the future) in implementing this within Rust so we can parse more directly?

bjchambers (approved):

python/pysrc/kaskada/sources/arrow.py
epinzur: i think this works for now, but we should talk about maybe cleaning up these methods.

or add a high-level `File(Source):` that does `csv`, `parquet`, `jsonl` based on extension.

PR #777, opened 2023-09-26
bug: allow floats as timestamp column

### LINKED ISSUE #776: Loading parquet sources with Float timestamp is broken
getting:
```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: failed to prepare batch
├╴at [/Users/eric.pinzur/src/github.com/kaskada-ai/kaskada/crates/sparrow-session/src/table.rs:150:14](https://file+.vscode-resource.vscode-cdn.net/Users/eric.pinzur/src/github.com/kaskada-ai/kaskada/crates/sparrow-session/src/table.rs:150:14)
│
├─▶ internal error
│   ╰╴at [/Users/eric.pinzur/src/github.com/kaskada-ai/kaskada/crates/sparrow-runtime/src/prepare/preparer.rs:122:18](https://file+.vscode-resource.vscode-cdn.net/Users/eric.pinzur/src/github.com/kaskada-ai/kaskada/crates/sparrow-runtime/src/prepare/preparer.rs:122:18)
│
├─▶ failed to create Parquet file reader
│   ╰╴at [/Users/eric.pinzur/src/github.com/kaskada-ai/kaskada/crates/sparrow-runtime/src/prepare.rs:79:18](https://file+.vscode-resource.vscode-cdn.net/Users/eric.pinzur/src/github.com/kaskada-ai/kaskada/crates/sparrow-runtime/src/prepare.rs:79:18)
│
╰─▶ Internal error
    ├╴at [/Users/eric.pinzur/src/github.com/kaskada-ai/kaskada/crates/sparrow-runtime/src/prepare.rs:78:18](https://file+.vscode-resource.vscode-cdn.net/Users/eric.pinzur/src/github.com/kaskada-ai/kaskada/crates/sparrow-runtime/src/prepare.rs:78:18)
    ╰╴Expected column 'ts' to be castable to Timestamp(Nanosecond, None), but Float64 was not', src/table.rs:95:24
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
---------------------------------------------------------------------------
RustPanic                                 Traceback (most recent call last)
[/var/folders/_7/czrlt9jx79n78l_zsyfwycs40000gp/T/ipykernel_50657/1098232920.py](https://file+.vscode-resource.vscode-cdn.net/var/folders/_7/czrlt9jx79n78l_zsyfwycs40000gp/T/ipykernel_50657/1098232920.py) in <cell line: 1>()
      2 # Use the "ts" column as the time associated with each row,
      3 # and the "channel" column as the entity associated with each row.
----> 4 messages = await kd.sources.Parquet.create(
      5     path = "slack-export/messages.parquet",
      6     time_column = "ts",

[~/.pyenv/versions/3.11.4/lib/python3.11/site-packages/kaskada/sources/arrow.py](https://file+.vscode-resource.vscode-cdn.net/Users/eric.pinzur/src/github.com/kaskada-ai/beep-gpt/~/.pyenv/versions/3.11.4/lib/python3.11/site-packages/kaskada/sources/arrow.py) in create(path, time_column, key_column, schema, subsort_column, grouping_name, time_unit)
    579 
    580         if path:
--> 581             await source.add_file(path)
    582         return source
    583 

[~/.pyenv/versions/3.11.4/lib/python3.11/site-packages/kaskada/sources/arrow.py](https://file+.vscode-resource.vscode-cdn.net/Users/eric.pinzur/src/github.com/kaskada-ai/beep-gpt/~/.pyenv/versions/3.11.4/lib/python3.11/site-packages/kaskada/sources/arrow.py) in add_file(self, path)
    586         path = Source._get_absolute_path(path)
    587 
--> 588         await self._ffi_table.add_parquet(path)

RustPanic: rust future panicked
```

jordanrfrazier:
Allows `timestamp` column to be `float` for parquet sources in the new prepare path. 

Also fixes the `time_unit`, which was previously not being taken into account. 

Closes #776


crates/sparrow-runtime/src/prepare/column_behavior.rs
```diff
@@ -311,6 +376,52 @@ impl ColumnBehavior {
...
+                )
+                .into_report()
+                .change_context_lazy(error)?)
+            }
+            _ => Ok(time.clone()),
+        }
+    } else {
+        Ok(time.clone())
+    }
+}
```
jordanrfrazier, line 423: A lot of this is halfway duplicated from the `prepare_batch` code. It's not a complete  duplicate, but there's definitely room to simplify, but it would require more refactoring of the `column_behavior` code. Not sure how much we want to do now.

crates/sparrow-runtime/src/prepare/column_behavior.rs
```diff
@@ -24,8 +25,15 @@ pub enum ColumnBehavior {
     /// Cast the given column to the given data type.
```
bjchambers, line 25: Do we need to use / change the column behavior? I think we could just use the `prepare_batch` method and avoid that. Alternatively, it seems like we could have a case here that called into the `numeric_to_timestamp` and at least shared that case (rather than adding more complexity to column behaviors).

crates/sparrow-runtime/src/prepare/column_behavior.rs
bjchambers: The best options seem to be either:

1. Drop the column behavior stuff and have both paths call `prepare_batch` (the newer code).
2. Add a column behavior corresponding to `numeric_to_timestamp` which `prepare_batch` uses for the timestamp case.

Either would let us avoid adding as much code to `column_behavior`.

python/pytests/parquet_source_test.py
```diff
@@ -24,3 +24,22 @@ async def test_read_parquet_with_subsort(golden) -> None:
 
     await source.add_file("../testdata/purchases/purchases_part2.parquet")
     golden.jsonl(source)
+
+
+async def test_time_column_as_float_can_cast_s(golden) -> None:
```
bjchambers, line 29: Comment with issue this is a regression test for?

python/pytests/parquet_source_test.py
```diff
@@ -24,3 +24,22 @@ async def test_read_parquet_with_subsort(golden) -> None:
...
+    source = await kd.sources.Parquet.create(
+        "../testdata/purchases/purchases_float_time.parquet",
+        time_column="purchase_time",
+        key_column="customer_id",
+        time_unit="s",
+    )
+    golden.jsonl(source)
+
+async def test_time_column_as_float_can_cast_ns(golden) -> None:
```
bjchambers, line 38: .

PR #778, opened 2023-09-26
ref: use new prepare code and remove some old 

### LINKED ISSUE #776: Loading parquet sources with Float timestamp is broken

jordanrfrazier:
This removes the `column_behavior` in favor of the simpler, newer prepare_batch logic. This also fixes the "floats not allowed as timestamps" behavior, since the new logic handles that. 

Closes #776


bjchambers:
If this passes all the tests... that's awesome. Deletes a bunch of old code and fixes bugs. I think the one thing that we maybe lose is the ability to drop/evolve columns and such -- but I think that can be added to the existing prepare batch -- if we know the "expected" schema, we can just omit/evolve columns based on the file schema and expected schema. Definitely simpler!

bjchambers (approved):

crates/sparrow-runtime/src/prepare/preparer.rs
```diff
@@ -174,8 +175,8 @@ impl Preparer {
 
         let num_rows = batch.num_rows();
         let subsort = if let Some(subsort_column_name) = subsort_column_name.as_ref() {
-            let subsort = get_required_column(&batch, subsort_column_name)?;
-            arrow::compute::cast(subsort.as_ref(), &DataType::UInt64)
+            let subsort = get_required_column(&batch, &subsort_column_name)?;
+            order_preserving_cast_to_u64(subsort)
```
bjchambers: I'm not sure we *need* the order preserving cast here. Specifically, this would only be helpful if the user properly sorted data ahead of time and subsort wasn't a u64 and order preserving cast actually preserved order. It may actually be more limited though (eg., I don't know that order preserving cast handles `f64` to `u64`.

Since we're going to resort anyway... the only thing this means is the user can't rely on the *order* of the subsorts being meaningful if they didn't choose `u64`... but it would still work for deduplicating things.

Not sure whether we want to include this chaneg.

jordanrfrazier: Hm yeah, I would have to add float support to that method if we wanted it. I'm okay with not having it for now, and we can add it if we need it then. I'll revert this.

PR #780, opened 2023-09-27
feat: Expose DFG and query plans on Timestream

bjchambers:
This is useful for debugging and understanding how a query will be executed.


python/docs/source/examples/bluesky.py
```diff
@@ -7,11 +7,14 @@
 
 import asyncio
 import time
+
 import kaskada as kd
 import pyarrow as pa
-
-from atproto.firehose import AsyncFirehoseSubscribeReposClient, parse_subscribe_repos_message
 from atproto import CAR, AtUri, models
+from atproto.firehose import (
```
bjchambers, line 14: Nox linting changed the examples. Seems good to apply our linting conventions here as well.

python/pyproject.toml
```diff
@@ -34,6 +35,7 @@ python = ">=3.9,<4.0"
 pyarrow = "^12.0.1"
 typing-extensions = "^4.7.1"
 plotly = {version = "^5.16.1", optional = true}
+pydot = { version = "^1.4.2", optional = true }
```
bjchambers, line 38: There are other crates we could use. Not sure which is best.

https://pygraphviz.github.io/documentation/stable/ has newer releases. Maybe we should switch?

Either way, it's an optional dependency, so you don't need to have it.

python/pyproject.toml
```diff
@@ -17,6 +17,7 @@ dependencies = [
 [project.optional-dependencies]
 plot = [
```
bjchambers, line 18: I should probably put this in a different group of optional dependencies. This would cause `pip install kaskada[plot]` to install both plotly and pydot. Maybe we want it a different `diagnose` or `internal` or `debug` group? Thoughts?

python/pysrc/kaskada/_timestream.py
```diff
@@ -1192,6 +1194,69 @@ def run_iter(
...
+        options = _ExecutionOptions.create(
+            results, row_limit=None, max_batch_size=None, mode=mode
+        )
+
+        dot = expr._ffi_expr.plan(kind, options)
+        if format == "dot":
+            return dot
+
+        svg_str = ""
+        try:
```
bjchambers, line 1241: This pattern of importing within a try block ensures that we don't require it for other files. It also lets us report an error saying "this package needs to be installed for this option".

python/pysrc/kaskada/_timestream.py
```diff
@@ -1192,6 +1194,69 @@ def run_iter(
 
         raise AssertionError(f"Unhandled kind {kind}")
 
+    def plan(
+        self,
+        kind: Literal["initial_dfg", "final_dfg", "final_plan"] = "final_plan",
+        results: Optional[Union[kd.results.History, kd.results.Snapshot]] = None,
+        format: Literal["dot", "svg", "svg_str"] = "svg",
```
bjchambers, line 1201: We may want this to be optional, and then say the default depends on environment / available packages. Eg., I could see defaulting to `dot` if you don't have the pydot package, etc. At which point the logs below could be warnings.

python/pysrc/kaskada/_timestream.py
```diff
@@ -1192,6 +1194,69 @@ def run_iter(
...
+        mode: Literal["once", "live"] = "once",
+    ) -> Union[str, "IPython.display.SVG"]:
```
bjchambers, line 1203: We could have special cases for the return based on the `format`, but I don't think we need that strict of types here yet.

crates/sparrow-session/src/session.rs
```diff
@@ -63,6 +70,16 @@ pub struct ExecutionOptions {
...
+impl ExecutionOptions {
+    pub fn per_entity_behavior(&self) -> PerEntityBehavior {
+        match self.results {
+            Results::History => PerEntityBehavior::All,
+            Results::Snapshot if self.final_at_time_s.is_some() => PerEntityBehavior::FinalAtTime,
+            Results::Snapshot => PerEntityBehavior::Final,
+        }
+    }
+}
+
```
jordanrfrazier, line 82: Wonder if we should start thinking about moving some types out of `session.rs` soon

crates/sparrow-session/src/session.rs
```diff
@@ -411,48 +414,86 @@ impl Session {
...
-        let expr = self.dfg.extract_simplest(expr);
-        let expr = expr
+        Ok(self.dfg.extract_simplest(expr))
+    }
+
+    fn optimize_dfg(
+        &self,
+        dfg: sparrow_compiler::DfgExpr,
+    ) -> error_stack::Result<sparrow_compiler::DfgExpr, Error> {
+        let dgg = dfg
```
jordanrfrazier, line 432: should this be `dfg`?

crates/sparrow-session/src/session.rs
```diff
@@ -503,6 +544,49 @@ impl Session {
             schema,
         ))
     }
+
+    pub fn plan(
```
jordanrfrazier, line 548: Hm, the verb `plan` sounds like it could be in the execution path, rather than an option to get the string representation of the plan. Maybe would prefer more explicit `get_string_for_plan`, or something like that, but :shrug:

python/pyproject.toml
jordanrfrazier: Yeah, agreed. `diagnose` sounds good to me, unless we want to be explicit and do a `kaskada[plan]` group?

python/pysrc/kaskada/_execution.py
```diff
@@ -38,6 +41,39 @@ class _ExecutionOptions:
...
+            if results.since is not None:
+                options.changed_since = int(results.since.timestamp())
+            if results.until is not None:
+                options.final_at = int(results.until.timestamp())
+        elif isinstance(results, kd.results.Snapshot):
+            options.results = "snapshot"
+            if results.changed_since is not None:
+                options.changed_since = int(results.changed_since.timestamp())
+            if results.at is not None:
+                options.final_at = int(results.at.timestamp())
```
jordanrfrazier, line 72: Do we want any logic here that asserts `final_at` is greater than `changed_since`?

python/pysrc/kaskada/_ffi.pyi
```diff
@@ -44,6 +44,11 @@ class Expr:
     def session(self) -> Session: ...
     def execute(self, options: Optional[_ExecutionOptions] = None) -> Execution: ...
     def grouping(self) -> Optional[str]: ...
+    def plan(
+        self,
+        kind: Literal["initial_dfg", "final_dfg", "final_plan"],
```
jordanrfrazier, line 49: alternate naming could be `["initial_dfg", "optimized_dfg", "plan"]`?

Though..I wonder if we'd ever like to be able to apply optimizations one by one, and see the progression of the `dfg` for each step. Not sure what that would look like here; if we want that, we'd likely need to swap off the enum and literal.

crates/sparrow-session/src/session.rs
```diff
@@ -43,6 +43,13 @@ pub enum Results {
     Snapshot,
 }
 
+#[derive(Debug, PartialEq, Eq)]
+pub enum QueryPlanKind {
```
jordanrfrazier, line 47: Maybe `QueryRepresentationKind`? Is a `dfg` a type of `QueryPlan`? 

I ask because the `FInalPlan` would correspond to the `ComputePlan` so just want to clarify naming here.

python/pysrc/kaskada/_execution.py
```diff
@@ -38,6 +41,39 @@ class _ExecutionOptions:
     #: For snapshots, this determines when the snapshot is produced.
     final_at: Optional[int] = None
```
jordanrfrazier, line 42: Can you (or I can add later) specify the units for these options

jordanrfrazier (approved):

python/pysrc/kaskada/_timestream.py
bjchambers: @jordanrfrazier any thoughts on whether the format should depend on which packages / environment? I could see a case for not defaulting to `svg` since that only works in IPython notebooks. On the other hand, when I'm in such a notebook I'd really like that to be the default.

python/pyproject.toml
bjchambers: @jordanrfrazier any thoughts on dependencies / packages? I think they'll all accomplish the same end goal.

crates/sparrow-session/src/session.rs
bjchambers: Yeah. Once we switch to logical instead of dfg it should be easier too.

python/pysrc/kaskada/_ffi.pyi
bjchambers: For the plan, I was thinking we'd eventually want to show `initial_plan` and `final_plan`. Definitely could see having more options here, although another way to do the progressino would be to allow the options to control which passes get run, and then view the `final_plan`.

crates/sparrow-session/src/session.rs
bjchambers: I'm considering for now the `dfg` to be approximately a logical plan. Once the DFG goes away, then the equivalent would be choosing between rendering the logical plan and the physical plan.

python/pysrc/kaskada/_execution.py
bjchambers: Let's do that later -- this just moved the code so we could use it for execute and rendering, don't think we should pull in extra changes.

crates/sparrow-session/src/session.rs
bjchambers: Actually, the best option might be `explain`:

https://duckdb.org/docs/guides/meta/explain.html

https://www.geeksforgeeks.org/explain-in-sql/

Different SQLs use different ways to "explain" the query, but that seems like what this is. Basically, you can ask us to "explain" what we'll do to compute something in various ways.

jordanrfrazier: Ah, nice. Yeah, `explain` then

bjchambers: Changed to `explain`.

python/pyproject.toml
bjchambers: Went with `explain` for the function and the optional dependency.

python/pysrc/kaskada/_execution.py
jordanrfrazier: Hm, yeah. Document in the API, then we can enforce it in Sparrow at a single spot before compilation then

crates/sparrow-session/src/session.rs
bjchambers: Went with `ExplanationKind`

python/pyproject.toml
jordanrfrazier: Both `graphviz` and `pydot` look like good options to me. `pygraphviz` looks like the heavier-weight option with manipulation of dot files, that I don't believe we need. 

https://github.com/xflr6/graphviz

Not sure if there's significant difference between graphviz and pydot. Both lightweight, both used about the same, but pydot is pure python, (whereas graphviz is a python interface to the graphviz c library?), not sure if that's a factor to lean towards pydot.

python/pysrc/kaskada/_timestream.py
jordanrfrazier: I wasn't sure the difficulties involved in determining what packages are installed. I think it would be fine to default to `svg`, especially since we're still mainly showing demos and working in notebooks, though.

bjchambers: Will be fixed by switch to graphviz package.

python/pyproject.toml
bjchambers: Switch to `graphviz` -- it has a return type that automatically renders the SVG, but also lets you interact with it to save as a file of different formats, etc. It also seems newer / more maintained.

PR #781, opened 2023-09-28
ref: move prepared files to ~/.cache dir

### LINKED ISSUE #779: bug: Tests running prepare leave the `tables` directory
**Description**
After running the python tests using `pytest`, a `tables` directory is created.

**To Reproduce**
Run the Python tests.

**Actual Behavior**
The `tables` directory is created, full of prepared data from the tests.

**Expected Behavior**
The `tables` directory is created using a temporary directory and/or cleaned up after tests.

jordanrfrazier:
Moves the prepared files directory to `~/.cache/kaskada/`. 

The other alternative is to, for now, create a temporary directory that cleans up after use. While the `~/.cache` is generally used for application-specific files (which is what the prepared files are), we aren't utilizing them and won't until we have state. But, if it's not a large concern to accumulate prepared files in the cache, we might as well start with it here in anticipation of using them. 

Closes #779


bjchambers:
Until we start re-using them, we really should delete them. Otherwise, every run of the test creates files that build up over time, eventually leading to out of disk space. And now it's in a hidden cache directory.

I think:

1. Preparing to a temporary or cache directory and deleting it is the way to go for now.
2. Soon, we want to re-enable prepare to S3 (more storage) and then (possibly) use the local cache directory for things like the rocksdb and a cache of downloaded files... at which point we should manage it like a cache and limit the size, etc.

crates/sparrow-runtime/src/prepare/preparer.rs
```diff
@@ -173,6 +168,18 @@ impl Preparer {
             self.time_multiplier,
         )
     }
+    // Prepared files are stored in the following format:
+    // file:///<temp_dir>/<KASKADA_PATH>/tables/<table_uuid>/prepared/<uuid>/part-<n>.parquet
+    pub fn prepared_output_prefix(&self) -> String {
+        let uuid = Uuid::new_v4();
+        let temp_dir = tempfile::tempdir().expect("failed to create temp dir");
```
jordanrfrazier, line 175: I was thinking about moving this to a `static_init::dynamic` block so we create one temp dir per..sparrow initialization? I wasn't exactly sure when they would be run

jordanrfrazier: But, probably more memory-efficient to just create and drop per query since we're not reusing anyway.

crates/sparrow-runtime/src/prepare/preparer.rs
```diff
@@ -93,13 +93,16 @@ impl Preparer {
...
     /// - This produces multiple parts if the input file is large.
     /// - This produces metadata files alongside data files.
+    /// Parameters:
+    /// - `to_prepare`: The path to the parquet file to prepare.
+    /// - `prepared_dir`: The directory to write the prepared files to.
     pub async fn prepare_parquet(
         &self,
-        path: &std::path::Path,
+        to_prepare: &std::path::Path,
+        prepared_dir: &std::path::Path,
```
bjchambers, line 102: Should this be a directory? We previously prepared directly to S3. I worry that we're backing in some path assumptions that won't hold true when we switch back to using `ObjectStoreUrl`. This should perhaps use that to start with.

crates/sparrow-runtime/src/prepare/preparer.rs
```diff
@@ -185,6 +183,53 @@ impl Preparer {
...
+                + "/"
+                + KASKADA_PATH
+                + "/tables/"
+                + &self.table_config.uuid.to_string()
+                + "/prepared/"
+                + &uuid;
+
+            let url = ObjectStoreUrl::from_str(&url).change_context_lazy(error)?;
+            Ok(url)
+        } else if prefix.is_s3() {
```
bjchambers, line 214: Why do we need to special case `s3`? If it's not local it's an object store. Why not treat that as general?

crates/sparrow-runtime/src/stores/object_store_url.rs
```diff
@@ -51,6 +51,16 @@ impl ObjectStoreUrl {
         self.url.make_relative(&url.url)
     }
 
+    pub fn is_local(&self) -> bool {
+        self.url.scheme() == "file"
+    }
+
+    // TODO: Currently only supports the "s3" scheme
+    // Several others are supported by ObjectStore: https://docs.rs/object_store/latest/object_store/aws/struct.AmazonS3Builder.html#method.with_url
+    pub fn is_s3(&self) -> bool {
```
bjchambers, line 60: Why do we need this?

crates/sparrow-runtime/src/prepare/preparer.rs
```diff
@@ -185,6 +183,53 @@ impl Preparer {
             self.time_multiplier.as_ref(),
         )
     }
+
+    /// Creates the output url prefix to use for prepared files.
+    ///
+    /// e.g. for osx: file:///<dir>/<KASKADA_PATH>/tables/<table_uuid>/prepared/<uuid>
+    ///      for s3:  s3://<path>/<KASKADA_PATH>/tables/<table_uuid>/prepared/<uuid>
+    pub fn prepared_output_url_prefix(
```
bjchambers, line 191: It feels like we're over complicating this. What if we just made the prefix optional -- if it is `None` then we write to the temp dir, if it is `Some(URL)` then the URL must be a directory where we will append the `tables/<table_uuid>/prepared/<uuid>` for preparation purposes.

That seems like it handles the cases:

1. Prepare to the temp dir. `None`.
2. Prepare to a specific local folder. `Some("file://path/to/prepare/in/")`.
3. Prepare to a specific object store folder. `Some("s3://path/to/prepare/in/")`.

crates/sparrow-session/src/session.rs
```diff
@@ -34,17 +35,37 @@ pub struct Session {
...
 }
 
 impl Default for Session {
     fn default() -> Self {
+        // TODO: Support object stores
+        // Likely will need the option to pass in the destination url when executing the
+        // query or creating a table.
+        let temp_dir = tempfile::tempdir().expect("create temp dir");
+        let mut url = "file://".to_owned();
+        url.push_str(temp_dir.path().to_str().expect("valid string"));
```
bjchambers, line 57: I suspect this will be a problem on windows -- paths will have `\` but URLs will want `/`.

You can likely do something with https://docs.rs/object_store/latest/object_store/path/struct.Path.html#method.from_filesystem_path to convert the filesystem path to an `object_store::path::Path` which is what our `ObjectStoreUrl` is based on. Maybe even just add a `ObjectStoreUrl::from_filesystem_dir(...)` or something?

crates/sparrow-session/src/table.rs
```diff
@@ -134,7 +137,7 @@ impl Table {
         Ok(())
     }
 
-    pub async fn add_parquet(&self, path: &std::path::Path) -> error_stack::Result<(), Error> {
+    pub async fn add_parquet(&self, path: &str) -> error_stack::Result<(), Error> {
```
bjchambers, line 140: Should this take `&ObjectStoreUrl`?

crates/sparrow-runtime/src/prepare/preparer.rs
jordanrfrazier: We could, but then we're always passing both the `tempdir` and the `option<url>` everywhere

crates/sparrow-session/src/session.rs
jordanrfrazier: `from_filesystem_dir`: "Convert a filesystem path to a [Path](https://docs.rs/object_store/latest/object_store/path/struct.Path.html) relative to the filesystem root"

I'm not sure how this would help. I can convert the tempdir relative to the root, but then I still need the `file://` prefix for it to be formatted as a URL, unless I misunderstand.

bjchambers: The `object_store::path::Path` is actually a URL (just poorly named). So I believe that it would convert `C:\\path\to\tempfile` into `file://C:/path/to/tempfile`.

crates/sparrow-runtime/src/prepare/preparer.rs
bjchambers: Why? We create the tempdir in the session, and it lives long enough. When you create the session, you take an `prepare_path_prefix: Option<Url>`. You then convert it to: `prepare_path_prefix: ObjectStoreUrl`. Every where else, you just pass `&prepare_path_prefix`:

```
let tempdir = tempdir::NamedTempDir::new();
let prepare_path_prefix = prepare_path_prefix.unwrap_or(|| {
  ObjectStoreUrl::from_filesystem_dir(tempdir.path())
});
Ok(Session { tempdir, prepare_path_prefix })
```

Now we've let the user configure the prepare path prefix, but we only need to pass that everywhere else.

crates/sparrow-session/src/table.rs
jordanrfrazier: Most likely? But for now, the prepare code still uses `SourceData` which uses a string and stuff, so I can change that once we get rid of that code.

bjchambers: Ah. Hmm... seems like that could be a good one to work on soon (relatively self contained).

jordanrfrazier: I'll add a task for it

crates/sparrow-runtime/src/prepare/preparer.rs
```diff
@@ -185,6 +182,36 @@ impl Preparer {
...
+            .change_context_lazy(error)?
+            .join(&self.table_config.uuid.to_string())
+            .change_context_lazy(error)?
+            .join("prepared")
+            .change_context_lazy(error)?
+            .join(&uuid.to_string())
+            .change_context_lazy(error)?;
+
+        // Ensure it's treated as a directory
+        let path = url.std_path().change_context_lazy(error)?;
```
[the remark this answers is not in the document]
jordanrfrazier: I was hoping to do as little manual manipulation of URLs as possible. But if that should work then I'll go with that

crates/sparrow-runtime/src/prepare/preparer.rs
```diff
@@ -185,6 +182,36 @@ impl Preparer {
...
+    ///      for s3:  s3://<path>/<KASKADA_PATH>/tables/<table_uuid>/prepared/<uuid>/
+    pub fn prepared_output_url_prefix(
+        &self,
+        prefix: &ObjectStoreUrl,
+    ) -> error_stack::Result<ObjectStoreUrl, Error> {
+        let error = || Error::InvalidUrl(prefix.to_string());
+
+        let uuid = Uuid::new_v4();
+
+        let url = prefix
```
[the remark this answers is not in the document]
jordanrfrazier: This makes sense. I was wondering if that was acceptable, pattern-wise, or if a specific file should keep all `Errors` the same type.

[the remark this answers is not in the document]
bjchambers: I like to keep them generally the same in a file, but in some cases, it's a *lot* cleaner to say "this method is basically an extension of the underlying `ObjectStoreUrl`, so it makes sense for it to return the same type of error".

bjchambers (approved):

PR #782, opened 2023-09-29
ref: Simplify expressions in physical plan

bjchambers:
Change expressions to always produce the result of the expression. This simplifies evaluation, without affecting the expressiviness:

1. We can now produce a primitive result (such as a boolean or integer) if that is the type of the final expression.
2. We can still produce records by adding a `record` expression to the end (which is the standard way of creating intermediate records).
3. We can still reference parts of a record input using `FieldRef`.

This eliminates the weird "some expressions are marked as output" pattern. It also means that we can operate on `ArrayRef` generally and use `DataType` as the type, instead of sometimes using `RecordBatch` and `SchemaRef`.

Other smaller changes in this PR:
- Move the `Expressions` out of the `StepKind`. This allows the `StepKind` to serve as an identifier of the step, regardless of the expressions it computes, simplifying some of the de-duplication work in the logical-to-physical compilation.
- Rename the `column` expression to `input`, since it accesses the input data to the expression evaluation.
- Addition of a method to "intern" expression names, replacing a `&str` with `&'static str` for expression operators available in the binary.


crates/sparrow-expressions/src/evaluators.rs
```diff
@@ -51,7 +51,8 @@ struct EvaluatorFactory {
 inventory::collect!(EvaluatorFactory);
 
 pub struct StaticInfo<'a> {
-    input_schema: &'a Schema,
+    /// Type of the input that the expressions are run on.
+    input_type: &'a DataType,
```
jordanrfrazier, line 55: Is this similar to the existing `StaticInfo`? 

Would saying `Type of the input that this evaluator accepts` be more accurate?

crates/sparrow-logical/src/expr.rs
```diff
@@ -20,14 +20,14 @@ pub struct Expr {
...
-#[derive(Debug)]
+#[derive(Debug, Hash, PartialEq, Eq, Ord, PartialOrd, Clone, enum_as_inner::EnumAsInner)]
 pub enum Literal {
     Null,
     Bool(bool),
     String(String),
     Int64(i64),
     UInt64(u64),
-    Float64(f64),
+    Float64(decorum::Total<f64>),
```
jordanrfrazier, line 30: Document why we need `decorum`?

crates/sparrow-physical/src/step.rs
```diff
@@ -32,45 +33,59 @@ pub struct Step {
     pub kind: StepKind,
     /// Inputs to this step.
     pub inputs: Vec<StepId>,
-    /// The schema for this step.
-    pub schema: SchemaRef,
+    /// The data type produced by this step.
+    pub data_type: DataType,
```
jordanrfrazier, line 37: Or `result_type/output_type`, so as to not have to look at the docs to clarify that it's not the "input type to this step"?

jordanrfrazier (approved):
Looks good, good simplification of expressions

crates/sparrow-expressions/src/evaluators.rs
bjchambers: I believe so. Agreed the comment/naming isn't great, but I think the suggestion may be more confusing. Specifically:

The weird part is that this isn't "the input types to the evaluator" (those are the arguments to the expression / evaluator). This is the input that the entire set of expressions are being run on.

Consider: `project`

1. This runs on the input batch, so the input type is the `DataType::Struct(...)` describing the input.
2. The `input` expression reads that and produces the `DataType::Struct(...)` corresponding to the input.
3. The `field_ref` expression takes the result of (2) and returns a given field.
4. Another `field_ref` expression takes the result of (2) and returns a different field.
5. An `add` expression takes the results of (3) and (4) and returns the result.

So, the `input_type` in all of their `StaticInfo` is information about "how these expressions are executed" (on the input batch) not the individual evaluator.

crates/sparrow-physical/src/step.rs
bjchambers: Yeah, I like that. Any concern that it may be confusing to have "output_type" of the step and "output_type" of each expression? I guess not.

crates/sparrow-expressions/src/evaluators.rs
bjchambers: Updated this struct to:

```
/// Static information available when creating an evaluator.
pub struct StaticInfo<'a> {
    /// Type of the input that all the expressions are run on.
    input_type: &'a DataType,
    /// Name of the instruction to be evaluated.
    name: &'a Cow<'static, str>,
    /// Literal (static) arguments to *this* expression.
    literal_args: &'a [ScalarValue],
    /// Arguments (dynamic) to *this* expression.
    args: Vec<&'a StaticArg<'a>>,
    /// Result type this expression should produce.
    ///
    /// For many instructions, this should be inferred from the arguments.
    /// It is part of the plan (a) for simplicity, so a plan may be executed
    /// without performing type-checking and (b) because some instructions
    /// need to know the result-type in order to execute (eg., cast).
    result_type: &'a DataType,
}
```

PR #792, opened 2023-10-04
feat: initial logical to physical compile

bjchambers:
This converts logical plans consisting of a tree of reference counted expressions into a physical execution plan.

Also renames the scan operation to `read` and `table_id` to `source_id`.

Work remaining in follow-ups:

- Type-checking of the expressions in the physical plan.
- Simplification pass to rewrite expressions within each step using arithmetic rules.
- Projection pushdown pass to (a) move projections towards sources, allowing us to eliminate unnecessary columns, (b) combine projection steps.
- Predicate pushdown pass to move `select` steps closer to sources.

### DIFF
--- a/crates/sparrow-backend/src/exprs/expr_lang.rs
+++ b/crates/sparrow-backend/src/exprs/expr_lang.rs
@@ -43,10 +43,16 @@ impl egg::Language for ExprLang {
     }
 
     fn matches(&self, other: &Self) -> bool {
-        // Note that this should *only* consider whether the node is the same operation.
+        // Note: As per
+        // https://egraphs-good.github.io/egg/egg/trait.Language.html#tymethod.matches,
+        // this should *only* consider whether the node corresponds to the same
+        // operation.
         //
         // egg itself will check whether the arguments are *equivalent*.
-        self.name == other.name && self.literal_args == other.literal_args
+        //
+        // Some instructions (especially `cast`) depend on the `result_type` to
+        // determine the operation being performed.
+        self.name == other.name && self.literal_args == other.literal_args && self.result_type == other.result_type
     }
 }
 
--- a/crates/sparrow-backend/src/exprs/expr_lang.rs
+++ b/crates/sparrow-backend/src/exprs/expr_lang.rs
@@ -52,9 +52,6 @@ impl egg::Language for ExprLang {
         // Some instructions (especially `cast`) depend on the `result_type` to
         // determine the operation being performed.
 
-        // `(field-ref["foo"] ?base)`
-        // `(cast[i64] ?base)`
-
         self.name == other.name
             && self.literal_args == other.literal_args
             && self.result_type == other.result_type
--- a/crates/sparrow-backend/src/exprs/expr_lang.rs
+++ b/crates/sparrow-backend/src/exprs/expr_lang.rs
@@ -55,7 +55,9 @@ impl egg::Language for ExprLang {
         // `(field-ref["foo"] ?base)`
         // `(cast[i64] ?base)`
 
-        self.name == other.name && self.literal_args == other.literal_args && self.result_type == other.result_type
+        self.name == other.name
+            && self.literal_args == other.literal_args
+            && self.result_type == other.result_type
     }
 }
 
--- a/crates/sparrow-backend/src/exprs/expr_lang.rs
+++ b/crates/sparrow-backend/src/exprs/expr_lang.rs
@@ -1,0 +1,68 @@
+use arrow_schema::DataType;
+use egg::Id;
+use smallvec::SmallVec;
+use sparrow_arrow::scalar_value::ScalarValue;
+
+use crate::Error;
+
+#[derive(Hash, PartialOrd, Ord, PartialEq, Eq, Clone, Debug)]
+pub(crate) struct ExprLang {
+    /// The name of the instruction being applied by this expression.
+    ///
+    /// Similar to an opcode or function.
+    ///
+    /// Generally, interning owned strings to the specific static strings is preferred.
+    pub name: &'static str,
+    /// Literal arguments to the expression.
+    pub literal_args: SmallVec<[ScalarValue; 2]>,
+    /// Arguments to the expression.
+    pub args: SmallVec<[egg::Id; 2]>,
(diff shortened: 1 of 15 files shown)


crates/sparrow-backend/src/exprs/expr_lang.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    fn children_mut(&mut self) -> &mut [egg::Id] {
+        &mut self.args
+    }
+
+    fn matches(&self, other: &Self) -> bool {
+        // Note that this should *only* consider whether the node is the same operation.
+        //
+        // egg itself will check whether the arguments are *equivalent*.
+        self.name == other.name && self.literal_args == other.literal_args
```
jordanrfrazier, line 49: A) Should this also be comparing `args`? 
B) Should this also compare `result_type`? e.g. for `casts` that take the same input but cast to a different type?

bjchambers: A) No -- see the comment here as well as docs for the trait (https://egraphs-good.github.io/egg/egg/trait.Language.html#tymethod.matches).
B) Hmm. Yes, for exactly that reason.

crates/sparrow-backend/src/exprs/expr_pattern.rs
```diff
@@ -0,0 +1,166 @@
...
+            result_type: input_type,
+        });
+        let mut subst = egg::Subst::with_capacity(1);
+        subst.insert(*INPUT_VAR, input_id);
+
+        let result = graph.add_instantiation(&self.expr, &subst);
+
+        let cost_function = egg::AstSize;
+        let extractor = egg::Extractor::new(&graph, cost_function);
+        let (_best_cost, expr) = extractor.find_best(result);
```
jordanrfrazier, line 82: Is there a reason `find_best` is needed here? Is it not just a single substitution in the graph here?

crates/sparrow-backend/src/exprs/expr_pattern.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /// Return true if this contains a single expression for the input.
+    pub fn is_identity(&self) -> bool {
+        // TODO: We may want to make this more intelligent and detect cases where
+        // the expression is *equivalent* to the identity. But for now, we think
+        // we can treat that as an optimization performed by a later pass.
+        let instructions = self.expr.as_ref();
+        instructions.len() == 1 && instructions[0] == egg::ENodeOrVar::Var(*INPUT_VAR)
+    }
```
jordanrfrazier, line 98: Is this true if the expression is a `cast` to a different type?

crates/sparrow-backend/src/logical_to_physical.rs
```diff
@@ -0,0 +1,345 @@
...
+            .map(|reference| reference.step_id)
+            .unique()
+            .sorted()
+            .at_most_one()
+        {
+            Ok(None) => todo!("handle all-literal arguments"),
+            Ok(Some(step)) => {
+                // There is only one step, which means all of the arguments are already
+                // available in the same place, and it is a projection. We just need the
+                // expression IDs of each argument.
```
jordanrfrazier, line 97: Q: Understand why it is a projection

crates/sparrow-backend/src/exprs/expr_pattern.rs
bjchambers: We could instantiate the pattern `ExprPattern` to make an `ExprVec`, but it would require iterating over all of teh nodes in the pattern manually. Using the graph let's us use `add_instantiation` (from EGraph) and also provides de-duplication, etc.

crates/sparrow-backend/src/exprs/expr_pattern.rs
```diff
@@ -0,0 +1,166 @@
...
+        // just returning the resulting `egg::Id` for each argument.
+        let output = graph.add(egg::ENodeOrVar::ENode(ExprLang {
+            name,
+            literal_args,
+            args: arg_ids,
+            result_type: data_type,
+        }));
+
+        let cost_function = egg::AstSize;
+        let extractor = egg::Extractor::new(&graph, cost_function);
```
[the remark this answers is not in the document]
bjchambers: Kind of? The `ExprLang` is passed to `egg::EGraph<ExprLang>` and `egg::RecExpr<ExprLang>`, and tells it what nodes to use. `matches` is used to identify equivalent nodes.

Eg., if we add `add(a, b)` twice we get the same node back.

This will be most useful when we add "simplifications" so we can say "`add(a, b)` is equivalent to `add(b, a)`" and these are recorded as equivalent. Then, `foo(add(a, b))` will automatically be equivalent to `foo(add(b, a))` because it is a matching operator (`foo`) on equivalent arguments.

crates/sparrow-backend/src/exprs/expr_lang.rs
```diff
@@ -0,0 +1,79 @@
...
+        // https://egraphs-good.github.io/egg/egg/trait.Language.html#tymethod.matches,
+        // "This should only consider the operator, not the children `Id`s".
+        //
+        // egg itself will check whether the arguments are *equivalent*.
+        //
+        // Some instructions (especially `cast`) depend on the `result_type` to
+        // determine the operation being performed.
+
+        // `(field-ref["foo"] ?base)`
+        // `(cast[i64] ?base)`
```
jordanrfrazier, line 56: .

crates/sparrow-backend/src/exprs/expr_pattern.rs
```diff
@@ -0,0 +1,190 @@
...
+        name: &'static str,
+        literal_args: smallvec::SmallVec<[ScalarValue; 2]>,
+        args: Vec<ExprPattern>,
+        data_type: arrow_schema::DataType,
+    ) -> error_stack::Result<ExprPattern, Error> {
+        // NOTE: This adds the pattern for each argument to an `EGraph`
+        // to add an instruction. This may be overkill, but does simplify
+        // (a) de-duplicating expressions that appear in multiple arguments
+        // (b) managing things like "all of these arguments should have a
+        // singe input".
```
jordanrfrazier, line 45: single

jordanrfrazier (approved):

PR #799, opened 2023-10-08
Switch to quarto for docs

epinzur:
see readme changes for more info on how to get started with quarto

Also, the community page still needs to be added.


python/README.md
```diff
@@ -33,3 +33,42 @@ Alternatively, install nox and run the tests inside an isolated environment:
...
+  See: https://quarto.org/docs/get-started/
+
+* Generate reference docs
+
+  ```shell
+  nox -s gen-docs
+  ```
+
+  You should re-run this after making any updates to the `pysrc` docstrings.
+  If _Preview Docs_ is running in another shell, the system should auto-refresh with your changes.
```
bjchambers, line 51: Does this mean that I need to run `nox -s gen-docs docs` to first generate the reference docs and then start a preview, but after I don't need to rerun `nox -s gen-docs` to update? If the later, would it make sense for `nox -s docs` to run the equivalent of `gen-docs` internally, so we only need to run a single command?

.github/workflows/ci_python.yml
```diff
@@ -61,13 +61,13 @@ jobs:
           poetry install --only=main --only=lint
       - name: black
         run: |
-          poetry run black --diff pysrc pytests docs/source
```
bjchambers, line 64: Are we sure we don't want to lint the python code in the docs? This seems useful for ensuring that all of the example code we include looks good, etc.

python/README.md
epinzur: you do need to run `nox -s  gen-docs` anytime you want to update the generated docs. this is why they weren't combined into a single command.

run `nox -s docs` in one shell and let it keep running.  make changes to anything outside generated docs, and it will auto refresh.  make changes to python docstrings, run `nox -s gen-docs` and then quarto will auto refresh.

there is also a command to watch for docstrings changes and auto refresh the generation of docs, but I wanted to limit the options a bit.  Also the two executables need to run separately.  I was looking for a way in nox to run two commands simultaneously in parallel, but that doesn't seem to be an option.

.github/workflows/ci_python.yml
epinzur: this is still running on docs source, but I stopped `flake8` due to it complaining about several issues in the `_scripts` folder that are related to how the code was written.  

I'll re-enable `flake8` on `docs` and just mark the `_scripts` as being excluded.

python/README.md
epinzur: the `docs-gen` will only re-write output files that have changed, so that the preview doesn't try to re-render the whole set of reference docs on each update. speeds up iterating on docs.

.github/workflows/ci_python.yml
```diff
@@ -132,28 +132,33 @@ jobs:
...
-        # Needed by sphinx-social-cards.
-        # https://github.com/2bndy5/sphinx-social-cards/blob/main/.github/workflows/build.yml#L54
+      - name: Lint reference docs
         run: |
-          sudo apt-get install -y libgl1-mesa-dev libxkbcommon-x11-0
-          echo "QT_QPA_PLATFORM=offscreen" >> "$GITHUB_ENV"
+          poetry env use 3.11
+          source $(poetry env info --path)/bin/activate
+          poetry install --with=docs
+          pip install ${WHEEL} --force-reinstall
```
epinzur, line 140: @bjchambers should I do this in each sub-task, or is doing it here enough: https://github.com/kaskada-ai/kaskada/pull/799/files#diff-47057ea60e76362be7390920540cdc8bfa927006bad30cd6f59189f05daab815L125

bjchambers: Each task in the job runs in the same environment. So istalling it here should mean that it is available to `Set up quarto`, etc. However, you then `deactivate` the python environment, so it's unclear if that will persist.

python/docs/_extensions/machow/interlinks/.gitignore
```diff
@@ -0,0 +1,3 @@
+*.html
```
bjchambers, line 1: What's the story on the _extensions? Is it basically vendoring of the doc dependencies? It may be helpful to drop a `python/docs/_extensions/README.md` to explain the directory / rationale for what's in here and how/when to update?

python/docs/_reference.yml
```diff
@@ -0,0 +1,200 @@
...
+        - name: Timestream
+          members:
+            - data_type
+            - is_continuous
+        - name: LiteralValue
+          dynamic: false
+        - name: Arg
+          dynamic: false
+    - subtitle: Aggregation
+      desc: |
```
bjchambers, line 30: Idea for cleaning this up -- what if we had a file for this, and made this:

```
- subtitle: Aggregation
  desc: {{< include _aggregation.qmd >}}
  contents:
    ...
```

Benefits:
1. Don't need to know to look in `_reference.yml` to find the description of each section.
2. Can use normal markdown editor since it's a QMD file rather than a string embedded in yaml.

python/docs/_scripts/builder.py
bjchambers, line 1: It may be helpful for there to be a `python/docs/_scripts/README.md` explaining what each of the scripts does / where it came from / etc. Thinking forward, if I needed to make a change, it would be good to know which file its in, etc.

python/docs/_templates/announcement.html
```diff
@@ -0,0 +1,3 @@
+<div class="announcement">
```
bjchambers, line 1: Could maybe have this use a config in the quarto.yml, and only show up if that config is set? Alternatively.. should we just drop the announcement at this point?

python/docs/_templates/card.ejs
```diff
@@ -0,0 +1,30 @@
+```{=html}
+<div class="sd-container-fluid sd-sphinx-override sd-mb-4 gallery-directive docutils">
```
bjchambers, line 2: I suspect most of this is unnecessary -- it was based on code that I based on code that used sphinx-design to render cards. We should probably just use bootstrap cards, and then this (and much/all of `design-style.css` can go away, I think). At least probaly worth a comment here and/or in design-style so we know to clean that up.

python/docs/blog/index.qmd
```diff
@@ -0,0 +1,12 @@
+---
+title: Blog
+listing:
+  contents: posts
+  sort: "date desc"
+  type: default
+  categories: true
+  sort-ui: true
+  filter-ui: true
+  fields: ["title", "date", "author", "subtitle", "description", "reading-time", "categories"] # "image", "image-alt",
```
bjchambers, line 10: commented out fields?

python/docs/examples/bluesky.md
```diff
@@ -12,13 +17,7 @@ You can see the full example in the file [bluesky.py](https://github.com/kaskada
...
 We'll provide a schema and configure the time and entity fields.
 
-```{literalinclude} bluesky.py
-:language: python
-:start-after: "[start_setup]"
-:end-before: "[end_setup]"
-:linenos:
-:lineno-match:
-:dedent: 4
+```{.python include="bluesky.py" code-line-numbers="true" start-line=26 end-line=53 dedent=4}
```
bjchambers, line 20: The fact we need to switch to line numbers isn't great. Maybe worth a feature request and/or PR to the plugin to provide something similar to the tags we used. Those are nice because (a) we can see in the file which parts are included and (b) as we make changes it doesn't break the lines we've chosen to include.

python/docs/examples/bluesky.py
```diff
@@ -26,7 +26,6 @@ async def main():
     # The firehose doesn't (currently) require authentication.
     at_client = AsyncFirehoseSubscribeReposClient()
 
-    # [start_setup]
```
bjchambers, line 29: It may be worth keeping these comments, and hoping to switch back to them later.

1. They won't show up if we choose lines in between the tags.
2. It lets us know what lines are used or not though while we look at the script.

python/docs/examples/reddit.md
bjchambers, line 1: This markdown should change to `.qmd` and have the include files updated, correct?

python/pyproject.toml
```diff
@@ -3,9 +3,7 @@ name = "kaskada"
 version = "0.6.0-a.2"
 description = "Kaskada query builder and local execution engine."
 requires-python = ">=3.9,<4.0"
-classifiers = [
-    "Development Status :: 3 - Alpha",
-]
+classifiers = ["Development Status :: 3 - Alpha"]
```
bjchambers, line 6: Separately, we may want to add more classifiers here. And or change status.

python/pysrc/kaskada/_execution.py
```diff
@@ -3,7 +3,7 @@
 from dataclasses import dataclass
 from typing import AsyncIterator, Callable, Iterator, Literal, Optional, TypeVar, Union
 
-import kaskada as kd
+import kaskada
```
bjchambers, line 6: Hmm. Any reason we need to change the pattern from importing as `kd`? I thought that was our pretty much standard convention?

python/pysrc/kaskada/_timestream.py
```diff
@@ -31,14 +31,18 @@
 if TYPE_CHECKING:
     import graphviz
 
-#: A literal value that can be used as an argument to a Timestream operation.
+# docstring must be under the definition :(
 LiteralValue: TypeAlias = Optional[Union[int, str, float, bool, timedelta, datetime]]
```
bjchambers, line 36: If so, then let's not have a space between definition and docstring (here and below)

bjchambers (approved):

python/docs/_quarto.yml
```diff
@@ -0,0 +1,94 @@
...
+        text: Examples
+      - href: community/index.qmd
+        text: Community
+      - href: reference/index.qmd
+        text: API
+      - href: blog/index.qmd
+        text: Blog
+    right:
+      - icon: slack
+        href: https://join.slack.com/t/kaskada-hq/shared_invite/zt-1t1lms085-bqs2jtGO2TYr9kuuam~c9w
```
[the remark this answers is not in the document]
epinzur: I tried this, it doesn't work.

Both places we specify this URL are already in yml. It seems we can't use the _variable.yml file from other yml config files, only from markdown files.

PR #800, opened 2023-10-09
Quarto tweaks

bjchambers:
Adds the community page

Also:

- Fixes typo in README.md
- Adds error checking to noxfile for missing quarto
- switch to using bootstrap icons over font awesome icons
- update card.ejs to allow for links


python/README.md
```diff
@@ -44,7 +44,7 @@ nox
 * Generate reference docs
 
   ```shell
-  nox -s gen-docs
+  nox -s docs-gen
```
epinzur, line 47: oops

python/noxfile.py
```diff
@@ -149,6 +153,10 @@ def docs(session: nox.Session) -> None:
 @nox.session(name="docs-build", python=python_versions[0])
 def docs_build(session: nox.Session) -> None:
     """Build the docs."""
+    if not shutil.which("quarto"):
+        session.error("Unable to execute `quarto`. Is it installed and in your path?\n"
+                      "See https://quarto.org/docs/get-started/")
+
```
epinzur, line 159: i wish nox had a way to call other nox commands before.  like dependency running this one first. so we could avoid duplication like this.

it has a way to run other nox commands afterwards, but not before.

epinzur (approved):

python/noxfile.py
bjchambers: Yep. I like that it lets us write Python to check things -- makes it easy for cases like this. I like that it makes it easy to run all python versions using virtualenvs. I'm watching https://github.com/wntrblm/nox/issues/453 for dependency support (would be awesome to only build the maturin extension once, for instance).

It looks like https://github.com/rmorshea/noxopt#noxopt may be worth keeping an eye on too -- it would let us register a setup method for a prefix of the commands.

PR #801, opened 2023-10-09
made some quarto tweaks

epinzur:
* removed 1-element sidebar from Community page
* re-added `docs` to linting
* Setup overrides for error F811 for reference generation scripts. https://www.flake8rules.com/rules/F811.html 
* Fixed other linting issues for reference generation scripts

### DIFF
--- a/python/docs/_quarto.yml
+++ b/python/docs/_quarto.yml
@@ -50,9 +50,6 @@ website:
       - guide/execution.qmd
   - title: Examples
     contents: examples
-  - title: Community
-    contents:
-      - community/index.qmd
   - title: API
     style: docked
     contents: reference
--- a/python/docs/_scripts/builder.py
+++ b/python/docs/_scripts/builder.py
@@ -6,7 +6,7 @@ from pathlib import Path
 from typing import Any, Union
 
 from pydantic import ValidationError
-from quartodoc import blueprint, collect, layout, preview
+from quartodoc import blueprint, collect, layout  # , preview
 from quartodoc.inventory import convert_inventory, create_inventory
 from quartodoc.validation import fmt
 from renderer import Renderer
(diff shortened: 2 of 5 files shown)


python/docs/_quarto.yml
```diff
@@ -50,9 +50,6 @@ website:
       - guide/execution.qmd
   - title: Examples
     contents: examples
-  - title: Community
```
bjchambers, line 53: Should we keep the community page? Or is this just removing the side nav for now? I could see a case where we either have a separate "Contributor Guide" or put that under the community page. Something like https://pydata-sphinx-theme.readthedocs.io/en/stable/community/index.html. It could be a place for documentation about how things work internally / how to contribute to them.

python/docs/_scripts/builder.py
```diff
@@ -6,7 +6,7 @@
 from typing import Any, Union
 
 from pydantic import ValidationError
-from quartodoc import blueprint, collect, layout, preview
+from quartodoc import blueprint, collect, layout  # , preview
```
bjchambers, line 9: delete?

python/docs/_scripts/renderer.py
```diff
@@ -108,48 +108,52 @@ def _render_table(self, rows, headers):
     # render_annotation method --------------------------------------------------------
 
     @dispatch
-    def render_annotation(self, el: str) -> str:
+    def render_annotation(self, el: str) -> str:  # noqa: F811
```
bjchambers, line 111: It may be possible to use:

https://beartype.github.io/plum/integration.html

The pattern described there to make this compatible with both `mypy` and `flake8`, if I am reading it correctly.

bjchambers (approved):

python/docs/_quarto.yml
epinzur: this only removes the side nav.  I think it looks funny having a single page with one item on the side nav.

I think adding a contributor guide would be great. we can add back the side bar on this page when we want to add more pages under it.

python/docs/_scripts/builder.py
epinzur: I wanted to keep this since its super helpful for debugging... so I don't forget it exists.  I can move it to a separate line and add a comment about it.

python/docs/_scripts/renderer.py
epinzur: I'll try this.  Great find!

PR #803, opened 2023-10-10
feat: add source trait 

jordanrfrazier:
Adds the `source` trait for sources to implement. The idea being they'll be created as async functions and we can just call next on the returned `stream` -- no need to make them a whole `ReadPipeline` or anything we think. 

* Adds the `sparrow-sources` crate
* Adds the `sparrow-interfaces` crate
* Refactors `batch` into `sparrow-batch`.


crates/sparrow-core/src/read_config.rs
```diff
@@ -0,0 +1,24 @@
+use arrow::datatypes::SchemaRef;
+
+/// Defines the configuration for a read from a source.
+#[derive(Debug)]
+pub struct ReadConfig {
+    /// The schema of the requested data.
+    ///
+    /// The actual schema of the input may have additional columns.
+    pub projected_schema: SchemaRef,
```
jordanrfrazier, line 9: This may not be plausible to have here. This may be just the `prepared_schema`. We'll see how it turns out

crates/sparrow-sources/src/in_memory.rs
```diff
@@ -0,0 +1,221 @@
...
+    fn schema(&self) -> SchemaRef {
+        self.prepared_schema.clone()
+    }
+
+    fn read(
+        &self,
+        read_config: ReadConfig,
+    ) -> futures::stream::BoxStream<'_, error_stack::Result<Batch, sparrow_interfaces::Error>> {
+        // TODO: Do we get any benefit from projecting the schema here?
+        // Or should the prepared schema just match the projected for in-memory sources?
```
jordanrfrazier, line 52: I think we don't need to do field projection here, since we already have the batch in memory and need to read it as a `RecordBatch` anyway -- we can do a `ProjectStep` to get the fields we want afterwards. No need to overload the read by messing with columns/schemas, since (I don't think) there's a significant performance benefit. 

(For parquet files that can read per column it's different and we should do field projection as we read)

crates/sparrow-sources/src/in_memory.rs
```diff
@@ -0,0 +1,221 @@
...
+            futures::stream::empty().boxed()
+        };
+
+        input_stream
+    }
+}
+
+/// Struct for managing in-memory batches.
+#[derive(Debug)]
+pub struct InMemoryBatches {
```
jordanrfrazier, line 85: No changes to this, just moved over.

crates/sparrow-batch/src/batch.rs
```diff
@@ -338,8 +504,7 @@ impl BatchInfo {
         debug_assert!(data.len() > 0);
         debug_assert_eq!(time.null_count(), 0);
 
-        let time_column: &TimestampNanosecondArray =
-            downcast_primitive_array(time.as_ref()).expect("time column to be timestamp");
+        let time_column: &TimestampNanosecondArray = time.as_primitive();
```
bjchambers, line 507: Nice simplifications here.

crates/sparrow-batch/src/batch.rs
```diff
@@ -26,6 +32,50 @@ impl Batch {
...
+        let subsort = data.column(1).clone();
+        let key_hash = data.column(2).clone();
+
+        let schema = data.schema();
+        let fields: Vec<Arc<Field>> = schema.fields()[3..].to_vec();
+        let fields = Fields::from(fields);
+        let columns: Vec<ArrayRef> = data.columns()[3..].to_vec();
+
+        // TODO: ...I can't remember if I need to get the null buffers out for this
+        let data = Arc::new(StructArray::new(fields, columns, None));
```
bjchambers, line 64: You only need the null buffers if you want the struct to have null values, so I think you should be ok.

crates/sparrow-batch/src/batch.rs
```diff
@@ -26,6 +32,50 @@ impl Batch {
...
+        let min_present_time = time.value(0);
+        let max_present_time = time.value(time.len() - 1);
+
+        let time = data.column(0).clone();
+        let subsort = data.column(1).clone();
+        let key_hash = data.column(2).clone();
+
+        let schema = data.schema();
+        let fields: Vec<Arc<Field>> = schema.fields()[3..].to_vec();
+        let fields = Fields::from(fields);
```
bjchambers, line 60: The fields themselves have an `Arc`, so it would probably be better if we pass in the fields (rather than cloning each field separately and making a separate `Fields`). Can probably defer though.

crates/sparrow-batch/src/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+
+mod batch;
+mod error;
+mod row_time;
+
+pub use batch::*;
+pub use error::Error;
+pub use row_time::*;
+
+#[cfg(any(test, feature = "testing"))]
```
bjchambers, line 18: Just to call it out since we discussed. This is a pattern that adds the `testing` module if either:

1. We're in this crate and compiling for tests `#[cfg(test)]`
2. OR, we're outside this crate and have the `testing` feature enabled.

We could then have *other* crates turn on the `testing` dependency if they have the `test` cfg. But either way means that we can have these used in other places, but somewhat clear that it's only for tests.

crates/sparrow-batch/src/row_time.rs
```diff
@@ -33,3 +33,9 @@ impl From<RowTime> for i64 {
         val.0
     }
 }
+
+impl Into<RowTime> for i64 {
```
bjchambers, line 37: See check failure here.

crates/sparrow-core/src/lib.rs
```diff
@@ -11,12 +11,14 @@ use std::pin::Pin;
...
 mod context_code;
 mod key_triple;
+mod read_config;
 mod table_schema;
 
 use arrow::record_batch::RecordBatch;
 pub use context_code::*;
 use futures::Stream;
 pub use key_triple::{KeyTriple, KeyTriples};
+pub use read_config::ReadConfig;
```
bjchambers, line 21: Should `ReadConfig` live in interfaces? What needs it without needing the interfaces (eg., why should it be in `core`)?

Ideally, we'd be moving things out of `sparrow-core` into more appropriate places where possible.

crates/sparrow-core/src/read_config.rs
bjchambers: I don't think it's needed here. Specifically, the logical plan has a `DataType` for what the read produces, which allows it to have the projected `DataType::Struct(projected_fields)`, so we can get that to pass into the read method when needed.

crates/sparrow-core/src/read_config.rs
```diff
@@ -0,0 +1,24 @@
...
+    /// of data available at the time of the read has been processed.
+    pub keep_open: bool,
+    /// Optional timestamp in nanos at which to start reading.
+    ///
+    /// Defaults to the earliest available timestamp.
+    pub start_time: Option<i64>,
+    /// Optional timestamp in nanos at which to end reading.
+    ///
+    /// Defaults to reading until the source is closed.
+    pub end_time: Option<i64>,
```
bjchambers, line 23: I wonder if it would be clearer to combine `keep_open` and `end_time`:

```
pub enum End {
  Never,
  AtTime(i64),
  AtEndOfCurrentData,
```

crates/sparrow-interfaces/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "sparrow-interfaces"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+publish = false
+description = """
+Defines the common interfaces for the Sparrow compilation and runtime.
```
bjchambers, line 9: nit: Can just be `Common interfaces for ..."

crates/sparrow-interfaces/src/source.rs
```diff
@@ -0,0 +1,12 @@
+use arrow_schema::SchemaRef;
+use futures::stream::BoxStream;
+use sparrow_batch::Batch;
+use sparrow_core::ReadConfig;
```
bjchambers, line 4: If `ReadConfig` was here, then it wouldn't need to depend on `sparrow_core`, right?

crates/sparrow-interfaces/src/source.rs
```diff
@@ -0,0 +1,12 @@
...
+
+use crate::Error;
+
+/// Defines how a source provides data to the execution layer.
```
bjchambers, line 8: I think that's just the `read` method. Probably want a different comment on the trait, and then move this to the `read` method (with some more details)?

crates/sparrow-interfaces/src/source.rs
```diff
@@ -0,0 +1,12 @@
...
+pub trait Source: Send + Sync {
+    fn schema(&self) -> SchemaRef;
```
bjchambers, line 10: `projected` or `prepared`?

crates/sparrow-scheduler/Cargo.toml
```diff
@@ -17,6 +17,7 @@ index_vec.workspace = true
 itertools.workspace = true
 serde.workspace = true
 sparrow-arrow = { path = "../sparrow-arrow" }
+sparrow-batch = { path = "../sparrow-batch" }
```
bjchambers, line 20: Just a note for next time -- this probably would have been cleaner if we had a PR to split out the batch stuff and then a separate PR for the sources/interfaces (eg., separate refactoring from introducing new code). Would have made a really fast review for the refactor, and then a smaller review for the new code.

crates/sparrow-sources/src/in_memory.rs
```diff
@@ -0,0 +1,221 @@
...
+use futures::{Stream, StreamExt, TryStreamExt};
+
+use sparrow_batch::Batch;
+use sparrow_core::ReadConfig;
+use sparrow_interfaces::Source;
+use sparrow_merge::old::homogeneous_merge;
+
+use crate::Error;
+
+/// Wraps the mutable in-memory batch container.
```
bjchambers, line 15: I don't love describing something as `Wraps ...` since it means "to understand this you need to go understand that". It's the "See ..." as a dictionary definition.

Can we instead say -- `Shared, synchronized container for in-memroy batches` or something like that?

crates/sparrow-sources/src/in_memory.rs
bjchambers: What I worry about with that is it seems to lead to compilation needing to know if `source.supports_projection()` or something like that. So it may be beneficial to push it down to the read in all cases, so it can implement it as appropriate.

crates/sparrow-sources/src/in_memory.rs
bjchambers: Does this need to be public? Ideally, the niteractions would happen through the `InMemory` struct that provides the synchronized access.

crates/sparrow-sources/src/in_memory.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// Retrieve the current in-memory batch.
+    pub fn current(&self) -> Option<RecordBatch> {
+        let batch = self.current.read().unwrap().batch.clone();
+        if batch.num_rows() == 0 {
+            None
+        } else {
+            Some(batch)
+        }
+    }
+}
```
bjchambers, line 221: Does it seem like a good time to add some tests for this?

crates/sparrow-sources/src/in_memory.rs
```diff
@@ -0,0 +1,221 @@
+use std::sync::{Arc, RwLock};
+
+use arrow_array::RecordBatch;
+use arrow_schema::SchemaRef;
+use error_stack::{IntoReportCompat, ResultExt};
+use futures::{Stream, StreamExt, TryStreamExt};
+
+use sparrow_batch::Batch;
+use sparrow_core::ReadConfig;
```
bjchambers, line 9: Another point -- this only depends on `sparrow_core` for `ReadConfig`. If that was in `sparrow-interfaces` we wouldn't need to depend on it.

crates/sparrow-sources/src/error.rs
```diff
@@ -0,0 +1,21 @@
+#[derive(derive_more::Display, Debug)]
+pub enum Error {
```
bjchambers, line 2: Questions for the future / as this evolves: Should we have per-source errors, or should we have a more general set of possible errors sources produce (possibly defined in sparrow interfaces)? Does it matter?

crates/sparrow-interfaces/src/error.rs
```diff
@@ -0,0 +1,17 @@
+#[derive(derive_more::Display, Debug)]
```
bjchambers, line 1: I wonder if we want a single top-level errors for this crate, vs. a per-trait error. Eg., I could see us having a `SourceError` for errors coming out of sources. It's kind of nice because it lets us say "sources should return this set of errors, which allows anything that calls the source to actually understand on that", rather than having each source define a separate error enum that is opaque to the user.

bjchambers (approved):

crates/sparrow-sources/src/in_memory.rs
jordanrfrazier: It does not. Changed

crates/sparrow-batch/src/batch.rs
```diff
@@ -311,6 +357,126 @@ impl Batch {
...
+    }
+
+    validate_bounds(data.column(0), data.column(1), data.column(2), up_to_time)
+}
+
+#[cfg(debug_assertions)]
+/// Validate that the result is sorted.
+///
+/// Note: This only validates the up_to_time bound.
+pub(crate) fn validate_bounds(
```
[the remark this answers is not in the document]
jordanrfrazier: Yeah, though the implementation seems to be just iterating over the rows and comparing one by one. Likely the `lexsort` will be cleaner, so I'll see about updating that

crates/sparrow-sources/src/in_memory.rs
bjchambers: I think it's good to defer / push to a follow-up.

1. We don't do projection pushdown so it will never need to do it.
2. We're deferring projection pushdown.

So, would suggest (a) adding "implement projection pushdown" to the deferred list wherever we're tracking that and (b) maybe passing the `projected_datatype: &DataType` into the read method (which we'll need to do pushdown), and then you can just do a `assert_eq!(&DataType::Struct(schema.fields.clone()), projected_datatype, "Projection not yet supported");` as the minimum fix here (so we get a runtime assertion to debug)

crates/sparrow-batch/src/batch.rs
jordanrfrazier: Hm..I updated to
```
        let fields = &schema.fields()[3..];
```
but I don't think that that is what you meant. I suppose it avoids allocating for the `vec` at least?

crates/sparrow-core/src/lib.rs
jordanrfrazier: Hm..likely nothing at this point, though I could see a world where we create the `ReadConfig` in `session` (parameters passed from the client), then just passed through until they're used in `sparrow-execution` (or wherever `read/execute` is called)

bjchambers: But session will depend on `sparrow-interfaces` so it can do `add_source(source: Arc<dyn Source>)` right? Which means moving `ReadConfig` to `sparrow-interfaces` would meanthat session doesn't need to depend on `sparrow-core` right?

crates/sparrow-interfaces/src/source.rs
jordanrfrazier: I suspect this is actually just the `raw` schema - when a user creates a source, all we can do is get the schema of the given input. So..I guess it could be the `prepared` schema too. I don't know how this method will be used yet though. The `_ffi.parquet_schema` method currently just gets the schema from the given file. 

`Execution` also has a `schema`, which is used to construct a `table` from batches to go to pandas. So that's the result schema.

crates/sparrow-sources/src/in_memory.rs
jordanrfrazier: That makes sense. updated

crates/sparrow-sources/src/error.rs
jordanrfrazier: This was a question I had when making this. In order to define the stream produced by the `read`, I have to define the `Error`. Thus, all implementors of the trait need to produce a `sparrow_interface::Error`, which is not necessarily bad, I think. If they want, they can chain a specific source error before changing to a more general interface error. 

So -- I think I'm in favor of source specific errors, then converting to general interface errors.

```
parquet.read(..).change_context(sparrow_sources::parquet::Error::SomethingParquetSpecific).change_context(sparrow_interfaces::Error::Read
```

crates/sparrow-interfaces/src/error.rs
jordanrfrazier: Yeah, I'm also not against this idea though. The downside being that if we support many sources, that Error file will get large and perhaps confusing knowing when to use what error. But we can figure that out easily if we want -- prefix parquet sources errors with `Error::Parquet..`, `Error::Pulsar..`, if we want. 

I'm ambivalent here. To keep `sparrow-interfaces` simpler with more general errors, or to simplify errors by having them all in one place :shrug:

crates/sparrow-interfaces/src/source.rs
jordanrfrazier: Pick one? and rename method to match for now

crates/sparrow-sources/src/error.rs
jordanrfrazier: Parquet read already returns a parquet-specific error, so not really a huge need to have sparrow-source/parquet::Error that adds probably not much additional context. 

So

* Get rid of `sparrow_sources::Error`
* Change sparrow-interfaces to SourceError

jordanrfrazier: mark non exhaustive? Adding or deleting case would be compile error. If marked, they have to have a catch all case.

crates/sparrow-core/src/lib.rs
jordanrfrazier: `ReadConfig` would theoretically be unpacked into its component parts, so it makes sense that it could live within the `read` method. `source.rs` trait.

jordanrfrazier:
TODO: add task for adding machete to ci

crates/sparrow-core/src/read_config.rs
jordanrfrazier: add to doc and comment about keepopen/endtime

crates/sparrow-batch/src/batch.rs
jordanrfrazier: TODO: add to doc, file issue. 

Issues for acceptance

crates/sparrow-batch/src/batch.rs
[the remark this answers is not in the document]
jordanrfrazier: Added a follow up actually, should be easy, but want to get this merged then look at your e2e PR.

crates/sparrow-interfaces/src/error.rs
bjchambers: I'd lean towards the source errors being more general so it shouldn't matter (as much) what sources we support. Eg., it's almost like HTTP codes -- `NotFound` could be used by ParquetSource to say "s3 bucket doesn't exist" or PulsarSource to say "topic doesn't exist". This then let's us surface to the user something meaningful (failed because the source couldn't be found).

Issue #787, opened 2023-10-02
Wrapper function/type for creating Scalars


bjchambers:
Could add a helper for creating the static somewhere, seen it in a few instances now

_Originally posted by @jordanrfrazier in https://github.com/kaskada-ai/kaskada/pull/784#discussion_r1342918969_

bjchambers:
Arrow recently added [`Scalar`](https://docs.rs/arrow/latest/arrow/array/struct.Scalar.html) which simplifies many of the APIs handling scalar values. We could potentially use this to replace our [`ScalarValue`](https://github.com/kaskada-ai/kaskada/blob/main/crates/sparrow-arrow/src/scalar_value.rs) enum.

See https://github.com/apache/arrow-rs/pull/4793/files for some notes:

1. We can use `Int64Array::new_scalar(...)` to simplify creating a scalar of a specific type.
2. We can create a `Scalar<ArrayRef>` for using an array as a scalar.

This should let us create something like:

```
struct ScalarValue(Scalar<ArrayRef>);

// 1. Serde the underlying ArrayRef
// 2. Support use as a scalar anywhere
```

bjchambers:
@jordanrfrazier FYI -- I think we should be able to completely replace our ScalarValue enum with `Scalar<ArrayRef>`. I think it's probably worth doing that for how we serialize scalars in the `physical plan`.

Thoughts on:

A) Trying to change the `ScalarValue` to use that representation (potentially messy since we have that protobuf in many places0.
B) Introducing a parallel `Scalar???` to represent that?
C) Just using `arrow_array::Scalar<arrow_array::ArrayRef>` in those places?

I think with `C` we'd need to do the custom serialization like we did for structs with an `ArrayRef`, while with `A` or `B` we could encapsulate that in our serialization of the wrapper type we create.

PR #806, opened 2023-10-11
feat: compile and execute partitioned query

bjchambers:
The main change here is having the `sparrow-execution` tests compile
from a logical plan, created within a session, and execute things.

Changes Include

- Fix naming `field_ref` vs. `fieldref` (matches Python)
- Tweak pipeline scheduling so that only transforms are grouped together.
- Tweak pipelines so that each may output to multiple consumers
- Introduces a `SourceTasks` so we can poll all inputs on a single
  tokio task.
- Fixes to how the scheduler detects "completion" (empty queue isn't
  sufficient).
- Add a "monitor" to detect failing threads ASAP
- Add a "worker idle" so that workers with no tasks sleep.


crates/sparrow-backend/src/logical_to_physical.rs
```diff
@@ -50,7 +50,12 @@ pub(super) struct LogicalToPhysical {
 #[derive(Debug)]
 struct Reference {
     /// Reference to a specific step.
-    step_id: sparrow_physical::StepId,
+    ///
+    /// This should only be `None` if the expression is a literal. It should
+    /// not be `None` if the expression is a literal arising within a specific
+    /// step (domain) -- only for cases where the value originally has no
+    /// domain.
+    step_id: Option<sparrow_physical::StepId>,
```
bjchambers, line 58: During compilation, literals don't belong to any step. This actually helps some with some of the weird literal behavior.

1. When a literal is first converted, we say `step_id=None, expr=[CodeForLiteral]`.
2. When that is used (merged) with other arguments, we add the expression to the merged step (which may be a specific step if all arguments are in the same step or following a merge if there is one).

But at that point, we have a reference to that literal in a specific step.

crates/sparrow-backend/src/logical_to_physical.rs
```diff
@@ -216,11 +248,32 @@ impl LogicalToPhysical {
                     )));
                 };
 
-                let args = node
-                    .args
-                    .iter()
-                    .map(|arg| self.visit(arg))
-                    .collect::<Result<_, _>>()?;
+                let args = if instruction != "record" {
```
bjchambers, line 251: Records are annoying. I'd like to clean this up, but I'm not sure yet on the best way.

1. A record in the logical expression has `(record name1 value1 name2 value2 ...)` because Python doesn't distinguish between "literal arguments" and "arguments".
2. A record in the physical expression has `(record[name1, name2, ...] value1 value2 ...)`.

My current thinking is that I might add a trait in the `sparrow-interfaces` crate that is basically "how does the expression typecheck (logically and physically) and how does it convert", with a default implementation for things not registered. Not sure yet.

crates/sparrow-backend/src/logical_to_physical.rs
```diff
@@ -216,11 +248,32 @@ impl LogicalToPhysical {
...
+                    // it would give us a place to see and test that the result of creating the
+                    // logical node can be turned into physical nodes, etc.
+                    node.args
+                        .iter()
+                        .skip(1)
+                        .step_by(2)
+                        .map(|arg| self.visit(arg))
+                        .collect::<Result<_, _>>()?
+                };
+                println!("Resolving for {other}: {args:?}");
```
bjchambers, line 276: println.

crates/sparrow-backend/src/logical_to_physical.rs
```diff
@@ -296,6 +359,24 @@ impl LogicalToPhysical {
     }
 }
 
+fn logical_to_physical_literal(
```
bjchambers, line 362: I'll probably file an issue -- I think we can replace our `ScalarValue` with an arrow `Scalar(ArrayRef)`, at which point we can have Arrow serialize them, delete all that code, and make it a lot easier to represent literals and scalars. We'll see.

crates/sparrow-backend/src/pipeline_schedule.rs
```diff
@@ -49,29 +49,42 @@ pub fn pipeline_schedule(steps: &IndexSlice<StepId, [Step]>) -> Vec<Pipeline> {
...
             true
         }
-        _ if references[step.inputs[0]] > 1 => {
+        _ if !steps[step.inputs[0]].kind.is_transform() => {
             debug_println!(
                 DEBUG_SCHEDULING,
-                "Step {index} is new pipeline since it's only input ({}) is referenced {} times",
-                step.inputs[0],
-                references[step.inputs[0]]
+                "Step {index} is a new pipeline since it's input is not a trasnform"
```
bjchambers, line 78: trasnform

crates/sparrow-backend/src/pipeline_schedule.rs
```diff
@@ -49,29 +49,42 @@ pub fn pipeline_schedule(steps: &IndexSlice<StepId, [Step]>) -> Vec<Pipeline> {
 }
 
 /// Return true if the step is "pipeline breaking".
-fn is_pipeline_breaker(index: StepId, step: &Step, references: &IndexVec<StepId, usize>) -> bool {
+///
+/// A "pipeline breaker" must start a new pipeline.
+fn is_pipeline_breaker(
```
bjchambers, line 54: Changed the scheduling a bit. Now, a sequence of transforms will become a pipeline. If one of them has multiple consumers, then it will end the pipeline.

crates/sparrow-backend/src/pipeline_schedule.rs
```diff
@@ -49,29 +49,42 @@ pub fn pipeline_schedule(steps: &IndexSlice<StepId, [Step]>) -> Vec<Pipeline> {
...
+    index: StepId,
+    step: &Step,
+    references: &IndexVec<StepId, usize>,
+    steps: &IndexSlice<StepId, [Step]>,
+) -> bool {
     match &step.kind {
+        _ if !step.kind.is_transform() => {
+            debug_println!(
```
bjchambers, line 62: Should this use `debug_println!(...)` or `tracing::trace!(...)`?

crates/sparrow-execution/src/lib.rs
```diff
@@ -8,254 +8,193 @@
...
-        assert_eq!(
-            abc.as_primitive(),
-            &Int64Array::from(vec![Some(25), None, Some(399), Some(101)])
-        );
-    }
+const DEBUG_EXECUTOR_CREATION: bool = true;
 
-    /// Execute a physical plan.
-    pub async fn execute(
+impl PlanExecutor {
```
bjchambers, line 40: This is one of the major changes. This struct (don't love the name) let's us create an executor (using the `sparrow-scheduler::WorkerPool`) from a physical plan. See the tests for it's use.

crates/sparrow-execution/src/lib.rs
```diff
@@ -8,254 +8,193 @@
...
+    ///
+    /// This creates the worker threads but does not inject inputs, meaning
+    /// the threads won't have anything to do until `run` is called.
+    pub fn try_new(
         query_id: String,
-        input_type: DataType,
-        mut input: tokio::sync::mpsc::Receiver<Batch>,
-        output_type: DataType,
+        plan: sparrow_physical::Plan,
+        mut inputs: HashMap<Uuid, tokio::sync::mpsc::Receiver<Batch>>,
```
bjchambers, line 48: Inputs are provided as a map from uuid to receiver. This will change with your PR to be UUID to `Arc<dyn Source>`.

crates/sparrow-execution/src/source_tasks.rs
```diff
@@ -0,0 +1,65 @@
...
+use futures::stream::BoxStream;
+use futures::StreamExt;
+use sparrow_arrow::Batch;
+use sparrow_scheduler::{Injector, PipelineInputs};
+
+use crate::Error;
+
+/// Manages the async streams that must be read for input.
+#[derive(Default)]
+pub(super) struct SourceTasks {
```
bjchambers, line 11: Added this to collect the source streams and pipeline inputs for the purposes of running them in tokio.

crates/sparrow-execution/src/source_tasks.rs
```diff
@@ -0,0 +1,65 @@
...
+    ) {
+        let index = self.reads.len();
+        self.read_inputs.push(sink);
+        self.reads
+            .push(stream.map(move |batch| (batch, index)).boxed());
+    }
+
+    /// Execute all of the sources.
+    ///
+    /// This currently joins them and runs them directly. Depending on the amount
```
bjchambers, line 39: I suspect we'll need to change this. I believe that some sources (eg., Parquet) are potentially CPU intensive, so we probably want each source/stream to be run as a separate tokio task.

crates/sparrow-execution/src/tests.rs
```diff
@@ -0,0 +1,97 @@
+use arrow_array::cast::AsArray;
+use arrow_array::{Int64Array, StructArray, TimestampNanosecondArray, UInt64Array};
+use hashbrown::HashMap;
+use std::sync::Arc;
+
+use arrow_schema::{DataType, Field, Schema};
+use sparrow_arrow::{Batch, RowTime};
+
+#[test]
+fn test_logical_query() {
```
bjchambers, line 10: Main test showing how it fits together -- create a tree from `sparrow-logical`, compile it, execute it.

crates/sparrow-execution/src/lib.rs
bjchambers: It also feels weird (maybe?) that the usage is:

```
PlanExecutor::try_new(query_id, plan,...).execute()
```

crates/sparrow-logical/src/typecheck.rs
```diff
@@ -31,6 +32,37 @@ pub(crate) fn typecheck(name: &str, args: &[ExprRef]) -> error_stack::Result<Typ
             };
             Ok(types)
         }
+        "record" => {
```
bjchambers, line 35: Added special type-checking for records. This is part of why I'd like to make this a trait -- ideally, the type-checking logic is defined in the same place we specify how to map logical to physical.

crates/sparrow-physical/src/plan.rs
```diff
@@ -15,6 +15,27 @@ pub struct Plan {
...
+impl Plan {
+    pub fn yaml(&self) -> impl std::fmt::Display + '_ {
+        PlanYaml(self)
+    }
+}
+
+/// Prints a plan as Yaml.
+struct PlanYaml<'a>(&'a Plan);
+
+impl<'a> std::fmt::Display for PlanYaml<'a> {
```
bjchambers, line 27: Added this as a quick helper for rendering plans as yaml.

crates/sparrow-scheduler/src/pending.rs
```diff
@@ -0,0 +1,122 @@
...
+use std::sync::Arc;
+
+use hashbrown::HashMap;
+use parking_lot::Mutex;
+
+use crate::Partition;
+
+/// Track the number of pending pipeline partitions.
+#[derive(Default)]
+pub(crate) struct PendingSet {
```
bjchambers, line 11: Pulled this in from the prototype and added some logic to the scheduler as noted in worker. Just because the queue is empty doesn't mean we're "done". So instead, we need to track "which partitions are still running".

We *could* perhaps just do this for sources (make it part of the source tasks) but it seems nice to be able to say "which pipelines aren't completed" to diagnose things.

crates/sparrow-scheduler/src/queue.rs
```diff
@@ -49,25 +51,41 @@ impl<T> GlobalQueue<T> {
...
     pub(crate) fn push_yield(&mut self, item: T) {
+        debug_println!(
+            crate::DEBUG_SCHEDULER,
+            "Pushing {item:?} for (deferred) local execution"
+        );
         self.queue.push_yield(item)
     }
 
     pub(crate) fn push_global(&self, item: T) {
+        debug_println!(
```
bjchambers, line 85: I can probably delete these... it was for debugging some very specific things (why we were immediately ending, which turned out to just be because the worker loop immediately ended since the queue was empty)

crates/sparrow-scheduler/src/sink.rs
```diff
@@ -11,19 +11,69 @@ pub struct PipelineInput {
...
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    pub fn add_consumer(&mut self, pipeline: Arc<dyn Pipeline>, input: usize) {
+        self.0.push(PipelineInput::new(pipeline, input));
```
bjchambers, line 27: Added this -- can probably delete or hide `PipelineInput` but basically needed to be able to say "this pipeline has multiple consumers"

crates/sparrow-scheduler/src/task.rs
```diff
@@ -81,10 +78,21 @@ impl Task {
...
+            .do_work(self.pending.partition, scheduler)
             .change_context_lazy(|| self.error("do_work"))?;
         Ok(guard.finish())
     }
+
+    /// Mark this task as completed.
+    ///
+    /// After this, it should not be scheduled nor should work be done.
+    pub fn complete(&self) {
+        self.pending.complete()
```
bjchambers, line 90: This should let the existing operations be simplified, since the task now tracks completion (previously each operation tracked completion itself).

crates/sparrow-scheduler/src/worker.rs
```diff
@@ -71,15 +76,38 @@ pub struct Worker {
...
+                    );
+                    self.queue.push_global(task);
+                }
+            }
+
+            let pending_count = pending_set.pending_partition_count();
+            if pending_count == 0 {
+                break;
+            } else {
+                // Right now, this "busy-waits" by immediately trying to pull more work.
```
bjchambers, line 103: Note this: the busy waiting is definitely not ideal. I may take a quick stab at fixing it in this PR, if not it should be a fast-follow-up.

crates/sparrow-scheduler/src/worker_pool.rs
```diff
@@ -12,71 +13,39 @@ const DEFAULT_THREAD_COUNT: usize = 8;
...
     injector: Injector,
-    handles: Vec<std::thread::JoinHandle<error_stack::Result<(), Error>>>,
+    workers: Vec<Worker>,
     /// A vector of the pipelines we created.
     pipelines: Vec<Arc<dyn Pipeline>>,
+    /// Track which pipelines / partitions are still running.
+    pending: Arc<PendingSet>,
 }
 
 impl WorkerPool {
```
bjchambers, line 26: The name `WorkerPool` seems incorrect. At this point:

1. It creates the workers
2. We add the pipelinesn to it
3. We tell it to start, which creates the threads and runs the pipelines on it.

So, it feels more like `WorkerBuilder` or something like that. Thoughts?

crates/sparrow-execution/src/tests.rs
```diff
@@ -0,0 +1,97 @@
...
+        input_data,
+        time,
+        subsort,
+        key_hash,
+        RowTime::from_timestamp_ns(3),
+    );
+    input_tx.blocking_send(input_batch).unwrap();
+    std::mem::drop(input_tx);
+
+    session.rt.block_on(executor.execute()).unwrap();
```
bjchambers, line 86: I don't like this.

Originally the test was `tokio::test` (which meant it got a runtime), but then the runtime inside the created session was getting dropped.

I think there are two options:

1. Codify something about what is done here -- instead of `session.rt.block_on(executor.execute())`, I can just add the code to session to take a `sparrow_logical::Expr` and execute it. Then I don't need to expose the runtime here. That is probably the best option.
2. Add a method to create a session using a specific runtime or runtime handle. Then pass in the current runtime to the session. Don't love that due to the unclear dropping of the runtime.

crates/sparrow-execution/src/write_channel_pipeline.rs
```diff
@@ -0,0 +1,75 @@
+use error_stack::{IntoReport, ResultExt};
+use parking_lot::Mutex;
+use sparrow_arrow::Batch;
+use sparrow_scheduler::{Pipeline, PipelineError};
+
+/// Pipeline for writing to a Tokio channel.
+///
+/// This is currently a place holder for a more complete write solution.
+#[derive(Debug)]
+pub(super) struct WriteChannelPipeline {
```
bjchambers, line 10: This was just moved out of the test.

crates/sparrow-scheduler/src/sink.rs
bjchambers, line 1: name of this file sucks... thoughts?

crates/sparrow-execution/src/lib.rs
```diff
@@ -8,254 +8,193 @@
...
+                .steps
+                .iter()
+                .all(|id| plan.steps[*id].kind.is_transform())
+            {
+                debug_assert_eq!(
+                    first_step_inputs.len(),
+                    1,
+                    "Transforms should have a single input"
+                );
+                let first_step_input_id = first_step_inputs[0];
```
bjchambers, line 93: Could have `add_transform_pipeline` determine this -- would keep the main code simpler.

crates/sparrow-execution/src/lib.rs
```diff
@@ -8,254 +8,193 @@
...
+                )?
+            } else {
+                assert_eq!(
+                    pipeline.steps.len(),
+                    1,
+                    "Non-transforms should be in separate pipelines"
+                );
+
+                let step = &plan.steps[pipeline.steps[0]];
+                if let Some(pipeline) =
```
bjchambers, line 110: Theoretically, we could instead add a "noop" or pass-through pipeline for the source. This would have the benefit of every pipeline in the plan having a corresponding executable pipeline. It could also make it easier to add logic that hapens within the worker threads for sources in the future (and/or have a "pull based source model", for instance)...

crates/sparrow-scheduler/src/pending.rs
bjchambers: Thoughts on complexity vs. value of tracking "what's blocking/outstanding"? It seems like we could just wait for the sources to be done and then wait for the queues to be empty.

bjchambers: Main difficulty is figuring out how the pipelines communicate "done" back to the worker pool / manager.

This currently says: "each pipeline receives a `PendingPartition` instance which it can use to determine (a) if it's still pending and (b) report that it is complete. The worker then waits for all partitions to be done and can print which partitions (or the earliest pipelines) holding things up.

Alternatives:

1. Wait for all sources to be done then wait for the queue to be empty. This would require sources communicating to the worker pool that they are complete, but that shouldn't be too difficult since sources already exist outside the pool.
2. Have a special "PipelineEnd" that is a pipeline which when closed (or all partitions closed) signals the worker pool it's done. Have the `Write` actually close the pipeline end as it goes.

crates/sparrow-scheduler/src/worker.rs
bjchambers: Changed to idle the thread and wake it up when work is queued.

crates/sparrow-interfaces/src/source.rs
```diff
@@ -21,12 +23,12 @@ pub trait Source: Send + Sync {
     fn read(
         &self,
         projected_datatype: &DataType,
-        read_config: ReadConfig,
-    ) -> BoxStream<'_, error_stack::Result<Batch, SourceError>>;
+        read_config: Arc<ReadConfig>,
+    ) -> BoxStream<'static, error_stack::Result<Batch, SourceError>>;
```
bjchambers, line 27: Changed this to return `'static`. This allows the stream to be used without worrying about the lifetime of the source. This allows (eg.,) borrowing the source to start execution, and then getting a free-standing reader from it.

bjchambers: @jordanrfrazier fyi

crates/sparrow-execution/src/tests.rs
```diff
@@ -1,263 +1,106 @@
...
+    )
+    .unwrap();
+    session.rt.block_on(source.add_batch(input_batch)).unwrap();
+
+    // Create the exececutor after the data is added.
+    // Currently, we snapshot the data when performing a non-live query.
+    //
+    // TODO: Support live query (via configuration).
+    let executor = {
+        let mut sources: HashMap<_, &dyn Source> = HashMap::with_capacity(1);
```
bjchambers, line 90: Having to make this hashmap to pass to the execution was a little gross, since we'll likely need to have a `HashMap<_, Arc<dyn Source>>` else where for the sources we know about. Figuring out how this interacts with Python is probably a good follow-up.

crates/sparrow-batch/src/batch.rs
```diff
@@ -77,6 +77,23 @@ impl Batch {
...
+                RecordBatch::try_new(schema, columns).expect("create_batch")
+            } else {
+                RecordBatch::try_new(
+                    schema,
+                    vec![data.time, data.subsort, data.key_hash, data.data],
+                )
+                .expect("create_batch")
+            }
+        })
+    }
```
jordanrfrazier, line 95: Think we would want to store the `schema` in the `Batch`? Seems like it could be helpful for this case and might as well if we have it in an `Arc`.

crates/sparrow-execution/src/lib.rs
```diff
@@ -8,5 +8,228 @@
...
+    pub fn try_new(
+        query_id: String,
+        plan: sparrow_physical::Plan,
+        sources: &HashMap<Uuid, Arc<dyn Source>>,
+        output: tokio::sync::mpsc::Sender<RecordBatch>,
+    ) -> error_stack::Result<Self, Error> {
+        let mut executor = PlanExecutor {
+            worker_pool: WorkerPool::new(query_id).change_context(Error::Creating)?,
+            source_tasks: SourceTasks::default(),
+            read_config: Arc::new(ReadConfig {
```
jordanrfrazier, line 78: TODO: to plumb these through?

crates/sparrow-execution/src/lib.rs
```diff
@@ -8,5 +8,228 @@
...
+                    first_step_input_id,
+                    &plan,
+                    &pipeline.steps,
+                    consumers,
+                )?
+            } else {
+                assert_eq!(
+                    pipeline.steps.len(),
+                    1,
+                    "Non-transforms should be in separate pipelines"
```
jordanrfrazier, line 140: Would "Non-transforms should have a single step" be more descriptive here?

crates/sparrow-execution/src/lib.rs
jordanrfrazier: I think that's a good idea. Though, in that case, would it make sense to just have a full `ReadPipeline` with minimal logic?

crates/sparrow-execution/src/lib.rs
bjchambers: I guess it's between this and worker pool, the combination.

1. `WorkerPool` is what lets us add pipelines and execute them. But even that is somewhat split between the "builder" phase and the "execution" phase.
2. `PlanExecutor` just takes the plan, creates the appropriate pipelines, and adds them to the worker pool.

Thinking about the naming together, I wonder if we'd be better with:

1. `WorkerPool` is renamed to `ExecutorBuilder` and `Executor` for the "adding pipelines to a thing" and then "the thing that waits for the pipelines to finish".
2. Then maybe this stays as `PlanExecutor` or `PlanExecutorBuilder` in that it is a layer on top of `ExecutorBuilder` for dealing with plans?

crates/sparrow-backend/src/logical_to_physical.rs
```diff
@@ -260,20 +319,21 @@ impl LogicalToPhysical {
...
-            result.step_id
+            result_step_id
         } else {
             // Add a projection for the remaining expressions.
             //
             // We could attempt to add them to the step if it is already a projection, but we'd
             // need to deal with mutating the expressions in the projection. For now, we think it
             // is easier to treat steps as immutable once created, and then combine steps in a
             // later optimization pass of the physical plan.
+            let input_type = self.plan.step_result_type(result_step_id).clone();
```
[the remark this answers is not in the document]
bjchambers: Yeah -- here the `input_type` is the "input to the projection" which is applied to the `result_step`. Therefore, the `input_type` is the result type of the result step. What clarification would you suggest? Where could it be elaborated?

crates/sparrow-backend/src/logical_to_physical.rs
```diff
@@ -132,27 +150,32 @@ impl LogicalToPhysical {
...
+                                // step ID being merged. This may change depending
+                                // on (a) optimizations that change steps and
+                                // whether it is difficult to update these and (b)
+                                // how we choose to implement merge.
+                                //
+                                // It may be more practical to use the "index of the
+                                // step ID in the inputs" which would be more stable
+                                // as we change step IDs.
+                                smallvec![ScalarValue::Utf8(Some(format!("step_{}", input_step)))],
+                                smallvec![egg::Id::from(0)],
```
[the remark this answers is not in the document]
bjchambers: Basically, we can choose to use the step ID or the input index. The step ID is "nice" because it is stable across input re-orderings, or if we were to combine merges. On the other hand, the "input index" is nice, because it is stable as we rewrite steps and change step IDs. I suspect we'll get more benefits from using the input index here, but we can adapt that later.

And yeah, `Id::from` is an argument that references the first expression is the input (since we used `ExprPattern::new_input()`).

crates/sparrow-backend/src/logical_to_physical.rs
```diff
@@ -132,27 +150,32 @@ impl LogicalToPhysical {
...
+                                data_type,
+                            )?;
+
+                            // Then add the actual expression, replacing the `?input` with the fieldref.
+                            let mut subst = egg::Subst::with_capacity(1);
+                            subst.insert(*crate::exprs::INPUT_VAR, egg::Id::from(1));
```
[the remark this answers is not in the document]
bjchambers: Yeah. Cleaned up to use some variables to hopefully help.

crates/sparrow-backend/src/pipeline_schedule.rs
```diff
@@ -153,13 +165,16 @@ mod tests {
...
                 Pipeline {
                     steps: vec![3.into()]
                 },
                 Pipeline {
-                    steps: vec![5.into()]
+                    steps: vec![4.into(), 5.into()]
+                },
+                Pipeline {
+                    steps: vec![6.into()]
                 }
```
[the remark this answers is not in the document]
bjchambers: Currently, in the plan, the last pipeline is the one which we want to output. We then *add* a "write" pipeline after that when we run the query, based on the right that was requested. This is subject to change.

crates/sparrow-backend/src/snapshots/sparrow_backend__logical_to_physical__tests__logical_to_physical_arithmetic-3.snap
```diff
@@ -59,7 +59,7 @@ steps:
       - name: input
         literal_args: []
         args: []
-        result_type: Float64
+        result_type: "Null"
```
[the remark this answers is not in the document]
bjchambers: I think it's because the merge currently doesn't set the merged type correctly in `logical_to_physical`. See:

```
let merged_step = self.plan.get_or_create_step_id(
    sparrow_physical::StepKind::Merge,
    inputs.clone(),
    ExprVec::empty(),
    DataType::Null,
);
```

crates/sparrow-batch/src/batch.rs
bjchambers: Maybe? In a lot of places where we have a batch we currently don't already have a schema. Basically, schemas are only useful when we have a batch. I actually think that I'd like to change sources and sinks to operate on `Batch` rather than `RecordBatch` at which point we wouldn't need to get the schema here.

Specifically, consider:

1. ReadParquet (`RecordBatch`)
2. Convert to Batch (`StructArray`, we have a `DataType` but no `SchemaRef`)
3. Run a `Project` (we have a `DataType` but no `SchemaRef`)
...

By operating on `DataType` and `ArrayRef` we allow reading in a primitive (single column) and writing it. Any place we use a `RecordBatch` we *need* to have named columns. Standardizing on `DataType` and `ArrayRef` allow us to operate on primitives as well as we can operate on a struct array. So I think I'd actually like to go the other way. Thoughts?

crates/sparrow-execution/src/lib.rs
```diff
@@ -8,5 +8,228 @@
...
+
+pub use error::*;
+
+use crate::source_tasks::SourceTasks;
+use crate::write_channel_pipeline::WriteChannelPipeline;
+
+pub struct PlanExecutor {
+    worker_pool: WorkerPool,
+    source_tasks: SourceTasks,
+    read_config: Arc<ReadConfig>,
```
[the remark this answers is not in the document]
bjchambers: I don't think we would. How would two sources in the same query have state at different times?

`(sourceA.col(x) + sourceB.col(y)).sum()`

The aggregation state is over a combination of `sourceA` and `sourceB`.

Similarly, the options in `ReadConfig` are actually part of the `ExecutionConfig` passed in. I suspect:

a. These options should be part of `ExecutionConfig` and just passed to the plan executor in an `Arc`.
b. There may be other options that *are* read specific (eg., predicate pushdown, environment permissions, etc.)

crates/sparrow-execution/src/lib.rs
bjchambers: I don't think so. The error here is that right now, we can't execute multiple non-transform steps as part of a single pipeline, since each of those non-transform steps is a separate pipelien. Given that the *step* is a transform or non-transform, it seems confusing to say "non transforms should have a single step"... every transform step is a single step as is every non-transform step.

crates/sparrow-execution/src/lib.rs
bjchambers: I can probably plumb it through pretty easily since it comes from the ExecutionConfig.

crates/sparrow-scheduler/src/worker_pool.rs
jordanrfrazier: I still don't dislike `WorkerPool` and `PlanExecutor`. 

`PlanExecutor` takes the plan, then offers an `execute` method to begin processing. I get your point from the other comment about how it's mainly adding pipelines to the worker pool. But I still like it as the `PlanExecutor` because from the caller's perspective it's the thing I call `execute` on. If it were a builder, I'd expect the ol pattern
```
PlanBuilder builder = new PlanBuilder();
builder.add_pipeline(..);
PlanExecutor executor = builder.build();
executor.execute()
```

As for `WorkerPool`, I like it because it hints to me we're working directly with threads (thread pool).

I do want to look at how `Pipelines` and `Tasks` are created to understand the necessity for the cyclic reference though. That looks potentially hard to debug

crates/sparrow-backend/src/logical_to_physical.rs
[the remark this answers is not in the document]
jordanrfrazier: Ah, I think it was confusing to me because I didn't understand the meaning of "Add a projection for the remaining expressions". A result step may still require execution of addition expressions to produce the output. So the confusion was around seeing the `result` being found, but that potentially not being the actual output of the query. 

The type vars make sense now. As does the logic here. I think it's clear as is now.

crates/sparrow-execution/src/lib.rs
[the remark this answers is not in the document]
jordanrfrazier: I'd think it depends on how we manage state. It's possible for a source to be ahead of another source, up until the `merge` channel it produces to fills up and is blocked on sending new batches. In that case, the `merge` state may hold up to time `T+10` for source `A`, but only up to `T` for source `B`. In that case, we'd want source `A` to begin reading from `T+10`, and source `B` from `T`. 

I suspect (worry) it may be more complicated than just that, though.

crates/sparrow-batch/src/batch.rs
jordanrfrazier: Yeah, agreed on `Batch (structarray)` being passed everywhere. Not having to do "gimme this column" everywhere will hopefully simplify execution (and reduce code, as you say). 

I missed that the only place this `into-record_batch` is called is in the `WriteChannelPipeline`, which is only used for tests. So I think we're on the right track currently.

crates/sparrow-execution/src/lib.rs
jordanrfrazier: Ah, right. I'm still getting the grasp of the step to pipeline conversion. 

Can you add "step" to the error to help remind me? 

`Non-transform steps should be in (or, be executed) as separated pipelines`

bjchambers: Yep. Done.

crates/sparrow-execution/src/lib.rs
bjchambers: Actually, I'll probably defer. I think we should discuss whether it makes more sense to just pass the `ExecutionOptions` everywhere, vs. having this separate `ReadConfig`.

crates/sparrow-backend/src/logical_to_physical.rs
bjchambers: Already had it. https://github.com/kaskada-ai/kaskada/issues/787

crates/sparrow-execution/src/lib.rs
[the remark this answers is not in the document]
bjchambers: Hmm... interesting point. I think we should probably discuss separately -- this would be stateful *merge* not just stateful *aggregation*. I'm not sure how well it would work out, when we consider per-entity state and/or late data. But I haven't thought much about it.

crates/sparrow-execution/src/lib.rs
bjchambers: I think if the pipeline could access the tokio runtime handle, it could actually be responsible for launching the tokio task. This would give pipelines (in general) a nice ability to actually execute async tasks, which could be beneficial.

crates/sparrow-scheduler/src/worker_pool.rs
bjchambers: Heh. I spent a week or two trying different things to avoid the cyclic reference. The problem is that:

1. Pipelines need to know how to schedule their various partitions (so they need a "schedule-token")
2. When the worker receives a task (from the queue), it needs the pipeline (to get the code). Call this "execute-token".
3. It's really helpful if the "schedule-token" and "execute-token" are the same. For instance, we don't want to instances of the same partition running -- so the "combined token" can instead increment a count on schedule, and then check it after execution and say "since this was scheduled during execution, I should requeue it".

The cycle emerges from: Pipeline -> (ScheduleToken == ExecuteToken) -> Pipeline.

jordanrfrazier: Ah..yeah looking at the code more, that definitely looks tricky, to say the least.

crates/sparrow-execution/src/lib.rs
jordanrfrazier: Yeah `ReadPipeline` would be a `no-op`, would just be there for readability if we want that.

crates/sparrow-execution/src/lib.rs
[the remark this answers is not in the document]
jordanrfrazier: Yeah I think we'll have to dedicate some more time to it. But at the least, any stateful operation will need to save state. Doing _just_ stateful aggregation doesn't give us anything, does it?

crates/sparrow-execution/src/lib.rs
bjchambers: Filed https://github.com/kaskada-ai/kaskada/issues/814

crates/sparrow-backend/src/logical_to_physical.rs
```diff
@@ -118,11 +136,21 @@ impl LogicalToPhysical {
                 // of binary merges by choosing the order to perform the merges based on the number
                 // of rows.
                 let inputs: Vec<_> = must_merge.into_iter().collect();
+                let result_fields: arrow_schema::Fields = inputs
+                    .iter()
+                    .map(|input_step| {
+                        self.step_type(*input_step).map(|data_type| {
+                            Field::new(format!("step_{}", input_step), data_type, true)
```
jordanrfrazier, line 143: I worry about when we have to manually set the nullability. Any reason we think it could ever be `false`?

crates/sparrow-backend/src/logical_to_physical.rs
```diff
@@ -296,6 +375,24 @@ impl LogicalToPhysical {
...
+        sparrow_logical::Literal::Bool(_) => todo!(),
+        sparrow_logical::Literal::String(s) => Ok(ScalarValue::Utf8(Some(s.clone()))),
+        sparrow_logical::Literal::Int64(_) => todo!(),
+        sparrow_logical::Literal::UInt64(_) => todo!(),
+        sparrow_logical::Literal::Float64(_) => todo!(),
+        sparrow_logical::Literal::Timedelta {
+            seconds: _,
+            nanos: _,
+        } => todo!(),
+        sparrow_logical::Literal::Uuid(_) => todo!(),
```
jordanrfrazier, line 392: Leaving these to follow up in #787 ?

crates/sparrow-execution/src/lib.rs
```diff
@@ -8,5 +8,224 @@
...
+    ) -> error_stack::Result<Arc<dyn sparrow_scheduler::Pipeline>, Error> {
+        tracing::trace!(
+            "Creating transform pipeline: {}",
+            steps.iter().format_with(",", |step_id, f| {
+                let step = &plan.steps[*step_id];
+                f(&format_args!("{step}"))
+            })
+        );
+
+        // Determine the producer of the
```
jordanrfrazier, line 221: .

crates/sparrow-execution/src/source_tasks.rs
```diff
@@ -0,0 +1,101 @@
...
+    // /// with indices corresponding to the input in `read_inputs` to send the batch
+    // /// to.
+    // ///
+    // // TODO: We may want to encapsulate this differently / better. Could we use
+    // // `futures::sink(...)` to do this in some way? We may also want to allow
+    // // the polling of these sources to be weighted towards the inputs that need
+    // // a new batch to progress (eg., back-pressure).
+    // reads: Vec<BoxStream<'static, error_stack::Result<(Batch, usize), SourceError>>>,
+    // /// For each read operation, the `PipelineInputs` to send the batches to.
+    // read_inputs: Vec<InputHandles>,
```
jordanrfrazier, line 26: Leaving this for future improvements?

crates/sparrow-physical/src/step.rs
```diff
@@ -90,3 +100,42 @@ pub enum StepKind {
...
+    ///
+    /// - `project` is a transform because it computes new columns for each row in
+    ///   every batch
+    /// - `filter` is a transform because it removes rows from each batch, and omits
+    ///   empty batches
+    /// - `merge` is not a transform because it accepts multiple inputs
+    /// - `shift` may be a transform if it uses the time in the input batch to determine
+    ///   which rows to output or it may be an operation if it interacts with scheduling
+    ///   in a more sophisticated way.
+    pub fn is_transform(&self) -> bool {
```
jordanrfrazier, line 123: I like the examples here. But the `shift` one is confusing coupled with this boolean method. `may be a transform` implies we'd need context to determine, but if we just call `is_transform` we'd get `false`. A caller may unknowingly call this without reading the docs and assume it's always not. 

(Though, I thought `shift` was never a `transform`)

crates/sparrow-scheduler/src/worker.rs
```diff
@@ -35,49 +38,89 @@ pub trait Scheduler {
...
-        while let Some(task) = self.queue.pop() {
-            if task.do_work(&mut self)? {
-                // This means that the task was schedule while we were executing.
-                // As a result, we didn't add it to any queue yet, so we need to
-                // do so now.
-                self.queue.push_global(task);
+    pub(crate) fn work_loop(mut self, index: usize) -> error_stack::Result<(), Error> {
+        let thread_id = std::thread::current().id();
+        let _span = tracing::info_span!("Worker", ?thread_id, index).entered();
+        tracing::info!("Starting work loop on thread {thread_id:?}");
```
jordanrfrazier, line 96: General question: If we're within the span here, do we need to reiterate variables that are already apart of the span? Does having `thread_id` in this `info` duplicate that variable in the log line?

crates/sparrow-scheduler/src/worker_pool.rs
```diff
@@ -12,70 +11,36 @@ const DEFAULT_THREAD_COUNT: usize = 8;
...
-    handles: Vec<std::thread::JoinHandle<error_stack::Result<(), Error>>>,
+    workers: Vec<Worker>,
     /// A vector of the pipelines we created.
     pipelines: Vec<Arc<dyn Pipeline>>,
 }
 
-impl WorkerPool {
-    pub fn start(query_id: String) -> error_stack::Result<Self, Error> {
+impl WorkerPoolBuilder {
+    /// Create a worker pool.
```
jordanrfrazier, line 23: worker pool builder?

jordanrfrazier (approved):

crates/sparrow-backend/src/logical_to_physical.rs
```diff
@@ -50,7 +50,12 @@ pub(super) struct LogicalToPhysical {
 #[derive(Debug)]
 struct Reference {
     /// Reference to a specific step.
-    step_id: sparrow_physical::StepId,
+    ///
+    /// This should only be `None` if the expression is a literal. It should
+    /// not be `None` if the expression is a literal arising within a specific
+    /// step (domain) -- only for cases where the value originally has no
+    /// domain.
```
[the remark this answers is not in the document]
bjchambers: Exactly. Consider `a - a`. We could represent that as the literal `0` in the domain of `a`. That said, I *don't think* we'll be handling that in the `logical_to_physical` case. It seems like that would be better to handle in a later step.

crates/sparrow-backend/src/logical_to_physical.rs
bjchambers: Yeah. Either handle in #787 or as part of testing.

crates/sparrow-execution/src/source_tasks.rs
bjchambers: nope -- stale comment. deleted.

crates/sparrow-scheduler/src/worker.rs
bjchambers: We don't need to repeat them, although duplication depends on the tracing backend. Eg. in jaeger you may see the span as a section in the visualization, while the log line is a specific point. On the console, we can configure whether the span is included at all, etc.

Generally -- doesn't seem to *hurt* to have it repeated, although as you say, probably not necessary. I'll drop it.

crates/sparrow-physical/src/step.rs
bjchambers:
```
/// - `shift` depends on how we choose to implement it. If it is implemented as
///   a stateful transform that just buffers and emits as processing proceeds
///   through time, then it would be a transform. If we find ways to implement
///   it more efficiently by implementing the pipeline interface, then it may
///   not be a transform.
```

crates/sparrow-scheduler/src/worker.rs
```diff
@@ -87,19 +130,31 @@ impl Worker {
...
+        } else {
+            tracing::trace!("2 {task:?} already executing");
         }
+        self.idle_workers.wake_one();
     }
 
     fn schedule_yield(&mut self, task: TaskRef) {
         if task.schedule() {
-            self.queue.push_yield(task)
+            tracing::trace!("3 Added {task:?} to queue");
```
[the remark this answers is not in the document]
bjchambers: just easy ways. removing

crates/sparrow-backend/src/logical_to_physical.rs
jordanrfrazier: Hm no suppose not. The main problems I had with this was when I had to verify the type and it was getting mad that a nullable vs not field was different. In this case I think it's okay.

PR #808, opened 2023-10-11
ci: fix release drafting

bjchambers:


.github/workflows/pr_labeler.yml
```diff
@@ -29,6 +29,7 @@ jobs:
       - name: Label pull request
         uses: release-drafter/release-drafter@v5
         with:
+          config-name: release-drafter.yml # (default)
```
bjchambers, line 32: This is the default, but I thought it may be better to be explicit about that.

epinzur (approved):

PR #810, opened 2023-10-16
fix: release the GIL during sync iteration

bjchambers:


python/pytests/execution_test.py
```diff
@@ -34,14 +34,27 @@ async def test_iter_pandas(golden, source_int64) -> None:
...
+    print("In UDF")
+    return x + 1
+
+
+async def test_iter_udf(source_int64) -> None:
+    results = source_int64.col("m").pipe(add_one).run_iter("row", row_limit=2)
+    assert next(results)["result"] == 6
+    assert next(results)["result"] == 25
+
+
```
epinzur, line 57: I would have expected a non-async test.

epinzur: ah, maybe the test is async but the call to `run_iter` is non-async

epinzur (approved):

python/pytests/execution_test.py
bjchambers: I think all of our tests require `async` because under the hood some of the APIs use async.

PR #819, opened 2023-10-17
ref: split up the expression crate

bjchambers:
This moves the traits for defining an expression evaluator into `sparrow-interfaces` and the code for executing a sequence of expressions into `sparrow-expr-execution.`.

This allows flattening the `sparrow-expressions` crate, and only using it as a test dependency in a wide variety of places. It is only needed at execution time.

This is preparation for adding `sparrow-aggregations` to implement the stateful aggregations.


.vscode/settings.json
```diff
@@ -1,4 +1,7 @@
 {
     "rust-analyzer.cargo.features": "all",
     "editor.formatOnSave": true,
+    "rust-analyzer.linkedProjects": [
```
jordanrfrazier, line 4: I see: 
```
rust-analyzer.linkedProjects (default: [])

    Disable project auto-discovery in favor of explicitly specified set of projects
```
but I don't really know what that means. Can you explain?

crates/sparrow-expressions/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition.workspace = true
 license.workspace = true
 publish = false
 description = """
-Execution of expressions over record batches.
+Implementation of stateless expressions over batches.
```
jordanrfrazier, line 9: I forgot if we added machete to CI, but did you run it here? Figured maybe `sparrow-physical` wouldn't be needed anymore here

jordanrfrazier (approved):

crates/sparrow-expressions/Cargo.toml
bjchambers: We had added it, but due to the default working directory, it only affected Python. I may change the default back to that.

crates/sparrow-expr-execution/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+publish = false
+description = """
+Execution of expressions over record batches.
+"""
+
+[dependencies]
+arrow-array.workspace = true
```
bjchambers, line 13: Execution now has very minimal deps, which is nice.

PR #820, opened 2023-10-18
ref: use lexsort to verify batch order

jordanrfrazier:
Uses lexsort for ordering validations

### DIFF
--- a/crates/sparrow-batch/src/batch.rs
+++ b/crates/sparrow-batch/src/batch.rs
@@ -385,49 +385,53 @@ pub(crate) fn validate_bounds(
     key_hash: &ArrayRef,
     up_to_time: RowTime,
 ) -> error_stack::Result<(), Error> {
+    use arrow::compute::SortColumn;
+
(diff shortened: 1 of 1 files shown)


crates/sparrow-batch/src/batch.rs
```diff
@@ -385,49 +385,53 @@ pub(crate) fn validate_bounds(
...
+        None,
+    )
+    .into_report()
+    .change_context(Error::internal_msg("lexsort_to_indices".to_owned()))?;
+
+    let is_sorted = sort_indices
+        .values()
+        .iter()
+        .enumerate()
+        .all(|(i, x)| i == *x as usize);
```
jordanrfrazier, line 419: More efficient way to do this?

bjchambers: This seems like a reasonable way to do it to tme. Less worried about efficiency since this is a validation we do in debug builds.

bjchambers (approved):

PR #821, opened 2023-10-18
test: add in-memory batch unit tests

jordanrfrazier:
Adds a few unit tests for the in-mem batch.

### DIFF
--- a/crates/sparrow-sources/src/in_memory.rs
+++ b/crates/sparrow-sources/src/in_memory.rs
@@ -3,7 +3,8 @@ use std::sync::{Arc, RwLock};
 use arrow_array::RecordBatch;
 use arrow_schema::{DataType, SchemaRef};
 use error_stack::{IntoReportCompat, ResultExt};
-use futures::{Stream, StreamExt, TryStreamExt};
+use futures::stream::BoxStream;
+use futures::{StreamExt, TryStreamExt};
 
 use sparrow_batch::Batch;
 use sparrow_interfaces::source::{Source, SourceError};
@@ -148,6 +149,8 @@ impl InMemoryBatches {
     /// Add a batch, merging it into the in-memory version.
     ///
     /// Publishes the new batch to the subscribers.
+    ///
+    /// Note: This assumes the batch has been prepared, and will likely panic if not.
     pub async fn add_batch(&self, batch: RecordBatch) -> error_stack::Result<(), SourceError> {
         if batch.num_rows() == 0 {
             return Ok(());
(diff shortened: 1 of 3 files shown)


crates/sparrow-sources/src/in_memory.rs
```diff
@@ -174,9 +177,7 @@ impl InMemoryBatches {
     ///
     /// The first batch will be the in-memory merged batch, and batches will be
     /// added as they arrive.
-    pub fn subscribe(
-        &self,
-    ) -> impl Stream<Item = error_stack::Result<RecordBatch, SourceError>> + 'static {
+    pub fn subscribe(&self) -> BoxStream<'static, error_stack::Result<RecordBatch, SourceError>> {
```
bjchambers, line 180: This is probably fine, although I suspect we're boxing it possibly again later (which we may want to check for and remove, if we are). Is there a reason it needed to be boxed?

crates/sparrow-sources/src/in_memory.rs
```diff
@@ -222,3 +224,180 @@ impl InMemoryBatches {
...
+        assert_eq!(batch1, b1_s2);
+
+        let b2_s1 = s1.next().await.unwrap().unwrap();
+        assert_eq!(batch2, b2_s1);
+        let b2_s2 = s2.next().await.unwrap().unwrap();
+        assert_eq!(batch2, b2_s2);
+    }
+
+    #[tokio::test]
+    async fn test_late_subscription_receives_merged_batch() {
```
bjchambers, line 338: Should there be one for the non-subscription case (eg., just a read)? What about the case where we don't have retention?

crates/sparrow-sources/src/in_memory.rs
```diff
@@ -222,3 +224,180 @@ impl InMemoryBatches {
...
+        // Verify later subscription gets the full merged batch
+
+        let time: ArrayRef = Arc::new(TimestampNanosecondArray::from(vec![0, 1, 2]));
+        let subsort: ArrayRef = Arc::new(UInt64Array::from(vec![0, 1, 2]));
+        let key: ArrayRef = Arc::new(UInt64Array::from(vec![0, 0, 0]));
+        let a: ArrayRef = Arc::new(Int32Array::from(vec![0, 1, 2]));
+        let b: ArrayRef = Arc::new(UInt64Array::from(vec![None, Some(1), Some(8)]));
+        let schema = Schema::new(vec![
```
bjchambers, line 346: May be helpful to create some helper methods for things like the schema / creating batches / etc.

bjchambers (approved):

crates/sparrow-sources/src/in_memory.rs
jordanrfrazier: ...I guess not. I would have sworn that I was unable to call `next()` on it because it wasn't `Unpin`, but maybe I hallucinated that.

PR #823, opened 2023-10-18
tests: Parameterized pytests for math operations

bjchambers:
This replaces some math tests within Kaskada with pytests coverage.


python/pytests/math_test.py
```diff
@@ -1,35 +1,120 @@
...
+            {"time": "1996-12-19T16:40:04", "user": "A", "m": 12.7, "n": 8.7 },
+        ],
+        time_column="time",
+        key_column="user",
     )
+
+    return SourceCols(m=source.col("m"), n=source.col("n"), datatype=datatype)
+
+async def test_add(source, golden) -> None:
+    golden.jsonl(source.m + source.n)
```
bjchambers, line 43: Maybe better to do:

```
golden.jsonl(kd.record({"m": source.m, "n": source.n, "result": source.m + source.n}))
```

so that the inputs show up. Thoughts? Or even have m and n in the source and then just do `source.extend({"result": source.m + source.n})`.

jordanrfrazier (approved):
Awesome. This is really nice

epinzur (approved):

PR #827, opened 2023-10-21
feat: add gatherer for sparrow-merge

jordanrfrazier:
Adds the gatherer to the sparrow-merge crate. The gatherer is used to progress a watermark for several inputs and emit batches up to that watermark, rather than doing any actual merge stuff, but we'll want this eventually.


crates/sparrow-batch/src/batch.rs
```diff
@@ -94,6 +94,52 @@ impl Batch {
         })
     }
 
+    #[cfg(any(test, feature = "testing"))]
+    pub fn record_batch(self) -> Option<RecordBatch> {
```
bjchambers, line 98: How does this differ from `into_record_batch`? Can it use that?

crates/sparrow-batch/src/batch.rs
```diff
@@ -179,42 +225,51 @@ impl Batch {
 
     /// Split off the rows less or equal to the given time.
```
bjchambers, line 226: comment is no longer correct.

crates/sparrow-batch/src/batch.rs
```diff
@@ -550,16 +605,29 @@ impl BatchInfo {
...
```
bjchambers, line 606: .

crates/sparrow-merge/src/gather.rs
```diff
@@ -0,0 +1,339 @@
+mod gathered_batches;
+
+use crate::gather::gathered_batches::GatheredBatches;
```
bjchambers, line 3: should this be `pub use gathered_batches::GatheredBatches`? It seems like things return `GatheredBatches`, so it is part of the API.

crates/sparrow-merge/src/gather/gathered_batches.rs
```diff
@@ -0,0 +1,22 @@
+use sparrow_batch::{Batch, RowTime};
+
+/// The result of a gather operation
+#[derive(Debug, PartialEq)]
+pub struct GatheredBatches {
```
bjchambers, line 5: Does this really merit a separate file? It seems like it's "the result of gather" so it could go in `gather.rs` near the top as part of the API?

bjchambers (approved):

crates/sparrow-merge/src/gather/gathered_batches.rs
jordanrfrazier: Eh, maybe. I made this in the assumption that I'd move some more stuff into `gather`, but since I haven't I'll move this over and we can split it in the future if we need to.

crates/sparrow-batch/src/batch.rs
jordanrfrazier: `into_record_batch` requires you to get/know the schema. This is just a helper that looks at the `data` data type and creates the schema for you inline. 

Though, did add comment that this is a "hack" and we want to standardize on Batch.

PR #828, opened 2023-10-24
feat: partitioned merge pipeline

jordanrfrazier:
Adds the `merge` pipeline (some functionality, at least). 

Blocker:
* Merge pipeline tests

Deferred:
* Grouping / Interpolation - Spread only functions in the unlatched capacity.


crates/sparrow-merge/src/merge_pipeline.rs
```diff
@@ -0,0 +1,356 @@
...
+};
+
+use crate::gather::Gatherer;
+use crate::merge::BinaryMergeInput;
+use crate::spread::Spread;
+
+/// Runs a merge operation.
+pub struct MergePipeline {
+    /// The state for each partition.
+    partitions: Partitioned<Mutex<MergePartition>>,
```
bjchambers, line 22: You have two layers of mutexes which are unnecessary:

`Partitioned<Mutex<MergePartition>>` locks the merge partition, but you can also lock the `MergePartition::inputs` and `MergePartition::gatherer`. You probably want to think about how to split things up.

As a general goal: ideally the inputs would go through some kind of channel without a lock.

bjchambers: So you end up with:

```
inputs: Partitioned<Channel<...>>,
state: Partitioned<Mutex<MergePartition>>,
```

and then you don't have any mutexes in the `MergePartition` -- basically, you only get the lock in `do_work` so there is no contention for the states.

crates/sparrow-merge/src/merge_pipeline.rs
```diff
@@ -0,0 +1,356 @@
...
+    /// Inputs (for either side) for this partition.
+    ///
+    /// the `usize` should be either 0 (left), or 1 (right).
+    inputs: Mutex<VecDeque<(Batch, usize)>>,
+    /// Implementation of `spread` to use for this input
+    #[allow(unused)]
+    spread_l: Spread,
+    /// Implementation of `spread` to use for this input
+    #[allow(unused)]
+    spread_r: Spread,
```
bjchambers, line 60: Thought: Could we have a `HeterogenousMerge` struct that manages the spreads? We could do something like:

```
let merger = HeterogeneousMerge::new(vec![schema1, schema2, schema3, ...]);
let merged = merger.merge(gathered_batches)?;
```

bjchambers: (separately, not sure if the gatherer should be baked into that heterogeneous merge or not)

crates/sparrow-merge/src/gather.rs
```diff
@@ -77,6 +81,11 @@ impl Gatherer {
         self.active.is_empty() || self.active.peek().unwrap().up_to_time > self.emitted_up_to_time
     }
 
+    /// Returns true if all inputs have been closed.
+    pub fn all_closed(&self) -> bool {
+        self.pending.iter().all(|p| p.closed)
```
bjchambers, line 86: This makes sense for now, but maybe worth a quick comment that if this ever shows up as taking time (such as if there are lots of pending items), it may be better to count the number of "pending" items separately, so this can just be `self.pending_count == 0` or something.

crates/sparrow-merge/src/merge/heterogeneous_merge.rs
```diff
@@ -0,0 +1,276 @@
...
+    spread_l: Spread,
+    spread_r: Spread,
+
+    /// Gathers batches from both sides and produces [GatheredBatches]
+    /// up to a valid watermark.
+    gatherer: Gatherer,
+}
+
+#[derive(derive_more::Display, Debug)]
+pub enum Error {
```
bjchambers, line 26: For now, if we think this will always be just one error, we could simplify to:

```
#[display(fmt = "internal error merging: {}", _0)]
pub struct Error(&'static str);
```

crates/sparrow-merge/src/merge_pipeline.rs
```diff
@@ -11,29 +11,32 @@ use sparrow_physical::StepId;
...
 pub struct MergePipeline {
+    /// The senders for each input and partition.
+    ///
+    /// Batches are first placed into the channels, which allows us to use
+    /// [Gatherer] to choose which input needs to be merged next.
+    ///
+    /// This vec should contains one element for each input.
+    txs: Vec<Partitioned<tokio::sync::mpsc::UnboundedSender<Batch>>>,
+    /// The matching receivers for each sender.
+    rxs: Vec<Partitioned<tokio::sync::mpsc::UnboundedReceiver<Batch>>>,
```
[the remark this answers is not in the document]
bjchambers: I guess you could even put the `txs` in there if you wanted?

crates/sparrow-merge/src/spread.rs
jordanrfrazier, line 1: I only brought over the generic fallback spread.

bjchambers: Would be good to measure generic vs. primitive spread (those are likely the least and most optimized). I could be convinced we only need the generic if the performance seems to support that, and then we could get rid of a lot of the generics / traits / etc. Otherwise, would be good to know the performance is such that we probably want to move the rest over at some point.

crates/sparrow-merge/src/merge_pipeline.rs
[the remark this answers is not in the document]
jordanrfrazier: I think I will yes, given how we `initialize` the partitions after creating the pipeline.

crates/sparrow-merge/src/merge/heterogeneous_merge.rs
```diff
@@ -0,0 +1,512 @@
...
+                    self.handle_single_merge(1, right, up_to_time)
+                }
+                (None, None) => {
+                    // This is in unexpected state -- if we call merge, we should be getting at least
+                    // one batch, even if it's empty.
+                    error_stack::bail!(Error::Internal("at least one batch"))
+                }
+            }
+        } else {
+            error_stack::bail!(Error::Internal("expected batch -- "))
```
jordanrfrazier, line 105: fix

crates/sparrow-merge/src/merge/binary_merge.rs
jordanrfrazier, line 1: This file is unchanged from the original -- but I didn't want to update all the references the old code path uses for this PR.

crates/sparrow-batch/src/batch.rs
```diff
@@ -231,47 +261,58 @@ impl Batch {
...
+                // all rows should be taken
+                return Some(Batch {
+                    data: self.data.take(),
+                    // Even though we took all rows, we need to pick the
+                    // minimum up_to_time. If
+                    // 1) self.up_to_time < time_exclusive: the input may have more rows
+                    // between self.up_to_time and time_exclusive.
+                    // 2) self.up_to_time > time_exclusive: We requested a split at
+                    // time_exclusive, so we can't return a batch with a greater time.
+                    up_to_time: self.up_to_time.min(time_exclusive),
```
jordanrfrazier, line 282: This could go either way, I think. Technically, this definition is true. If we ask for a split at `time_exclusive` then we shouldn't produce an `up_to_time` greater than that. 

But, we could also assert that `time_exclusive` is always `<=` the `self.up_to_time`. The `Gatherer` should never ask for a split time greater than the `up_to_time` of the minimum batch.

crates/sparrow-batch/src/batch.rs
```diff
@@ -231,47 +261,58 @@ impl Batch {
...
-            // If there's no data after we split at `time_exclusive`, return None
-            None
+            // No data, but we may have to produce an empty batch with
+            // a new up_to_time. Use the minimum, as
+            // 1) self.up_to_time < time_exclusive: the input may have more rows
+            // between self.up_to_time and time_exclusive.
+            // 2) self.up_to_time > time_exclusive: We requested a split at
+            // time_exclusive, so we can't return a batch with a greater time.
+            let new_up_to = time_exclusive.min(self.up_to_time);
+            Some(Batch::new_empty(new_up_to))
```
jordanrfrazier, line 315: It would be nice if we could split this logic out of `Batch`, and track watermarks separately

crates/sparrow-merge/src/merge_pipeline.rs
```diff
@@ -0,0 +1,316 @@
...
+    task: TaskRef,
+    /// The senders for each input.
+    ///
+    /// Batches are first placed into the channels, which allows us to use
+    /// [Gatherer] to choose which input needs to be merged next.
+    ///
+    /// This vec should contains one element for each input.
+    txs: Vec<tokio::sync::mpsc::UnboundedSender<Batch>>,
+    // Keeps track of the entities seen by this partition
+    // TODO: key_hash_index: KeyHashIndex,
```
jordanrfrazier, line 57: Not a blocker -- related to `Grouping` in `spread`, which we'll do as a follow-up

crates/sparrow-batch/src/batch.rs
```diff
@@ -77,6 +78,35 @@ impl Batch {
         })
     }
 
+    /// Updates the up_to_time.
+    pub fn with_up_to_time(self, up_to_time: RowTime) -> Self {
```
bjchambers, line 82: Any details on how this should be used? Eg., should we have an assertion that this only increases the `up_to_time`? Or should we limit this to `up_to_time >= `self.data.max_time` or anythnig like that in assertions?

crates/sparrow-batch/src/batch.rs
```diff
@@ -77,6 +78,35 @@ impl Batch {
...
+            Self {
+                data: Some(BatchInfo {
+                    data,
+                    time: info.time,
+                    subsort: info.subsort,
+                    key_hash: info.key_hash,
+                    min_present_time: info.min_present_time,
+                    max_present_time: info.max_present_time,
+                }),
+                up_to_time: self.up_to_time,
```
bjchambers, line 103: Similarly here -- I wonder if we should check the affect this has on the relationship between times int he data and the up to time?

crates/sparrow-batch/src/batch.rs
```diff
@@ -77,6 +78,35 @@ impl Batch {
...
+            }
+        } else {
+            panic!("Cannot replace data of empty batch");
```
bjchambers, line 106: Why does this panic? Couldn't this do `self.data = Some(BatchInfo { data ... })`? Or is this because it reuses the original `time`, `subsort`, `key_hash`, etc.?

I suspect the latter, which suggests:

1. We should document this changes the data, but does not update the time, subsort, or key hash.
2. Add an assertion that the `data.len()` is the same (since otherwise we don't satisfy the requirement that the data and time/subsort/key_hash have the same number of rows).

crates/sparrow-batch/src/batch.rs
bjchambers: I'm not sure I understand if there is a question here, but:

1. It seems like if we can assert / narrow the cases this supports, that may be good for ensuring it is being used as we intended.
2.

crates/sparrow-batch/src/batch.rs
bjchambers: Yeah. I know we've discussed the fact that we should separate `Batch` (and how data flows) from the watermark / `up_to_time`. Do we have a ticket we should reference, or should we create one?

crates/sparrow-merge/src/gather.rs
```diff
@@ -38,6 +42,15 @@ impl Gatherer {
         self.active.peek().map(|min| min.index)
     }
 
+    /// Returns a boolean indicating whether the gatherer can produce a batch.
+    pub fn can_produce(&self) -> bool {
```
bjchambers, line 46: Could shorten to:

`self.active.peek().is_some_and(|top| top.up_to_time > self.emitted_up_to_time)`

crates/sparrow-merge/src/gather.rs
```diff
@@ -46,6 +59,10 @@ impl Gatherer {
         self.pending[index].add_batch(batch);
 
         if let Some(mut entry) = self.active.pop() {
+            // Expect `blocking_input` to be called first, ensuring we're working
+            // on the current blocking input
+            assert_eq!(entry.index, index);
```
bjchambers, line 64: Maybe add a message to the assertion?

`assert_e!(entry.index, index, "Expected next batch from blocking input {} but was {index}", entry.index)`

crates/sparrow-merge/src/merge_pipeline.rs
```diff
@@ -0,0 +1,340 @@
...
+}
+
+struct Input {
+    /// The input id for this input
+    #[allow(unused)]
+    id: StepId,
+    /// The data type for this input
+    datatype: DataType,
+    /// Whether this side is closed
+    is_closed: Mutex<bool>,
```
bjchambers, line 41: I think `AtomicBool` is often preferred to `Mutex<bool>` (https://users.rust-lang.org/t/is-it-okay-to-use-mutex-bool/90742) but this is probably OK. The main risk is that we need to be careful of lock order if we ever lock both this and some other mutex, to always do so in the same order.

crates/sparrow-merge/src/merge_pipeline.rs
```diff
@@ -0,0 +1,340 @@
...
+use crate::merge::HeterogeneousMerge;
+
+/// Runs a merge operation.
+pub struct MergePipeline {
+    /// The state for each partition.
+    partitions: Partitioned<MergePartition>,
+    /// Input handles for the consuming (receiving) computations.
+    consumers: InputHandles,
+    /// The left input.
+    left: Input,
```
bjchambers, line 21: Should this be `Vec<Input>` to handle the multiple input case? Or is that in a follow-up?

crates/sparrow-merge/src/merge_pipeline.rs
```diff
@@ -0,0 +1,340 @@
...
+
+    fn is_closed(&self) -> bool {
+        self.is_closed.load(Ordering::Acquire)
+    }
+}
+
+/// Handles the merge logic.
+///
+/// Separated from the [MergePartition] to avoid multiple locks.
+struct MergePartitionHandler {
```
bjchambers, line 78: Naming: Maybe `MergePartitionState` instead? Eg., this is the mutable state for the merge partition, while the `MergePartition` itself is the immutable staet?

crates/sparrow-merge/src/merge_pipeline.rs
```diff
@@ -0,0 +1,340 @@
...
+    ) -> error_stack::Result<(), PipelineError> {
+        error_stack::ensure!(
+            input == 0 || input == 1,
+            PipelineError::InvalidInput {
+                input,
+                input_len: 2
+            }
+        );
+
+        {
```
bjchambers, line 224: Feels like even if you required this only has 2 elements, it would be cleaner if we used a vector here. Then you could do `*self.inputs[input].is_closed.lock() = true`. Maybe that's a good first step -- move away from `left/right` and these conditionals, towards a 2 element vector (and change assertions to use `input < self.inputs.len()`, etc. Then we'll be closer to making this support multiple inputs, without needing to do all of it right away.

crates/sparrow-merge/src/merge_pipeline.rs
```diff
@@ -0,0 +1,340 @@
...
+                    // not have disconnected from the channel yet. This is okay -- if the partition
+                    // is verified closed, we can continue to close our consumers here.
+                    return self.complete_partition(input_partition, partition, scheduler);
+                }
+                Err(TryRecvError::Disconnected) => {
+                    return self.complete_partition(input_partition, partition, scheduler)
+                }
+            }
+        } else {
+            // Though all inputs are closed, the merger may have batches remaining to flush.
```
bjchambers, line 311: Is this all closed? This just means "nothing blocking the gatherer", right?

bjchambers (approved):

crates/sparrow-batch/src/batch.rs
jordanrfrazier: It was between using the minimum of `self.up_to_time` and the given `time_exclusive` or just using `self.up_to_time` and asserting that it's not being split at a time greater than `self.up_to_time`. 

This method could be used outside of the gatherer, which is why I chose to use the minimum, as it's more in line with the actual definition of `split_up_to`.

crates/sparrow-batch/src/batch.rs
jordanrfrazier: Added reference to issue

crates/sparrow-batch/src/batch.rs
jordanrfrazier: Added assertion that this increases the existing up_to_time. Probably _could_ allow for the latter, but it's not used in that manner, and I don't think we need to support that.

crates/sparrow-batch/src/batch.rs
bjchambers: I'm fine with either. If we rely on the property, it makes sense to add an assertion. I could also see either being fine. Perhaps a way to think about is -- how would this work if the watermark was outside of the batch. Specifically, once we do #836, how would our preferences change?

jordanrfrazier: I would think it's basically the same implementation -- if we recognize we are going to take the entire `batch`, we have to decide whether to take the minimum of the `Watermarked` batch's `up_to_time`, or the time we split at. 

The benefits of using the `WatermarkedBatch` don't help much here, as the structure is the same in places that need to handle the watermark. (e.g. `Optional<BatchInfo>`. vs. `Optional<Batch>`, and a watermark).

crates/sparrow-batch/src/batch.rs
jordanrfrazier: Use `BatchInfo::new` for validations

crates/sparrow-batch/src/batch.rs
jordanrfrazier: Yes, the latter. Updating docs accordingly, and using `BatchInfo::new()` for validations

crates/sparrow-batch/src/batch.rs
bjchambers: Maybe. On the other hand:

1. The handling of `up_to_time` would land in the `MergePipeline` and/or `Gatherer`
2. The `Batch::split_up_to` would be exclusively focusing on splitting the data in the batch. It wouldn't have any logic on how to compute the `up_to_time`.

So in that world:

1. The `gatherer` would determine the `up_to_time` / `watermark` of the output.
2. The `Batch::split_up_to` would split the data at the request of the gatherer.

I *think* that would lead to: `Batch::split_up_to` should always split the batch at the requested time, and the gatherer would pick a properly watermark / time.

I guess it doesn't speak to what to do for the `split_up_to` for now.

crates/sparrow-merge/src/merge_pipeline.rs
jordanrfrazier: Follow up, added task in backlog

crates/sparrow-merge/src/merge_pipeline.rs
jordanrfrazier: Probably so -- but I'm not sure yet where the "do n-way merge" logic will live. Will it be handled in `heterogenous merge`? Or just in the pipeline here, calling merge on specific batches each time? 

hetergeneous merge is written to handle `left` and `right` atm, so I may leave this to avoid complicating that code right now

crates/sparrow-merge/src/merge_pipeline.rs
jordanrfrazier: I suppose both yeah..we technically know that `close` has been called on all inputs, and we want to assert that that is the case, because we're completing our task here, but the actual `if` is just checking whether the gatherer has any blocking input. I can update language

crates/sparrow-batch/src/batch.rs
jordanrfrazier: Hm, fair. I was imagining that the `WatermarkedBatch` would have a `split_up_to` method that was functionally the same as the current `Batch`'s, since it's likely we'll have mutiple places we'll need to split data + watermark (merge, shifts)

PR #832, opened 2023-10-26
test: Timeout with an async iteration

bjchambers:


python/pytests/execution_test.py
```diff
@@ -139,3 +140,33 @@ async def test_history(golden, source_int64) -> None:
...
+
+    # Await the first batch.
+    golden.jsonl(await execution.__anext__())
+
+    # Add data and await the second batch.
+    await source_int64.add_string(data2)
+    golden.jsonl(await execution.__anext__())
+
+    with pytest.raises(TimeoutError):
+        async with asyncio.timeout(1):
```
bjchambers, line 167: I don't know if we want to check this in without marking it as a slow test. Specifically, it will wait at least 1 second for the timeout to occur. This can be increased further (10s) to verify that it isn't just "getting lucky" in terms of hitting an empty batch or somethnig like that.

Issue #825, opened 2023-10-19
Quarto FollowUp


epinzur:

epinzur:
- [ ] [Consider improving the CI flow](https://github.com/kaskada-ai/kaskada/pull/799#discussion_r1355442808)
- [ ] [Add an _extensions README.md](https://github.com/kaskada-ai/kaskada/pull/799#discussion_r1355445376)
- [ ] [Move Slack-Join URL to a variable](https://github.com/kaskada-ai/kaskada/pull/799#discussion_r1355449489)
- [ ] [Attempt to move index text to separate files](https://github.com/kaskada-ai/kaskada/pull/799#discussion_r1355463760)
- [ ] [Add a _scripts README.md](https://github.com/kaskada-ai/kaskada/pull/799#discussion_r1355466674)
- [ ] [Re-add comments to the bluesky example](https://github.com/kaskada-ai/kaskada/pull/799#discussion_r1355479326)
- [ ] [Convert the reddit example to qmd](https://github.com/kaskada-ai/kaskada/pull/799#discussion_r1355480732)
- [ ] [Adjust spacing of docstrings for aliases](https://github.com/kaskada-ai/kaskada/pull/799#discussion_r1355491500)
- [ ] [Create Issue/PR to on include-code-files extension to support start & end tags](https://github.com/kaskada-ai/kaskada/pull/799/files#r1355478006) 
- [ ] [Add a comment to the commented-out fields](https://github.com/kaskada-ai/kaskada/pull/799/files#r1355472850)
- [ ] [Refactor the card template to not require design-styles.css](https://github.com/kaskada-ai/kaskada/pull/799/files#r1355470921)
- [ ] [Move announcement text to a variable](https://github.com/kaskada-ai/kaskada/pull/799/files#r1355468383)

PR #833, opened 2023-10-27
updated docs for 2 examples

epinzur:
* time_centric
* bluesky

related to #825


python/docs/examples/time_centric.qmd
```diff
@@ -0,0 +1,209 @@
+---
```
bjchambers, line 1: So we can't use notebooks anymore? I guess that's OK for our current uses (just making docs), but it was nice for examples to have a notebook people could copy. Either way, not a blocker.

bjchambers (approved):

python/docs/examples/time_centric.qmd
epinzur: you can use notebooks, but you should save the output in the notebook to the repo.

If you want it to generate dynamically on each docs run, then use markdown.

PR #835, opened 2023-10-27
added READMEs and simplified CI

epinzur:
related to #825

### DIFF
--- a/.github/workflows/ci_python.yml
+++ b/.github/workflows/ci_python.yml
@@ -232,7 +232,6 @@ jobs:
           poetry env use 3.11
           source $(poetry env info --path)/bin/activate
           poetry install --with=docs
-          pip install ${WHEEL} --force-reinstall
           cd docs
           python _scripts/lint_reference.py
           deactivate
@@ -242,8 +241,6 @@ jobs:
         run: |
           poetry env use 3.11
           source $(poetry env info --path)/bin/activate
-          poetry install --with=docs
-          pip install ${WHEEL} --force-reinstall
           cd docs
           python _scripts/gen_reference.py
           python -m quartodoc interlinks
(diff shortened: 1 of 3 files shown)


.github/workflows/ci_python.yml
```diff
@@ -242,8 +241,6 @@ jobs:
         run: |
           poetry env use 3.11
           source $(poetry env info --path)/bin/activate
-          poetry install --with=docs
-          pip install ${WHEEL} --force-reinstall
```
epinzur, line 246: these are repeat actions from previous steps

python/docs/_scripts/README.md
```diff
@@ -0,0 +1,36 @@
...
+Start here if you want to change how & where we output individual markdown files
+for our API docs.
+
+* `builder.py` iterates over a passed config file and then uses `renderer.py` and `summarizer.py`
+to output markdown for the API docs. It based off the [default quartodoc builder](https://github.com/machow/quartodoc/blob/v0.6.3/quartodoc/autosummary.py#L367),
+but is heavily modified.
+* `gen_reference.py` sets up logging, and calls `builder.py` with the `../_reference.yml` config file
+
+Note that we output files in a VERY different way than the default quartodoc executable behaves.
+For instance, we rely on the path of each file to build the side-bar navigation instead of
```
bjchambers, line 28: Is this something we should try to change, or are we generally happy with how we have this / believe it won't be strangely broken by future quarto changes?

bjchambers (approved):

python/docs/_scripts/README.md
epinzur: this is quartodoc and not quarto. I'm pretty sure this won't break on future updates to quartodoc.

basically I'm making quartodoc generate output that is closer to how quarto wants things than quartodoc itself does.

PR #838, opened 2023-10-31
feat: add destination/writer trait 

jordanrfrazier:
And implements them for tokio channel(s).


crates/sparrow-plan-execution/src/write_pipeline.rs
```diff
@@ -0,0 +1,216 @@
...
+        tracing::trace!("Closing input for writer {}", partition.task);
+        error_stack::ensure!(
+            !partition.is_closed(),
+            PipelineError::InputClosed {
+                input,
+                input_partition
+            }
+        );
+
+        // We may be currently executing a `do_work` loop... TODO: FRAZ
```
jordanrfrazier, line 156: I still don't quite understand this logic

Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [workspace]
 members = ["crates/*"]
+resolver = "2"
```
jordanrfrazier, line 3: I think this is the new and improved resolver -- https://doc.rust-lang.org/cargo/reference/resolver.html#feature-resolver-version-2

crates/sparrow-interfaces/src/destination.rs
```diff
@@ -0,0 +1,62 @@
+use sparrow_batch::Batch;
+use std::fmt::Debug;
+
+use crate::types::Partition;
+
+/// Trait implemented by destinations.
+pub trait Destination: Send + Sync + Debug {
+    /// Creates a new writer to this destination.
```
bjchambers, line 8: For the interfaces, let's start with some more details on arguments as well:

```
/// Creates a new writer for this destination.
///
/// Arguments:
/// - partition: The partition this writer will write to.
```

crates/sparrow-interfaces/src/destination.rs
```diff
@@ -0,0 +1,62 @@
...
+    fn new_writer(
+        &self,
+        partition: Partition,
+    ) -> error_stack::Result<Box<dyn Writer>, DestinationError>;
+}
+
+pub trait Writer: Send + Sync + Debug {
```
bjchambers, line 15: Rustdoc:

```
/// Writer for a specific partition of a destination.
```

crates/sparrow-plan-execution/src/write_pipeline.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    fn pop_input(&self) -> Option<Batch> {
+        self.inputs.lock().pop_front()
+    }
+}
+
+#[derive(Debug)]
+enum WriterState {
+    NotStarted,
+    Writing(Mutex<Box<dyn Writer>>),
```
bjchambers, line 53: I think you can eliminate the box since the `Mutex` already implies it's on the heap, which eliminates an indirection:

```Mutex<dyn Writer>```

**Review by bjchambers — APPROVED** [review]

File: crates/sparrow-plan-execution/src/write_pipeline.rs (hunk 5 above)
Comment (jordanrfrazier) [reply crates/sparrow-plan-execution/src/write_pipeline.rs:53]: Hm, I thought that a `dyn` needed to be boxed because the implementations are not statically sized:
```
the size for values of type `(dyn Writer + 'static)` cannot be known at compilation time
within `parking_lot::lock_api::Mutex<parking_lot::RawMutex, (dyn Writer + 'static)>`, the trait `Sized` is not implemented for `(dyn Writer + 'static)`
no field of an enum variant may have a dynamically sized type
change the field's type to have a statically known size
```

Is there anything I'm not understanding there?

Comment (bjchambers) [reply crates/sparrow-plan-execution/src/write_pipeline.rs:53]: Hmm. Maybe I missed it. Things need to be on the heap (boxed, arc, etc.) in order to be `dyn`. I was thinking that `Mutex` would support that, and thought I had tested that in Rust Playground, but apparently my test wasn't deep enough. It looks like we do need it in the box.

File: crates/sparrow-plan-execution/src/write_pipeline.rs (hunk 1 above)
Comment (jordanrfrazier) [reply crates/sparrow-plan-execution/src/write_pipeline.rs:156]: Ah okay, no that makes sense. Thanks

File: crates/sparrow-plan-execution/src/write_pipeline.rs [hunk 6]
```diff
@@ -0,0 +1,214 @@
...
+                WriterState::NotStarted => {
+                    error_stack::bail!(PipelineError::illegal_state("writer not initialized"))
+                }
+                WriterState::Complete => {
+                    error_stack::bail!(PipelineError::illegal_state("writer is complete"));
+                }
+            }
+        }
+
+        // Note: We don't re-schedule the write if there is input or it's
```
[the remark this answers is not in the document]
jordanrfrazier: Rescheduled task if queue isn't empty, and added issue to remind us to investigate the race conditions here. https://github.com/kaskada-ai/kaskada/issues/842

PR #844, opened 2023-11-01
ref: fix circular dependencies

jordanrfrazier:
Moves the sparrow-plan-execution test into sparrow-session to avoid circular dependencies.


crates/sparrow-session/Cargo.toml
```diff
@@ -21,13 +21,15 @@ smallvec.workspace = true
 sparrow-api = { path = "../sparrow-api" }
 sparrow-backend = { path = "../sparrow-backend" }
 sparrow-compiler = { path = "../sparrow-compiler" }
+sparrow-expressions = { path = "../sparrow-expressions" }
```
bjchambers, line 24: Could these be dev-dependencies? Are they only used in tests?

crates/sparrow-session/src/lib.rs
```diff
@@ -13,6 +13,7 @@ mod expr;
 pub mod partitioned;
 mod session;
 mod table;
+mod tests;
```
bjchambers, line 16: I'd put `#[cfg(test)]` on the `mod tests`, then you don't need it on each individual test.

crates/sparrow-session/src/tests.rs
```diff
@@ -0,0 +1,2 @@
+//! Partitioned execution tests
+mod partitioned;
```
bjchambers, line 2: Probably unnecessary levels of nesting -- I'd just put the tests in `tests.rs` for now -- we're testing "tat sessions work" or "`tests/execution.rs` -- we're testing that "sessions can execute expressions".

bjchambers: Alternatively, move this to `src/partitioned/tests.rs` rather than `src/tests.rs`.

crates/sparrow-session/src/tests/partitioned/basic_tests.rs
```diff
@@ -1,17 +1,15 @@
...
-use sparrow_session::partitioned::Session;
 use std::sync::Arc;
 
 use arrow_schema::{DataType, Field, Schema};
 
-fn query(
-    session: &Session,
-    source: ExprRef,
-) -> error_stack::Result<ExprRef, sparrow_session::Error> {
+#[cfg(test)]
```
bjchambers, line 11: No longer need this, since the module is only included when tests are enabled.

bjchambers (approved):

Issue #841, opened 2023-10-31
Journalling of input batches


bjchambers:
- [ ] Write batches to a journal
- [ ] Recover batches from the journal
- [ ] Checkpoint journal segments to a "prepared" file
- [ ] Prepare files directly to journal checkpoints
- [ ] Sources for reading from the input journal / checkpoints

PR #845, opened 2023-11-01
feat: Initial journalling implementation

bjchambers:
This is part of #841.

- Persistence of batches using Arrow IPC format.
- Journalling and recovery of batches via append-only `okaywal` crate.
- Concatenating batches to produce checkpoints (also in IPC format)
- Recovering from checkpoints
- Concatenating journal checkpoints with in-memory data for execution.

Next steps
- Allow adding checkpointed batches directly (eg., for import from Parquet).
- Allow subscribing to newly added batches (eg., possibly replacing existing in-memory implementation).

Future work:
- Handle late data by merging with overlapping in-memory and possibly checkpointed batches.
- Checkpoint to object store / read checkpoints from object stores. May lead to moving checkpoint storage to Parquet, even while using IPC for the write-ahead log.

### DIFF
--- a/crates/sparrow-journal/src/batch_io.rs
+++ b/crates/sparrow-journal/src/batch_io.rs
@@ -1,0 +1,216 @@
+use std::collections::HashMap;
+use std::io::Write;
+
+use arrow_array::{ArrayRef, RecordBatch};
+use arrow_buffer::Buffer;
+use arrow_ipc::writer::{DictionaryTracker, EncodedData, IpcDataGenerator, IpcWriteOptions};
+use arrow_schema::SchemaRef;
+use error_stack::{IntoReport, ResultExt};
+use itertools::Itertools;
+use sparrow_batch::{Batch, RowTime};
+
+use crate::error::Error;
+
+/// Write batches to a byte array.
+///
+/// This uses arrow_ipc::writer::IpcDataGenerator directly. This allows it to avoid
+/// copying buffers between the underlying encoded buffer and the stream being written.
+pub(super) struct BatchEncoder {
+    /// Track the dictionary entries in the current segment.
+    dictionary_tracker: DictionaryTracker,
+    /// Generator for encoded batches.
+    gen: IpcDataGenerator,
+    write_options: IpcWriteOptions,
+    schema: SchemaRef,
+}
+
+impl std::fmt::Debug for BatchEncoder {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("BatchEncoder")
+            .field("gen", &self.gen)
+            .field("write_options", &self.write_options)
+            .field("schema", &self.schema)
+            .finish_non_exhaustive()
+    }
+}
+
+pub(super) struct Encoded {
+    up_to_time: RowTime,
+    dictionary_batches: Vec<EncodedData>,
+    batch: Option<EncodedData>,
+}
+
+impl Encoded {
+    pub(super) fn write_to_journal(
+        self,
+        wal: &mut okaywal::WriteAheadLog,
+    ) -> std::io::Result<okaywal::EntryId> {
+        let mut entry = wal.begin_entry()?;
+        entry.write_chunk(&self.up_to_time.bytes())?;
+        for dictionary_batch in self.dictionary_batches {
+            entry.write_chunk(&dictionary_batch.ipc_message)?;
+            entry.write_chunk(&dictionary_batch.arrow_data)?;
+        }
+        if let Some(batch) = self.batch {
+            entry.write_chunk(&batch.ipc_message)?;
+            entry.write_chunk(&batch.arrow_data)?;
+        }
+        entry.commit()
+    }
+}
+
+impl BatchEncoder {
+    pub fn new(schema: SchemaRef) -> error_stack::Result<Self, Error> {
+        Ok(Self {
+            dictionary_tracker: DictionaryTracker::new(false),
+            gen: IpcDataGenerator {},
+            write_options: IpcWriteOptions::default(),
+            schema,
+        })
+    }
+
+    pub fn encode(&mut self, batch: &Batch) -> error_stack::Result<Encoded, Error> {
+        // TODO: Split batches to fit in a certain size.
+        let up_to_time = batch.up_to_time;
+
+        // Create a record batch for the given batch.
+        if let Some(batch) = &batch.data {
+            let batch = RecordBatch::try_new(
+                self.schema.clone(),
+                vec![
+                    batch.time.clone(),
+                    batch.subsort.clone(),
+                    batch.key_hash.clone(),
+                    batch.data.clone(),
+                ],
+            )
+            .into_report()
+            .change_context(Error::WriteBatch)?;
+            let (dictionary_batches, batch) = self
+                .gen
+                .encoded_batch(&batch, &mut self.dictionary_tracker, &self.write_options)
+                .into_report()
+                .change_context(Error::WriteBatch)?;
+            Ok(Encoded {
+                up_to_time,
+                dictionary_batches,
+                batch: Some(batch),
+            })
+        } else {
+            Ok(Encoded {
+                up_to_time,
+                dictionary_batches: vec![],
+                batch: None,
+            })
+        }
+    }
+}
+
+#[derive(Debug)]
+pub(super) struct BatchDecoder {
+    schema: SchemaRef,
+
+    /// Optional dictionaries for each schema field.
+    ///
+    /// Dictionaries may be appended to in the streaming format.
+    dictionaries_by_id: HashMap<i64, ArrayRef>,
+}
+
+impl BatchDecoder {
+    pub fn new(schema: SchemaRef) -> Self {
+        Self {
+            schema,
+            dictionaries_by_id: HashMap::new(),
+        }
+    }
+
+    /// Decode the batch from the given entry.
+    ///
+    /// If it was completely written, will return `Some(batch)`. If it wasn't completely
+    /// written, will return `None`.
+    pub fn decode(
+        &mut self,
+        entry: &mut okaywal::Entry<'_>,
+    ) -> error_stack::Result<Option<Batch>, Error> {
+        // Read the chunks.
+        let Some(chunks) = entry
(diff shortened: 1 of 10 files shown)


crates/sparrow-journal/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+        // TODO: This should use `_last_checkpointed_id` to ensure that only batches
+        // that are part of the checkpoint are included.
+        self.batches
+            .checkpoint()
+            .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
bjchambers, line 252: Move tests to a separate `tests.rs` file.

crates/sparrow-journal/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+        Ok(())
+    }
+
+    fn checkpoint_to(
+        &mut self,
+        _last_checkpointed_id: okaywal::EntryId,
+        _checkpointed_entries: &mut okaywal::SegmentReader,
+        _wal: &okaywal::WriteAheadLog,
+    ) -> std::io::Result<()> {
+        // TODO: This should use `_last_checkpointed_id` to ensure that only batches
```
bjchambers, line 243: I think this is probably critical, and it may reveal an issue with how things are structured (and how we anticipated implementing them).

Specifically:

1. The checkpointing happens asynchronously. So we *should not* just use the current in-memory batches for checkpointing.
2. This means that some batches may have been added to the new segment (and thus the "in-memory" batch vector) since we decided to checkpoint. These batches *should not* be part of the checkpoint (they'll be in the next one) or we'll have duplicate rows.

Possible options:

1. De-duplicate if rows show up in the checkpoint and segment.
2. When checkpointing, use the entries themselves, rather than the "in-memory batches" -- this wouldn't work since the in-memory batches need to be flushed at some point.
3. Use some kind of "timestamped in-memory batches" (associated with the entry IDs). It shouldn't be difficult to associate each element in the vector with the corresponding `EntryId`, and only take a prefix of the vector corresponding to those entries.

I suspect that case 3 is the best option. While we're not trying to handle late data, it does mean that the late data strategy gets a little trickier. We have a few options:

A) When late data arrives, force a checkpoint. Simple, but in the presence of lots of late data this would lead to slow-down and excessively small checkpoints.
B) Leave the batches in memory unordered. When we need to execute on them, we can check to see if there is any overlap, and decide to do (i) concatenation if there is none or (ii) concatenation and sorting if there is or (iii) sorting and merging.
C) Something more intelligent, that manages both the entry IDs associated with batches and the late data.

I suspect B may be the best option, unless we have a way of detecting when a checkpoint is taken.

crates/sparrow-journal/src/batch_io.rs
```diff
@@ -0,0 +1,216 @@
...
+        Self {
+            schema,
+            dictionaries_by_id: HashMap::new(),
+        }
+    }
+
+    /// Decode the batch from the given entry.
+    ///
+    /// If it was completely written, will return `Some(batch)`. If it wasn't completely
+    /// written, will return `None`.
```
jordanrfrazier, line 130: Is `None` a valid state?  What does it mean to not completely write a batch? I didn't see any further explanation in okayWal.

crates/sparrow-journal/src/batch_io.rs
```diff
@@ -0,0 +1,216 @@
...
+
+                return Ok(Some(Batch::new_with_data(
+                    batch.column(3).clone(),
+                    batch.column(0).clone(),
+                    batch.column(1).clone(),
+                    batch.column(2).clone(),
+                    up_to_time,
+                )));
+            }
+        }
```
jordanrfrazier, line 200: This logic feels interesting. Iterating through all chunk pairs and adding the batch to the buffer, but once we peek and see we're at the last chunk pair, return a record batch? 

Is it done this way because we have to set the last batch's message with the `header_as_record_batch()` instead of `as_dict_batch()`?

crates/sparrow-journal/src/batch_io.rs
```diff
@@ -0,0 +1,216 @@
...
+
+        // If we get here, we must not have had any batches above.
+        Ok(Some(Batch::new_empty(up_to_time)))
+    }
+}
+
+fn decode_message(buf: &[u8]) -> error_stack::Result<arrow_ipc::Message<'_>, Error> {
+    arrow_ipc::root_as_message(buf)
+        .map_err(|e| {
+            // We'd like to conrert the flatbuffer error directly.
```
jordanrfrazier, line 210: convert

crates/sparrow-journal/src/checkpoints.rs
```diff
@@ -0,0 +1,254 @@
...
+use futures::stream::BoxStream;
+use futures::{StreamExt, TryStreamExt};
+use sparrow_batch::{Batch, RowTime};
+
+use crate::error::Error;
+
+/// Manages the checkpointed files.
+///
+/// We could use Parquet files for check-pointing but we instead use
+/// Arrow IPC files, since we believe they should be easier to read.
```
jordanrfrazier, line 18: Can you elaborate on this for me? Why are Arrow IPC files > parquet

crates/sparrow-journal/src/checkpoints.rs
```diff
@@ -0,0 +1,254 @@
...
+            .ok_or(Error::RecoveringCheckpoints)?;
+        let min_present_time =
+            RowTime::from_str(min_present_time).change_context(Error::RecoveringCheckpoints)?;
+
+        let max_present_time = reader
+            .custom_metadata()
+            .get("max_present_time")
+            .ok_or(Error::RecoveringCheckpoints)?;
+        let max_present_time =
+            RowTime::from_str(max_present_time).change_context(Error::RecoveringCheckpoints)?;
```
jordanrfrazier, line 95: Sheesh. Wonder if there would be a macro invocation or splitting this up into a helper or two to decrease the amount of `into_report().change_context(Error::RecoveringCheckpoint)`s

crates/sparrow-journal/src/checkpoints.rs
```diff
@@ -0,0 +1,254 @@
...
+                .write(&record_batch)
+                .into_report()
+                .change_context(Error::WriteCheckpoint)?;
+
+            writer
+                .finish()
+                .into_report()
+                .change_context(Error::WriteCheckpoint)?;
+
+            tracing::info!("Checkpointed {rows} jouranl rows to {file_name}");
```
jordanrfrazier, line 168: journal

crates/sparrow-journal/src/checkpoints.rs
```diff
@@ -0,0 +1,254 @@
...
+            checkpoint_dir,
+            checkpoints,
+            schema,
+        })
+    }
+
+    /// Create a check-point from the given batches.
+    ///
+    /// This method assumes the batches are non-overlapping and ordered by time.
+    pub fn checkpoint(
```
jordanrfrazier, line 125: At this time, journals will include the "prepared" data, correct? (we have the Batch with separate time,subsort,key columns, so I assume that must be the case) Is there any reason we'd want to instead journal the user data directly? For example, if we support decimals in `prepare`, or any non-backwards-compatible change, we'd want to run prepare on the user data again.

crates/sparrow-journal/src/error.rs
```diff
@@ -0,0 +1,27 @@
+#[derive(derive_more::Display, Debug)]
+pub enum Error {
+    #[display(fmt = "checkpoint path ({}) is not a directory", "_0.display()")]
+    CheckpointDirNotDir(std::path::PathBuf),
```
jordanrfrazier, line 4: Maybe just `CheckpointDir`? The text is descriptive

crates/sparrow-journal/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+    ) -> error_stack::Result<Self, Error> {
+        Ok(Self {
+            batches: Mutex::new((vec![], RowTime::ZERO)),
+            checkpoints: Mutex::new(Checkpoints::try_new(checkpoint_dir, schema)?),
+        })
+    }
+
+    fn add_batch(&self, batch: Batch) -> error_stack::Result<(), Error> {
+        let mut lock = self.batches.lock();
+        batch.max_present_time();
```
jordanrfrazier, line 160: .

crates/sparrow-journal/src/lib.rs
jordanrfrazier: Agreed, critical to do. And agreed with option 3 -- disregarding late data issues for now, the checkpointer could keep track of the timestamps for each checkpoint, and have separate vectors for each set of batches. Or, we could include a `checkpoint_version` alongside each batch that flows through the system: 
```
WatermarkedAndCheckpointBatch {
  watermark: i64
  checkpoint: usize
  batch: Batch 
} 
```

I'm not sure I understand the idea for unordered batches in memory, but that likely needs a whole conversation around late data.

crates/sparrow-journal/src/batch_io.rs
bjchambers: https://khonsulabs.github.io/okaywal/main/okaywal/struct.Entry.html#method.read_chunk

crates/sparrow-journal/src/batch_io.rs
bjchambers: Yeah -- this roughly based on what the Arrow IPC writer does under the hood. Each "dictionary column" becomes a dictionary batch, and then you have the final batch. So you end up with:

- Dict 1 Metadata, Dict 1 Data
- Dict 2 Metadata, Dict 2 Data
- ...
- Dict N Metadata, Dict N Data
- Batch Metadata, Batch Metadata

In many cases, there will be no dicts, but figured it was worth keeping.

crates/sparrow-journal/src/checkpoints.rs
bjchambers: Arrow IPC files write data out using the same format as Arrow, so they can largely be mapped into memory directly, while Parquet requires some things like decompression and changing formats. So, we'd likely expect:

- Arrow IPC may be faster for read/write.
- Parquet may have wider support as an archival format (eg., on object store)

PR #846, opened 2023-11-01
tests: Fix session tests

bjchambers:


crates/sparrow-session/src/partitioned.rs
```diff
@@ -1,5 +1,5 @@
 mod session;
 pub use session::*;
 
-#[cfg(tests)]
```
bjchambers, line 4: The configuration is `#[cfg(test)]`, and the standard module name is `tests`. Since nothing set the `tests` configuration, this was never being included.

crates/sparrow-session/src/partitioned/partitioned.rs
```diff
@@ -1 +0,0 @@
-mod basic_tests;
```
bjchambers, line 1: Nothing was referencing `sparrow_session::partitioned::partitioned` which corresponded to this file.

crates/sparrow-session/src/tests.rs
```diff
@@ -1,2 +0,0 @@
-//! Partitioned execution tests
```
bjchambers, line 1: Nothing was referencing this -- the `mod tests` in partitioned.rs corresponded to `partitioned/tests.rs` which was missing.

crates/sparrow-session/src/partitioned/tests.rs
bjchambers, line 1: Moved `basic_tests.rs` to `tests.rs` to fix the missing `partitioned::tests` module. Since most of the coverage should be in Python, we can hopefully keep the quantity of tests small enough to not needing to split it up.

jordanrfrazier (approved):

PR #852, opened 2023-11-07
docs: rename timeline -> timestream

jordanrfrazier:
Hopefully. Need to check if all files were appropriately changed as well so we don't have missing images, etc.


crates/sparrow-qfr-tool/src/chrome_tracing/trace_event.rs
```diff
@@ -124,7 +124,7 @@ pub(crate) enum Event {
 #[derive(Serialize, Eq, PartialEq, Debug, Default)]
 #[allow(dead_code)]
 pub(crate) enum EventScope {
-    /// The event will be drawn from the top to bottom of the timeline.
+    /// The event will be drawn from the top to bottom of the timestream.
```
bjchambers, line 127: This was referring to the "timeline" in the chrome tracing view. Probably don't want to change.

docs-src/modules/developing/pages/queries.adoc
bjchambers, line 1: We should probably not change the `docs-src` directory. That is the source for the old site. We should separately discuss getting rid of that site at some point.

bjchambers (approved):

PR #853, opened 2023-11-08
draft: fix execution lock in session

jordanrfrazier:
Still issues around `sending` the future across threads, I think. Also had to use the `tokio::sync::Mutex` in order to hold the lock across the `await`, but I'm not sure that this is the correct pattern we should follow, as it may still result in a deadlock if we never receive a new batch from `next`.


python/src/execution.rs
```diff
@@ -13,15 +13,15 @@ use crate::error::{ErrorContext, IntoError, Result};
 #[pyclass]
 #[derive(Clone)]
 pub(crate) struct Execution {
-    execution: Arc<Mutex<Option<RustExecution>>>,
+    execution: Arc<Option<RustExecution>>,
```
bjchambers, line 16: Given the difficulties, it seems like we may be better off having submitting an execution return a pair of `Execution, StopChannel`. Then we can leave the mutex in the execution code, but interact with the stop channel without locking it. Thoughts?

jordanrfrazier: Yeah, I guess so 😞 . I'll try that route

PR #854, opened 2023-11-08
docs: update timestreams p1/p2 blogs

jordanrfrazier:
Updated p1 and p2


bjchambers:
> Don't think there's anything else required for p1 and p2 to publish them, unless we want to discuss any changes to the examples/graphics.

In p2 we should update the examples which currently used Fenl syntax to using Python syntax.

python/docs/blog/posts/2023-05-25-introducing-timestreams-part-2.qmd
```diff
@@ -39,16 +39,13 @@ The corresponding query is shown below written in two equivalent ways.
...
-```fenl
-sum(Purchases.amount)
 
-# OR
-
-Purchases.amount
-| sum()
+```{.python}
+Purchases.col("amount").sum()
+TODO: FRAZ - should these include '.to_pandas()' or '.preview()'/other?
```
jordanrfrazier, line 45: .

